Issue #216, opened 2020-12-09
Request Card - Seasons + Email overflow


azoitos:
**Describe the bug**
When there are many seasons in a tv show, the number of seasons overlap below the request card image
If a user has a long email, it will overlap on the request card image.

**Expected behavior**
Add a overflow-scroll x to seasons.
Add ellipsis to email.

**Screenshots**

![](https://cdn.discordapp.com/attachments/745822573492961382/786045692162473994/unknown.png)

![](https://cdn.discordapp.com/attachments/745822573492961382/786045757995876353/unknown.png)

PR #217, opened 2020-12-09
fix(frontend): update overflow issues with seasons + email

azoitos:
Add all message for full seasons

bug #216


src/styles/globals.css
```diff
@@ -36,3 +36,16 @@ body {
...
+/* Hide scrollbar for Chrome, Safari and Opera */
+.example::-webkit-scrollbar {
+  display: none;
+}
+
+/* Hide scrollbar for IE, Edge and Firefox */
+.example {
+  -ms-overflow-style: none;  /* IE and Edge */
+  scrollbar-width: none;  /* Firefox */
+}
```
sct, line 51: We should probably not name this class `example`. Maybe something like `hide-scrollbar`?

src/components/RequestCard/index.tsx
```diff
@@ -19,6 +19,7 @@ import StatusBadge from '../StatusBadge';
 const messages = defineMessages({
   requestedby: 'Requested by {username}',
   seasons: 'Seasons',
+  all: 'ALL',
```
sct, line 22: Instead of capitalizing this here, lets instead make it `All` and then apply the `uppercase` text transform to the place we use it.

sct (changes requested):

src/styles/globals.css
azoitos: Updated to hide-scrollbar

src/components/RequestCard/index.tsx
azoitos: Updated to All. Added uppercase to span

sct (approved):

PR #337, opened 2020-12-16
feat(settings): basic log api and ui

### LINKED ISSUE #272: About and Logs pages give 404
**Describe the bug**
If you go to About or Logs pages under settings it will give 404 error.

**Are you on latest or develop branch?**
Same on 1.3.2 or develop tip.

lochstar:
#### Description

This implements a very basic log UI. The number of rows returned is set to 1000. I thought I'd get this in so others can pick up from it and we can decide what we'd like from the log ui.

I separated the winston transport configs to use a JSON format for the file log, using the current text-based log that we output to the console wasn't compatible with `winston.query`. We should probably set some [options related to maximum files and filesize](https://github.com/winstonjs/winston/blob/2625f60c5c85b8c4926c65e98a591f8b42e0db9a/docs/transports.md#file-transport).

As the format of the log `overseerr.log` has changed, we may need to purge or migrate an existing log on upgrade?

Other features we could implement:

- Filters e.g. by label/level
- Download link/view raw
- Nicer styling
- Manual/auto refresh option

image

- Fixes #272

### DIFF
--- a/src/components/Settings/SettingsLogs/index.tsx
+++ b/src/components/Settings/SettingsLogs/index.tsx
@@ -1,14 +1,72 @@
 import React from 'react';
+import useSWR from 'swr';
+import Error from '../../../pages/_error';
+import LoadingSpinner from '../../Common/LoadingSpinner';
+import {
+  FormattedDate,
+  FormattedTime,
+  useIntl,
+  defineMessages,
+} from 'react-intl';
 
-// We will localize this file when the complete version is released.
+const messages = defineMessages({
+  logs: 'Logs',
+  logsDescription:
+    'You can access your logs directly in stdout (container logs) or looking in /app/config/logs/overseerr.logs',
+});
 
 const SettingsLogs: React.FC = () => {
+  const intl = useIntl();
+  const { data, error } = useSWR('/api/v1/settings/logs');
+
+  if (error) {
+    return <Error statusCode={500} />;
+  }
+
+  if (!data && !error) {
+    return <LoadingSpinner />;
+  }
+
+  if (!data) {
+    return <LoadingSpinner />;
+  }
+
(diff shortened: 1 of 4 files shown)


sct:
I converted this to a draft for now while we get it ironed out.

src/components/Settings/SettingsLogs/index.tsx
```diff
@@ -1,14 +1,72 @@
...
+  FormattedTime,
+  useIntl,
+  defineMessages,
+} from 'react-intl';
 
-// We will localize this file when the complete version is released.
+const messages = defineMessages({
+  logs: 'Logs',
+  logsDescription:
+    'You can access your logs directly in stdout (container logs) or looking in /app/config/logs/overseerr.logs',
```
sct, line 15: We will want to have the same `<code>` tags in here like we did in the temporary message. If you only put the tags in here by themselves though, they will be just rendered out in the string.

You can see how this is done in the `RequestList` component. I apply <strong> tags in there.

src/components/Settings/SettingsLogs/index.tsx
```diff
@@ -1,14 +1,72 @@
...
+                  hour12={false}
+                />
+              </span>
+              <span className="inline">
+                [{row.level}][{row.label}]:
+              </span>
+              <span className="inline">{row.message}</span>
+            </div>
+          ))}
+        </div>
```
sct, line 69: We can maybe re-use the table component here to make this look nice? Check out how i use it in the RequestList component.

sct (changes requested):

sct:
@lochstar Are you still working on this?

lochstar:
Planning on it, yes.

PR #389, opened 2020-12-18
Change default internal port to 5055

sct:
#### Description

This will change the default internal port to 5055. This will help avoid future confusion about which port to use or forward to.

This PR, once merged, will break access to Overseerr for anyone who is accessing the app through the Docker portforward of 5055 to 3000. It will need to be updated to 5055 internally as well.

The README is updated as well with the new command.

#### Todos

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,4 +25,4 @@ COPY --from=BUILD_IMAGE /app/node_modules ./node_modules
 
 CMD yarn start
 
-EXPOSE 3000
+EXPOSE 5055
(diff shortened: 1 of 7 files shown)


mazzetta86:
Hi, could you please update also the docker page to avoid confusion for new users? https://hub.docker.com/r/sctx/overseerr

sct:
@mazzetta86 Done. Thanks!

prattable:
It looks like the API Documentation section of the dicker page was overlooked and needs updated. There is a reference to use the 3000 port.

PR #428, opened 2020-12-21
feat: import users from plex & fix: user delete button re-enable

### LINKED ISSUE #281: Import all users from Plex
Currently, once a user logs in then you can change their permissions. I believe a better workflow would be to import all the users and be able to set permissions ahead of time. 

Be able to manage users before they log in. I believe a sync job to Plex to collect all the users will be needed and should be ran periodically.

johnpyp:
#### Description

Adds an api endpoint and corresponding UI button for importing users from the connected Plex server, as initially proposed in #281. 

Specifically, it imports all "shared library users" that are real users (have a corresponding username and email). Most of the user properties are imported, except notably the `plexToken`, but the auth flow takes care of populating that after first-login.

This PR also includes a minor one-line fix for the delete user button in the popup modal not re-enabling after you've already deleted one user, though I can move it into a different PR if desired.

Import Button

#### Todos

- Fixes #281


src/components/UserList/index.tsx
```diff
@@ -66,10 +70,31 @@ const UserList: React.FC = () => {
...
 
+  const importFromPlex = async () => {
+    setImporting(true);
+
+    try {
+      await axios.post('/api/v1/user/import-from-plex');
+      addToast(intl.formatMessage(messages.importedfromplex), {
+        autoDismiss: true,
+        appearance: 'success',
+      });
```
sct, line 86: What about having the toast say the # of new users that were added? `4 new users imported from Plex` for example.

src/components/UserList/index.tsx
```diff
@@ -66,10 +70,31 @@ const UserList: React.FC = () => {
         appearance: 'error',
       });
     } finally {
+      setDeleting(false);
```
sct, line 73: whoops. thanks!

src/components/UserList/index.tsx
```diff
@@ -18,6 +18,9 @@ import globalMessages from '../../i18n/globalMessages';
 
 const messages = defineMessages({
   userlist: 'User List',
+  importfromplex: 'Import Users From Plex',
+  importfromplexerror: 'Something went wrong importing users from plex',
+  importedfromplex: 'Users successfully imported from plex',
```
sct, line 23: Plex should be capitalized here.

server/routes/user.ts
```diff
@@ -142,4 +143,51 @@ router.delete<{ id: string }>('/:id', async (req, res, next) => {
   }
 });
 
+router.post('/import-from-plex', async (req, res, next) => {
+  try {
+    const defaultPermission = req.body.permissions || Permission.REQUEST;
```
sct, line 148: We can now set default permissions through the settings page so you can grab the current `defaultPermissions` value from the settings here.

sct (commented):
Fantastic PR! Looks really good!

Just a few small changes and we can get this merged!

sct:
Also don't forget to run `yarn i18n:extract` and commit the generated `en.json` file!

johnpyp:
Alright, these changes should be in there now 👍

src/components/UserList/index.tsx
```diff
@@ -66,10 +70,38 @@ const UserList: React.FC = () => {
...
+  const importFromPlex = async () => {
+    setImporting(true);
+
+    try {
+      const { data: createdUsers } = await axios.post(
+        '/api/v1/user/import-from-plex'
+      );
+      addToast(
+        `${createdUsers.length} ${intl.formatMessage(
+          messages.importedfromplex
```
sct, line 87: Not every language will put the number first. The actual localized string needs to use the number in it. `{userCount} new users imported from Plex`. Then you pass it the value of `userCount` as the second argument of formatMessage in an object.

johnpyp: Ah gotcha. I took a look at the docs for react-intl and went with the 3-branches strategy of "no new users", "1 new user", and "N new users".

sct:
I think it's safe to take this out of draft 😁

johnpyp:
ah yep 👍

sct (approved):
Perfect. Thanks a ton!

sct:
@all-contributors please add @johnpyp for code

Issue #283, opened 2020-12-15
Movie and TV Show Trailers


Fiala06:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
I would love to see trailers added to the Movie and TV Show pages. 

Should make it easier to decide to add a new movie/show or not :)

**Describe alternatives you've considered**
N/A

**Additional context**
Would prefer a embedded solution rather than a link to a website.

samwiseg0:
Please fill out the provided template.

Fiala06:
> Please fill out the provided template.

Fixed

PR #454, opened 2020-12-22
YouTube Movie/TV Trailers

jayesh100:
#### Description
Use TMDB API to determine youtube ID for trailer and add button to make this accessible through the UI 

#### Screenshot
watch-trailer-button

#### Todos

- [ ] Database migration created (if required) - N/A

- Helps accomplish #283


mrhotio:
Yay, just as I ordered. +1

sct:
Why not just use the videos call from TMDB to get trailers and other related videos? We can grab videos with the main request for movie/tv details by adding it to `append_to_response`. This is what I had planned anyways.

I also was thinking of having a more dedicated "View Trailer" button somewhere as well as in-app embeds for the playback.

danshilm:
I'm curious about this PR. How does this deal with multiple trailers from TMDB? Does this just grab the first one?
It might be a better option to get the latest trailer instead if that's the case.

sct:
Don't forget to update the `overseerr-api.yml` with any new data that we are returning here.

server/models/common.ts
```diff
@@ -84,3 +85,12 @@ export const mapExternalIds = (eids: TmdbExternalIds): ExternalIds => ({
   tvrageId: eids.tvrage_id,
   twitterId: eids.twitter_id,
 });
+
+export const getYoutubeTrailer = (
+  videoResult: TmdbVideoResult
+): string | undefined =>
+  videoResult?.results
+    .filter((r) => r.type === 'Trailer' && r.site === 'YouTube')
+    .sort((a, b) => a.size - b.size)
```
jayesh100, line 94: @danshilm you might be interested in this

jayesh100:
@sct cheers, missed that in the TMDB doco. Switched over to using the TMDB API, and applied changes for TV shows too. Moved the UI into a button (see updated description + screenshot above) - haven't had the chance to look at embedded playback yet, but i imagine this is a good base to build upon.

@danshilm now that it has been updated to use the TMDB API, i've added the logic in the common.ts "getYoutubeTrailer" function. It currently picks based on the best available resolution, as there's no timestamp or equivalent in the videos response.

sct:
> @sct cheers, missed that in the TMDB doco. Switched over to using the TMDB API, and applied changes for TV shows too. Made the trailers accessible through a button (see updated description + screenshot above) - haven't had the chance to look at embedded playback yet, but i imagine this is a good base to build upon.

This is a _fantastic_ base to build on. Really appreciate the PR.

I am not sure yet how I feel about the placement for the Watch Trailer button. I might move it in the future depending on some other changes I am considering. But it's fine to have it there for now.

I checked a few titles on TMDB and it _looks_ like the most recent "official" trailer tends to always be the first trailer returned. I can't say for sure that pattern will 100% be followed but I think with your filtering method it will likely still prefer the first trailer in most cases. So, should work well.

And yeah, embedding the video right now isn't super important. We can add that later!

danshilm:
> @danshilm now that it has been updated to use the TMDB API, i've added the logic in the common.ts "getYoutubeTrailer" function. It currently picks based on the best available resolution, as there's no timestamp or equivalent in the videos response.

I checked the TMDB API documentation before asking but I saw nothing about the sorting of the videos that comes from it. I was hoping there was some way to get the dates that are on the `/videos` page (those are the dates the videos are submitted to TMDB) and that either sct or you knew about a way to get those, but alas. We work with what we have.

overseerr-api.yml
```diff
@@ -1689,7 +1693,8 @@ paths:
             application/json:
               schema:
                 type: array
-                $ref: '#/components/schemas/User'
+                items:
+                  $ref: '#/components/schemas/User'
```
sct, line 1697: Oh great catch. Thanks 😁

sct:
What do you think about instead returning all the videos in the api response as an array and picking out which we want for the youtube trailer on the frontend instead? In the future we may want to show other videos somewhere on the details page.

jayesh100:
@sct Makes sense!

overseerr-api.yml
```diff
@@ -526,6 +560,8 @@ components:
           type: string
         posterPath:
           type: string
+        youtubeTrailerUrl:
+          type: string
```
sct, line 564: No longer needed?

jayesh100: Correct, it's not

sct (approved):
Awesome 🎉 

Let's get this merged!

sct:
@all-contributors please add @jayesh100 for code

PR #474, opened 2020-12-23
feat: simple failed request handling

johnpyp:
#### Description

When a movie or series is added with radarr or sonarr, if it fails, this changes the media state to unknown and sends a notification to admins.

Client side this will look like a failed state along with a retry button, which when clicked will delete the request and re-add it.

React Request List
Discord Notification

#### Todos


server/lib/notifications/index.ts
```diff
@@ -5,7 +5,8 @@ export enum Notification {
   MEDIA_PENDING = 2,
   MEDIA_APPROVED = 4,
   MEDIA_AVAILABLE = 8,
-  TEST_NOTIFICATION = 16,
+  MEDIA_FAILED = 16,
+  TEST_NOTIFICATION = 32,
```
sct, line 9: This is acceptable here because we aren't using these yet to determine which notifications are enabled/disabled... but normally we would never want to change the values here.

johnpyp: Gotcha yeah, I hadn't considered that.

src/components/RequestList/RequestItem/index.tsx
```diff
@@ -68,7 +77,30 @@ const RequestItem: React.FC<RequestItemProps> = ({ request, onDelete }) => {
...
+            mediaId: request.media.tmdbId,
+            tvdbId: request.media.tvdbId,
+            mediaType: MediaType.TV,
+            seasons: request.media.seasons,
+          };
+    try {
+      await axios.delete(`/api/v1/request/${request.id}`);
+      await axios.post<MediaRequest>(`/api/v1/request`, requestPayload);
+    } catch (e) {
+      // No error handling like in delete func, maybe add toasts for this too?
```
sct, line 99: Might as well add toasts here. Wouldn't hurt.

src/components/RequestList/RequestItem/index.tsx
```diff
@@ -68,7 +77,30 @@ const RequestItem: React.FC<RequestItemProps> = ({ request, onDelete }) => {
...
+        ? { mediaId: request.media.tmdbId, mediaType: MediaType.MOVIE }
+        : {
+            mediaId: request.media.tmdbId,
+            tvdbId: request.media.tvdbId,
+            mediaType: MediaType.TV,
+            seasons: request.media.seasons,
+          };
+    try {
+      await axios.delete(`/api/v1/request/${request.id}`);
+      await axios.post<MediaRequest>(`/api/v1/request`, requestPayload);
```
sct, line 97: If someone else requests this, and it fails. Won't this then change the creator of the request to the admin who retries?

There is an endpoint to actually update the status of a request back to approved. That _should_ actually try to send the request again. It will also prevent the request from jumping in the list from whatever page they are on back to the top.

sct:
I just thought of one more thing. If someone deletes all requests associated with any media, we keep the media but set its status back to Unknown. That would mark it as failed, but it isn't.

johnpyp:
> I just thought of one more thing. If someone deletes all requests associated with any media, we keep the media but set its status back to Unknown. That would mark it as failed, but it isn't.

So the idea is that, if all requests are deleted with any media, it wouldn't matter if it's marked as failed, since there's no request to actually see as failed. However, when there *are* requests and the media is *still* unknown, that signifies it's failed since that shouldn't be a possible state when the media is in Radarr/Sonarr or available.

sct:
> > I just thought of one more thing. If someone deletes all requests associated with any media, we keep the media but set its status back to Unknown. That would mark it as failed, but it isn't.
> 
> So the idea is that, if all requests are deleted with any media, it wouldn't matter if a request connected to it would be marked as failed, since there's no request to actually see as failed. However, when there _are_ requests and the media is _still_ unknown, that signifies it's failed since that shouldn't be a possible state when the media is in Radarr/Sonarr or available.

Ah yeah true. Since there are no requests it won't show up in the request list so it doesn't really matter.

src/components/RequestList/RequestItem/index.tsx
johnpyp: I ended up doing two requests, one to set it to pending and another to set it to approved, so it actually triggers the AfterUpdate hook. This should probably be a dedicated endpoint in the next iteration on this I think.

johnpyp:
One thing I didn't consider about setting to Pending and then back to Approved are the notifications there... might be better just to include a small refactor for a dedicated endpoint.

sct:
> One thing I didn't consider about setting to Pending and then back to Approved are the notifications there... might be better just to include a small refactor for a dedicated endpoint.

The dedicated endpoint would just try to send the media back to radarr/sonarr again, yes? Because if we touch the statuses on the request it will still refire the listeners and trigger notifications.

johnpyp:
Yeah, so I added a `POST /request/:requestId/:command` endpoint (like the status endpoint), with just `retry` as the only command for now. The retry itself will 
1) `updateParentStatus` (to effectively reset statuses as if it was a new request) 
2) Attempt to send the media to Sonarr/Radarr.

This is the same as the normal AfterUpdate hooks, with the exception of not sending notifications.

sct:
I think having its own endpoint is better. Right now you are setting an action in an endpoint that is supposed to accept a status. Even if its an overload, the docs will be clearer in that regard.

overseerr-api.yml
```diff
@@ -2155,7 +2179,7 @@ paths:
           required: true
           schema:
             type: string
-            enum: [pending, approve, decline, available]
+            enum: [pending, approve, decline, available, retry]
```
sct, line 2182: Don't need retry in this enum anymore

johnpyp: Ah yep, missed that.

src/components/StatusBadge/index.tsx
```diff
@@ -36,6 +36,12 @@ const StatusBadge: React.FC<StatusBadgeProps> = ({ status }) => {
           {intl.formatMessage(globalMessages.pending)}
         </Badge>
       );
+    case MediaStatus.UNKNOWN:
+      return (
+        <Badge badgeType="danger">
+          {intl.formatMessage(globalMessages.failed)}
+        </Badge>
+      );
```
sct, line 44: This will cause the failed status badge to show up in many place throughout the app, not just the RequestList.

RequestCard uses this. So does the movie/tv details pages now.

src/components/RequestList/RequestItem/index.tsx
```diff
@@ -15,16 +15,21 @@ import useSWR from 'swr';
...
 import axios from 'axios';
 import globalMessages from '../../../i18n/globalMessages';
 import Link from 'next/link';
+import { useToasts } from 'react-toast-notifications';
 
 const messages = defineMessages({
   requestedby: 'Requested by {username}',
   seasons: 'Seasons',
   notavailable: 'N/A',
+  failedretry: 'Something went wrong retrying that request',
```
sct, line 32:
```suggestion
  failedretry: 'Something went wrong retrying the request',
```

sct (approved):
LGTM!

Issue #453, opened 2020-12-22
Anime not being picked up


modem7:
**Describe the bug**
Anime doesn't seem to be getting picked up during library scan. 

**What version of Overseerr are you running?**
1.12

**To Reproduce**

Search for anime that is already within Plex.

**Expected behavior**
Like TV shows, expect "available". 

**Screenshots**
![image](https://user-images.githubusercontent.com/4349962/102846476-f91fc200-4407-11eb-8500-80f90c1fd598.png)

**Desktop (please complete the following information):**

- OS: win10x64
- Browser Vivaldi
- Version 3.5.2115.81

**Additional context**
Plex library is not using any 3rd party scanners or agent unlike #398 
![image](https://user-images.githubusercontent.com/4349962/102846544-22d8e900-4408-11eb-92ba-ccdec4fdc408.png)

samwiseg0:
Isn't that the new beta Plex scanner that is being tested?

modem7:
Unsure, whilst I am on the beta Plex release, I only have two options, Absolute Series Scanner (3rd party) or the Plex Series Scanner (same one as the TV series which are working as expected).

![image](https://user-images.githubusercontent.com/4349962/102846795-d7730a80-4408-11eb-9ac3-203e89dd4df0.png)

samwiseg0:
Are you running this because I don't even have that option?

https://forums.plex.tv/t/new-plex-media-server-tv-show-series-scanner-and-agent-preview/665778

modem7:
Not yet, I'm still on 1.21.1.3830 with no custom modifications bar the hama agent (which I'll be removing at some point as I don't actually use it anymore)

sct:
![image](https://user-images.githubusercontent.com/234213/102847886-cf41bc00-4456-11eb-9c3f-0b7c43aeeefd.png)
My TV library doesn't have that option either 🤔

samwiseg0:
> I haven't had the option to upgrade my library like I did with the new movie scanner.
What does this mean? This makes me think you are running the beta since that is what is in the forum post. 

I do not have that option and it clearly says is the new agent in your screen shot. This is a dupe of https://github.com/sct/overseerr/issues/378. You will need to run the standard TMDb or TVDb agents

samwiseg0:
Can you pull the XML of at least 2 items that are failing for us?

Go to the item git get info. Then view XML

[...]

PR #538, opened 2020-12-31
feat: map AniDB IDs from Hama agent to tvdb/tmdb/imdb IDs

mmozeiko:
#### Description

This adds hama agent support - maps AniDB IDs to tvdb/tmdb/imdb IDs. So once anime is identified in Plex for hama agent libraries, it will show up available as tv show or movie.

Hama agent stores tvdb or anidb in guid. See Absolute Series Scanner for details how it can be configured: https://github.com/ZeroQI/Absolute-Series-Scanner

If tvdb is detected in guid, then same code as regular tvdb lookups is used to map it to tvshow.
If anidb is detected in guid, then it is remapped using ScudLee mappings xml : https://github.com/ScudLee/anime-lists/ These are same mappings that Hama agent uses for matching metadata to content. XML mappings are downloaded once a day and cached on local disk (`config/anime-list.xml` file). 

Because of way how hama stores movies in Special season of tv show library, there's extra code in processShows to create movies. Either directly or by walking over each episode in Specials/ Season0 to try to map it to tmdb/imdb.

The solution is not ideal, as unknown anidb ids are not mapped at all, just skipped - same as before. But that's basically issue with ScudLee mappings - you need to go to their repo and update mappings, if you want missing/newer stuff to be picked up.

I've tested this on my 500+ anime library, and I see most of the stuff being picked up correctly.

#### Todos

- See #453 for talk about hama agent.


DJScias:
Hello there,

While I haven't looked over the code, I should mention that HAMA no longer uses Scudlee's repo as it has become stale and inactive.

They've switched to this one which is more active and actually maintained: https://github.com/Anime-Lists/anime-lists

As can be seen here: 
https://github.com/ZeroQI/Absolute-Series-Scanner/commit/1a155ab8948d1ee35dce1e0600b8ee210f1af173

It might be best to use the same switch.

Kind regards,
DJScias

mmozeiko:
Oh, nice that seems to be more up to date. Will switch to that.

server/api/animelist.ts
```diff
@@ -0,0 +1,174 @@
...
+          imdbId: imdbId,
+        };
+
+        if (tvdbId) {
+          const mappingLists = anime['mapping-list'];
+          if (mappingLists) {
+            for (const mappingList of mappingLists) {
+              for (const mapping of mappingList['mapping']) {
+                const text = mapping['_'];
+                if (text && mapping['$']['tvdbseason'] == '0') {
```
sct, line 71: Strict comparison here

```suggestion
                if (text && mapping['$']['tvdbseason'] === '0') {
```

server/api/animelist.ts
```diff
@@ -0,0 +1,174 @@
...
+  // each episode in season 0 from TVDB can map to movie (AniDB id)
+  private specials: { [tvdbId: number]: { [episode: number]: number } } = {};
+
+  public isLoaded = () => Object.keys(this.mapping).length !== 0;
+
+  private loadFromFile = async () => {
+    logger.info('Loading Anime-List mapping file');
+    try {
+      const file = await fsp.readFile(LOCAL_PATH);
+      const xml = await xml2js.parseStringPromise(file);
```
sct, line 38: Can we add some types for what we expect this XML to look like? We don't have to type everything exactly, just what we are using. Should be less error prone to things like typos in the future.

server/api/animelist.ts
```diff
@@ -0,0 +1,174 @@
...
+                }
+              }
+            }
+          }
+          // some movies do not have mapping-list, which means 1 episode in specials
+          // matches movie itself
+          if (imdbId) {
+            if (
+              a['defaulttvdbseason'] == '0' &&
+              anime['mapping-list'] == undefined
```
sct, line 89: Strict comparisons
```suggestion
              a['defaulttvdbseason'] === '0' &&
              anime['mapping-list'] === undefined
```

server/api/animelist.ts
```diff
@@ -0,0 +1,174 @@
...
+              this.specials[tvdbId][1] = anidbId;
+            }
+          }
+        }
+      }
+      logger.info(
+        `Loaded ${
+          Object.keys(this.mapping).length
+        } AniDB items from Anime-List mapping file`
+      );
```
sct, line 103: Add a label
```suggestion
      logger.info(
        `Loaded ${
          Object.keys(this.mapping).length
        } AniDB items from Anime-List mapping file`,
        { label: 'AniDB Sync' }
      );
```

server/api/animelist.ts
```diff
@@ -0,0 +1,174 @@
...
+    } catch (e) {
+      throw new Error(`Failed to load anidb mappings: ${e.message}`);
+    }
+  };
+
+  private downloadFile = async () => {
+    logger.info('Downloading latest Anime-List mapping file');
```
sct, line 110: Log label

```suggestion
    logger.info('Downloading latest Anime-List mapping file', { label: 'AniDB Sync' });
```

server/job/plexsync/index.ts
```diff
@@ -182,6 +231,68 @@ class JobPlexSync {
...
+        const tvdbId = matched?.[1];
+
+        if (tvdbId) {
+          tvShow = await this.tmdb.getShowByTvdbId({ tvdbId: Number(tvdbId) });
+          if (animeList.isLoaded()) {
+            await this.processHamaSpecials(plexitem, metadata, Number(tvdbId));
+          } else {
+            this.log(
+              `Hama id ${plexitem.guid} detected, but library agent is not set to Hama`,
+              'info'
```
sct, line 245: Might be better to have the log level "warn" here? Will need to set it to the accepted log levels in the private log method though.

server/job/plexsync/index.ts
```diff
@@ -182,6 +231,68 @@ class JobPlexSync {
...
+            );
+          }
+        }
+      } else if (metadata.guid.match(hamaAnidbRegex)) {
+        const matched = metadata.guid.match(hamaAnidbRegex);
+
+        if (!animeList.isLoaded()) {
+          this.log(
+            `Hama id ${plexitem.guid} detected, but library agent is not set to Hama`,
+            'info'
```
sct, line 255: Same thing here

server/job/plexsync/index.ts
```diff
@@ -182,6 +231,68 @@ class JobPlexSync {
...
+            }
+
+            // if getting tvShow above failed, try with tmdb/imdb for movie
+            if (tvShow == null) {
+              if (result.tmdbId) {
+                return await this.processMovieWithId(
+                  plexitem,
+                  undefined,
+                  result.tmdbId
+                );
```
sct, line 282: Maybe we should try to lookup by tmdb first if we have it? Then we can skip checking tvdb altogether.

server/job/plexsync/index.ts
```diff
@@ -351,6 +462,18 @@ class JobPlexSync {
...
+  // checks if any of this.libraries has Hama agent set in Plex
+  private async hasHamaAgent() {
+    const plexLibraries = await this.plexClient.getLibraries();
+    for (const library of this.libraries) {
+      const found = plexLibraries.find((lib) => lib.key == library.id);
+      if (found && found.agent == HAMA_AGENT) {
+        return true;
+      }
+    }
+    return false;
```
sct, line 474: Strict comparison and some optional chaining
```suggestion
    const plexLibraries = await this.plexClient.getLibraries();
    for (const library of this.libraries) {
      const found = plexLibraries.find((lib) => lib.key === library.id);
      if (found?.agent === HAMA_AGENT) {
        return true;
      }
    }
    return false;
```

I would probably write this code like this though. (Wrote this in GitHub so may not be syntactically correct)
```suggestion
    const plexLibraries = await this.plexClient.getLibraries();
    return this.libraries.some((library) => plexLibraries.some((plexLibrary) => plexLibrary.agent === HAMA_AGENT && library.id === plexLibrary.key));
```

server/job/plexsync/index.ts
```diff
@@ -182,6 +231,68 @@ class JobPlexSync {
...
+              if (extResponse.tv_results[0]) {
+                tvShow = await this.tmdb.getTvShow({
+                  tvId: extResponse.tv_results[0].id,
+                });
+              }
+              await this.processHamaSpecials(plexitem, metadata, result.tvdbId);
+            }
+
+            // if getting tvShow above failed, try with tmdb/imdb for movie
+            if (tvShow == null) {
```
sct, line 276: Instead just check if its truthy here. I believe `tvShow` here is undefined, not null, right? And those are treated differently in javascript. This is where strict comparison is important because you can easily have truthy values you don't expect without it. Your case here would have been fine, since undefined would come back truthy here with a non-strict comparison to null, but its unpredictable code.
```suggestion
            if (!tvShow) {
```

sct (changes requested):
This looks really great. Thanks so much. I just have a few little nitpicks for you but nothing too major!

server/api/animelist.ts
mmozeiko: I was kind of afraid you'll ask for this... I tried to do that initially, but I failed terrible due to not knowing typescript so I gave up. Let me try again.

server/job/plexsync/index.ts
mmozeiko: Majority of movies in mapping won't have tvdbid set, so previous lookup won't execute.
Only few movies have that - typically they are for anime that has tv show also (like Naruto).
I did tvdbid lookup first - then the problem is t hat your api calls to tmdb do not have code for TV lookup in tmdb. So tmdb lookup by tmdb id they would fail for tv shows - as mapping has tmdbid for tv show too. So instead of failing for majority tv shows lookups first by tmdbid in tmdb by and then succeeding by tvdbid in tmdb, I first check for tv show by tvdbid in tmdb - and only then for minority of movies first fail for tvdbid lookup in tmdb, and then succed by tmdbid lookup.

server/api/animelist.ts
sct: Bug me in discord if you need any help with it!

server/job/plexsync/index.ts
mmozeiko: tvShow is declared on line 38 to be null'able.

server/job/plexsync/index.ts
sct: Got it 👍

sct (approved):
I think we are good to try this in `develop`!  Rebase and we can merge!

PR #563, opened 2021-01-03
pre-populate server info from plex.tv API

### LINKED ISSUE #289: Use the Plex API to pre-populate the server information
Plex.tv has an API call that can be made to get a list of servers a user has access to. Pulling this list would make the initial setup for users much easier since it is the same method the players use to discover the server. 

Use the Plex API to pre-populate the Plex server information. 

Additional details can be found here. 
https://support.plex.tv/articles/206721658-using-plex-tv-resources-information-to-troubleshoot-app-connections/

chriscpritchard:
#### Description
Adds the ability to automatically configure plex servers from the plex.tv API.

image
image
image
image

#### Todos

- [X] Add API endpoint
- [X] Add form field in plex settings
- [X] Display available servers in form field
- [X] Allow selection of server retrieved via API
- [X] Complete API documentation for swagger
- [X] Check status of servers

- Fixes #289


samwiseg0:
We should only present the servers the admin owns. `owned="1"`

chriscpritchard:
Functional and should behave as expected, would welcome some testing prior to merge.

server/api/plextv.ts
```diff
@@ -87,6 +126,73 @@ class PlexTvAPI {
...
+          const connArr: PlexConnection[] = [];
+          pxml?.Connection?.forEach((conn: ConnectionResponse) => {
+            connArr.push({
+              protocol: conn.$.protocol,
+              address: conn.$.address,
+              port: parseInt(conn.$.port, 10),
+              uri: conn.$.uri,
+              local: conn.$.local == '1' ? true : false,
+            });
+          });
```
sct, line 153: You can simplify this with a map. You can do the same with the actual plex device array as well. This is the preferred pattern in the codebase.

```suggestion
          const connArr: PlexConnection[] = pxml?.Connection?.map((conn: ConnectionResponse) => ({
              protocol: conn.$.protocol,
              address: conn.$.address,
              port: parseInt(conn.$.port, 10),
              uri: conn.$.uri,
              local: conn.$.local == '1' ? true : false,
          });
```

server/api/plextv.ts
```diff
@@ -87,6 +126,73 @@ class PlexTvAPI {
...
+            presence: pxml.$?.presence == '1' ? true : false,
+            ownerID: pxml.$?.ownerID,
+            home: pxml.$?.home == '1' ? true : false,
+            sourceTitle: pxml.$?.sourceTitle,
+            connection: connArr,
+          });
+        });
+        return arr;
+      };
+      return devResp();
```
sct, line 186: Why are we wrapping everything in a function and calling it here? Why not just return the array directly?

server/api/plextv.ts
```diff
@@ -87,6 +126,73 @@ class PlexTvAPI {
...
+    } catch (e) {
+      logger.error(
+        `Something went wrong getting the devices from plex.tv: ${e.message}`,
+        { label: 'Plex.tv API' }
```
sct, line 190: I know this is a bit mixed and matched around the app but trying to keep these the same going forward. Instead of the error message in the template literal, add it as a property `errorMessage`:

```suggestion
        'Something went wrong getting the devices from plex.tv',
        { label: 'Plex.tv API', errorMessage: e.message }
```

src/components/Settings/SettingsPlex.tsx
```diff
@@ -223,6 +387,62 @@ const SettingsPlex: React.FC<SettingsPlexProps> = ({ onComplete }) => {
...
+                    htmlFor="preset"
+                    className="block text-sm font-medium leading-5 text-gray-400 sm:mt-px"
+                  >
+                    <FormattedMessage {...messages.serverpreset} />
+                  </label>
+                  <div className="mt-1 sm:mt-0 sm:col-span-2">
+                    <div className="flex max-w-lg rounded-md shadow-sm input-group">
+                      {/* Disables no-onchange as onBlur does not work as well, no-onchange is depreciated see: */}
+                      {/* https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-onchange.md */}
+                      {/* eslint-disable-next-line jsx-a11y/no-onchange*/}
```
sct, line 401: Ah, nice to know this was deprecated. We can diasble the rule in the eslint config then, instead of just here.

src/components/Settings/SettingsPlex.tsx
```diff
@@ -52,12 +62,143 @@ interface SyncStatus {
...
+  address: string;
+  host: string;
+  port: number;
+  local: boolean;
+}
 interface SettingsPlexProps {
   onComplete?: () => void;
 }
 
+function PlexPresetOption(props: { key: string; server: PresetServerDisplay }) {
```
sct, line 78: Use an arrow function to match coding style. You can also deconstruct the props.

```suggestion
const PlexPresetOption = ({ key, server }: { key: string; server: PresetServerDisplay }) => {
```

src/components/Settings/SettingsPlex.tsx
```diff
@@ -52,12 +62,143 @@ interface SyncStatus {
...
+        setIsUpdatingPreset(false);
+      }
+    } else {
+      setIsPresetServer(false);
+      setIsUpdatingPreset(false);
+    }
+  };
+  const { data: dataServers, revalidate: revalidateServers } = useSWR<
+    PlexDevice[]
+  >('/api/v1/settings/plex/devices');
```
sct, line 201: This should be above your methods to filter/check this data.

src/components/Settings/SettingsPlex.tsx
```diff
@@ -52,12 +62,143 @@ interface SyncStatus {
...
+            address: conn.address,
+            port: conn.port,
+            local: conn.local,
+            host: theHost,
+          });
+        });
+      }
+    });
+    return map;
+  };
```
sct, line 128: You are working with dataServers here before its defined. This only works because we call the method /after/ dataServers is set, but it's very confusing.

You should define this method (and the following ones) under your `useSWR` hook and work on the data directly when the state actually changes. You should also use `useMemo` so we only run through the array again if the data is actually different by setting `dataServers` as a dependency. You should just also return the map directly, instead of a function.

src/components/Settings/SettingsPlex.tsx
```diff
@@ -52,12 +62,143 @@ interface SyncStatus {
...
+          srv.host == data?.ip &&
+          srv.port == data.port &&
+          srv.ssl == data?.useSsl
+        ) {
+          resp = srv.uri;
+        }
+      });
+    }
+    return resp;
+  };
```
sct, line 143: If you implement the changes I recommended about returning the Map directly, you can use `availablePresets` directly here as well instead of making a function call which will rebuild the map every time this component re-renders.

src/components/Settings/SettingsPlex.tsx
```diff
@@ -52,12 +62,143 @@ interface SyncStatus {
...
+    return map;
+  };
+  const presetServerSelected = (): string => {
+    let resp = 'manual';
+    if (isPresetServer) {
+      availablePresets().forEach((srv) => {
+        if (
+          srv.host == data?.ip &&
+          srv.port == data.port &&
+          srv.ssl == data?.useSsl
```
sct, line 136: Use strict comparison

```suggestion
          srv.host === data?.ip &&
          srv.port === data.port &&
          srv.ssl === data?.useSsl
```

src/components/Settings/SettingsPlex.tsx
```diff
@@ -52,12 +62,143 @@ interface SyncStatus {
...
+    >();
+    // turns the URI into a usable hostname / IP
+    const regexp = /(http(s?):\/\/)(.*)(:[0-9]*)/;
+    dataServers?.forEach((dev) => {
+      if (dev.provides.includes('server') && dev.owned == true) {
+        dev.connection.forEach((conn) => {
+          const theHost = conn.uri.replace(regexp, '$3');
+          map.set(conn.uri, {
+            name: dev.name,
+            ssl: conn.protocol == 'https' ? true : false,
```
sct, line 117: Strict comparison
```suggestion
            ssl: conn.protocol === 'https' ? true : false,
```

src/components/Settings/SettingsPlex.tsx
```diff
@@ -52,12 +62,143 @@ interface SyncStatus {
...
 
+function PlexPresetOption(props: { key: string; server: PresetServerDisplay }) {
+  if (props.server.local) {
+    return (
+      <option key={props.server.uri} value={props.server.uri}>
+        {' '}
+        {props.server.name}
+        {' ('}
+        {props.server.address}
+        {') [local]'}
```
sct, line 86: Use a template literal here to clean this up (and the one below it). Also, instead of having an if statement to render two different options depending on the state of local, do it all in one line and drop the if statement.
```suggestion
        {`${server.name} (${server.address}) [${server.local ? 'local' : 'remote'}]`}
```

src/components/Settings/SettingsPlex.tsx
```diff
@@ -52,12 +62,143 @@ interface SyncStatus {
...
+  }
+  return (
+    <option key={props.key} value={props.server.uri}>
+      {' '}
+      {props.server.name}
+      {' ('}
+      {props.server.address}
+      {') [remote]'}
+    </option>
+  );
```
sct, line 98: This whole block isn't necessary. Drop the if statement above and do everything in one line.

src/components/Settings/SettingsPlex.tsx
sct: I believe you can also use `useMemo` here with a dependency of `[data, availablePresets]`?

src/components/Settings/SettingsPlex.tsx
```diff
@@ -52,12 +62,143 @@ interface SyncStatus {
...
+    return resp;
+  };
+  const changePresetServer = async (
+    e: React.ChangeEvent<HTMLSelectElement>
+  ) => {
+    setIsUpdatingPreset(true);
+    const targPreset = availablePresets().get(e.target.value);
+    if (targPreset) {
+      setIsPresetServer(true);
+      let toastID: string | undefined = undefined;
```
sct, line 151: By not assigning anything, it's undefined.

```suggestion
      let toastID: string | undefined;
```

src/components/Settings/SettingsPlex.tsx
sct: Actually, all of the hooks should be above your methods you added.

src/components/Settings/SettingsPlex.tsx
```diff
@@ -223,6 +387,62 @@ const SettingsPlex: React.FC<SettingsPlexProps> = ({ onComplete }) => {
...
+                        placeholder={intl.formatMessage(
+                          messages.serverpresetPlaceholder
+                        )}
+                        value={presetServerSelected()}
+                        className="flex-1 block w-full min-w-0 transition duration-150 ease-in-out bg-gray-700 border border-gray-500 rounded-none rounded-l-md form-input sm:text-sm sm:leading-5"
+                        onChange={(e) => changePresetServer(e)}
+                        disabled={isUpdatingPreset}
+                      >
+                        <option value="manual">
+                          {' '}
```
sct, line 414: Don't need this space?

src/components/Settings/SettingsPlex.tsx
```diff
@@ -163,6 +322,7 @@ const SettingsPlex: React.FC<SettingsPlexProps> = ({ onComplete }) => {
           hostname: data?.ip,
           port: data?.port,
           useSsl: data?.useSsl,
+          presetServerCheck: isPresetServer,
```
sct, line 325: What is the point of this?

server/lib/settings.ts
```diff
@@ -16,6 +16,7 @@ export interface PlexSettings {
   ip: string;
   port: number;
   useSsl?: boolean;
+  presetServer?: boolean;
```
sct, line 19: Is this setting actually used for anything?

sct (changes requested):

server/lib/settings.ts
chriscpritchard: Yes, it is, if the server was configured manually, the form is read/write, whereas if it was configured automatically the form becomes read only, unless you change the preset to manual - storing it server side stops people who manually set the server to the plex direct URL from being unable to modify it, however, it only needs to be updated on form submit.

sct: Is there any reason we actually must show the input boxes if they have as preset server selected? We can always just hide them in the case thats its preset and set the formik values manually with `setFieldValue` and then we don't really even need this `presetServer` config option.

sct: When the component renders we can determine again if its one of the preset servers, no? So in that case we just don't render the inputs again.

overseerr-api.yml
```diff
@@ -98,11 +98,128 @@ components:
           readOnly: true
           items:
             $ref: '#/components/schemas/PlexLibrary'
+        manualServer:
+          type: boolean
+          example: true
```
sct, line 103: `presetServer`?

src/components/Settings/SettingsPlex.tsx
```diff
@@ -14,8 +15,17 @@ const messages = defineMessages({
   plexsettings: 'Plex Settings',
   plexsettingsDescription:
     'Configure the settings for your Plex server. Overseerr uses your Plex server to scan your library at an interval and see what content is available.',
-  servername: 'Server Name (Automatically set after you save)',
+  servername: 'Server Name (Retrieved from Plex)',
```
sct, line 18: You will need to update this line manually in `en.json`

src/components/Settings/SettingsPlex.tsx
```diff
@@ -14,8 +15,17 @@ const messages = defineMessages({
...
   servernamePlaceholder: 'Plex Server Name',
+  serverpreset: 'Available Server',
+  serverpresetPlaceholder: 'Plex Server (Retrieved Automatically)',
+  serverpresetManualMessage: 'Manually Configure',
+  toastPlexRefreshSuccess: 'Retrieved server list from Plex',
+  toastPlexRefreshFailure: 'Unable to retrieve server list from Plex',
+  toastPlexConnecting: 'Attempting to connect to Plex server',
+  toastPlexConnectingSuccess: 'Connected to Plex server',
+  toastPlexConnectingFailure: 'Unable to connect to Plex server',
+  presetServerUsed: 'Available server used',
```
sct, line 28: Don't forget to run `yarn i18n:extract`

src/components/Settings/SettingsPlex.tsx
```diff
@@ -84,6 +108,103 @@ const SettingsPlex: React.FC<SettingsPlexProps> = ({ onComplete }) => {
...
 
+  const availablePresets = useMemo(() => {
+    const map: Map<string, PresetServerDisplay> = new Map<
+      string,
+      PresetServerDisplay
+    >();
+    // turns the URI into a usable hostname / IP
+    const regexp = /(http(s?):\/\/)(.*)(:[0-9]*)/;
+    dataServers?.forEach((dev) => {
+      if (dev.provides.includes('server') && dev.owned == true) {
```
sct, line 119:
```suggestion
      if (dev.provides.includes('server') && dev.owned) {
```

server/lib/settings.ts
chriscpritchard: I tried multiple ways to deal with this - because the available presets is loaded after the page loads, the most sensible way is to just default to manual on page load - server side variable has been removed.

sct:
I can't even start this at the moment: Getting this error:

```
{
message: "can't resolve reference #components/schemas/PlexSettings from id #"
}
```

Alos this promise rejection in the console

```
(node:27099) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:27099) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:27099) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 1)```

sct:
Is the intention for it to always be set back to `Manually Configure` after coming back to the page?

sct:
I think the "Timeout" feature needs more context. Maybe a small description that explains what the purpose of the option is.

sct:
Maybe we should rethink the UX for this. Because of the slow load time of the dropdown, maybe we should design this in a way where it's an option to "prefill" the input fields for you. It doesn't have to constantly fetch this every time we go to the plex page (or really ever after they initially set it), right?

sct (approved):

PR #591, opened 2021-01-06
feat(login): add local users functionality

### LINKED ISSUE #335: Add the ability to create local users for sign-in.
Currently only users with a Plex account can sign into Overseerr.

Add the ability to create local users for sign in where the same permissions can be assigned. This would be mostly for local users. In this context, my wife obviously enjoys Plex but doesn't have her own account and my server creds are too complex to expect her to remember.

Local users have to ask the server owner to request a title which defeats the purpose for those without a Plex account. With Ombi, my wife can log in with her own credentials and request titles.

**Additional context**
None

ankarhem:
Heres an album with desktop + mobile screens:
https://imgur.com/a/5NbJhvk

#### Todos

- [x] Use gravatar if it exists ~~and fallback to some img~~ (fallbacks automatically)
- [x] Improve login page styling a bit
~~Add a must-change-password step on first login~~
- [x] Add possibility to send email with generated password
- [x] Move password logic the user entity

- Closes #335


ankarhem:
Not sure how to solve `Add a must-change-password step on first login`.

My initial idea is to add a `mustChangePassword` boolean to the `User` entity and then setting that to true for new users. Then lock routing to `/changepassword` (if authenticated but `mustChangePassword` is true) until password has been changed. What do you think @sct?

sct:
> Not sure how to solve `Add a must-change-password step on first login`.
> 
> My initial idea is to add a `mustChangePassword` boolean to the `User` entity and then setting that to true for new local users. Then lock routing to `/changepassword` (if authenticated but `mustChangePassword` is true) until password has been changed. What do you think @sct? I think this could have some extendability further on with reset password links and all that.

I was thinking about this and I am already predicting people complaining about it being too complicated for family and such to deal with having to actually send the password via email and forcing them to change it. It makes me think of the way google handles passwords for new accounts where you have the choice to either generate it and sent it via email, or manually set it. Maybe an option is a good idea here?

As for implementation of forcing the password change, I would probably do exactly what you said. Have a boolean on the user that forces them to the change password page until they do it.

sct:
> Just a heads-up, I think creating the "must change" feature should be accompanied by a password complexity check, otherwise, the nontech users will just create a 123456 pass, and you know the consequences.

I don't think this will work. A lot of people will argue that their setup is secure and no external access is possible anyways so why do we enforce secure passwords?

costaht:
> A lot of people will argue that their setup is secure and no external access is possible

Unless these people control the source IP of their connections, there's nothing they can do against brute force. Even if the initial password is strong, if the application doesn't enforce a password policy, the 123456 passwords will happen. My instance for example is on the internet so my family that lives abroad can use it. But anyway, just my thoughts.

ankarhem:
> It makes me think of the way google handles passwords for new accounts where you have the choice to either generate it and sent it via email, or manually set it. Maybe an option is a good idea here?

This would be probably be more simple to implement. I'll take a jab at this when I get off work.

ankarhem:
This is what the email notification will look like + logs
image

ankarhem:
Right now this requires that the smtp settings are setup for email notifications. It's a bit odd I think. Maybe we should extract the smtp settings from the email notifications to its own section under general settings? We could instead reference that this needs to be setup in an alert on email notifications.

ankarhem:
Also, I'm thinking maybe the generation and setting of the password should belong to the user entity. What do you think? @sct 

I'm thinking it could be two methods `resetPassword()` and `setPassword(password: string)`. `setPassword` sets the user's password in hashed format, and `resetPassword` sets the password and sends out an email.

I think this will be more extendable to be able to reset the password for a user at any time via a button in userlist for instance.

src/components/Login/LocalLogin.tsx
```diff
@@ -0,0 +1,152 @@
...
+                  className="block my-1 text-sm font-medium leading-5 text-gray-400 sm:mt-px"
+                >
+                  {intl.formatMessage(messages.password)}
+                </label>
+                <div className="mt-1 mb-2 sm:mt-0 sm:col-span-2">
+                  <div className="flex max-w-lg rounded-md shadow-sm">
+                    <Field
+                      id="password"
+                      name="password"
+                      type="password"
```
sct, line 102: should this be a password field? I feel like we want to see what we are typing here because it doesn't have a confirmation and it's for another user. What do you think?

server/routes/user.ts
```diff
@@ -19,13 +20,33 @@ router.get('/', async (_req, res) => {
 
 router.post('/', async (req, res, next) => {
   try {
+    const settings = getSettings().notifications.agents.email;
+    if (!settings.enabled) {
+      throw new Error('Email notifications must be enabled');
+    }
```
sct, line 26: Even if I am not using an auto-generated password, it fails here.

src/components/Login/LocalLogin.tsx
```diff
@@ -0,0 +1,152 @@
...
+                        e.preventDefault();
+                        goBack();
+                      }}
+                    >
+                      {intl.formatMessage(messages.goback)}
+                    </Button>
+                  </span>
+                  <span className="inline-flex ml-3 rounded-md shadow-sm">
+                    <Button
+                      buttonType="default"
```
sct, line 133: This should be buttonType `primary`

src/components/Login/LocalLogin.tsx
```diff
@@ -0,0 +1,152 @@
...
+        touched,
+        isSubmitting,
+        // values,
+        isValid,
+        // setFieldValue,
+      }) => {
+        return (
+          <>
+            <Form>
+              <div className="mt-6 sm:mt-5 sm:border-t sm:border-gray-800">
```
sct, line 70: The margin top here is not necessary as the parent container already has sufficient padding for the form.

src/components/PlexLoginButton/index.tsx
```diff
@@ -38,7 +38,7 @@ const PlexLoginButton: React.FC<PlexLoginButtonProps> = ({
     }
   };
   return (
-    <span className="block w-full rounded-md shadow-sm">
+    <span className="block w-full pb-4 rounded-md shadow-sm">
```
sct, line 41: This component shouldn't have this padding. It shouldn't be concerned with where or how it's used. Instead, wrap the component with a div where you use it and add the padding there.

src/components/UserList/index.tsx
```diff
@@ -149,16 +183,178 @@ const UserList: React.FC = () => {
...
+        enterFrom="opacity-0"
+        enterTo="opacity-100"
+        leave="opacity-100 transition duration-300"
+        leaveFrom="opacity-100"
+        leaveTo="opacity-0"
+        show={createModal.isOpen}
+      >
+        <Modal
+          title={intl.formatMessage(messages.createuser)}
+          iconSvg={<AddUserIcon className="h-6" />}
```
sct, line 197: This modal needs `onCancel` and `onPrimary` instead of rendering the buttons yourself. Look at other uses of the modal component. (Also as a side note, the create button should be button style of primary, as well)

sct (changes requested):

src/components/Login/LocalLogin.tsx
ankarhem: I assume you meant to referr to the password field in the create modal and not on the login screen. Either way I think it should be a password field.

You might have other people who can see what you're doing on the screen that are not supposed to know the password. I also think it needs to be a password field in order to be correctly configured for password managers and the like.

Keeping it like it is for now and maybe if anyone has any objections this discussion can be had in an issue.

server/migration/1610070934506-LocalUsers.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE "temporary_user" ("id" integer PRIMARY KEY AUTOINCREMENT NOT NULL, "email" varchar NOT NULL, "username" varchar NOT NULL, "plexId" integer NOT NULL, "plexToken" varchar, "permissions" integer NOT NULL DEFAULT (0), "avatar" varchar NOT NULL, "createdAt" datetime NOT NULL DEFAULT (datetime('now')), "updatedAt" datetime NOT NULL DEFAULT (datetime('now')), "password" varchar, "userType" integer NOT NULL DEFAULT (1), CONSTRAINT "UQ_e12875dfb3b1d92d7d7c5377e22" UNIQUE ("email"))`
+    );
+    await queryRunner.query(
+      `INSERT INTO "temporary_user"("id", "email", "username", "plexId", "plexToken", "permissions", "avatar", "createdAt", "updatedAt") SELECT "id", "email", "username", "plexId", "plexToken", "permissions", "avatar", "createdAt", "updatedAt" FROM "user"`
+    );
+    await queryRunner.query(`DROP TABLE "user"`);
+    await queryRunner.query(`ALTER TABLE "temporary_user" RENAME TO "user"`);
```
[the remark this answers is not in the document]
ankarhem: I don't know if I'm doing anything wrong but my request is removed after running the migration. Can you try?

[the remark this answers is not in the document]
ankarhem: Did this:
checked out develop
deleted db
changed orm conf
ran migrations
imported users and req a movie with another account
checked out local-user
ran migrations

this resulted in the request being removed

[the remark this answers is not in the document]
sct: In the `ormconfig` you only want to turn off sync, not turn on migrations. We run migrations manually in `server/index.ts`.

```javascript
    // Run migrations in production
    if (process.env.NODE_ENV === 'production') {
      await dbConnection.query('PRAGMA foreign_keys=OFF');
      await dbConnection.runMigrations();
      await dbConnection.query('PRAGMA foreign_keys=ON');
    }
```

We also skip running them if the environment isn't production. So comment out the if statement around the migration here just temporarily to make sure it runs.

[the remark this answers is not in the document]
sct: If you turn on migrations in the `ormconfig` then it will not turn off foreign_keys first, which results in all relations being lost.

src/components/Login/LocalLogin.tsx
sct: Sorry yes, I meant the creation modal. My mistake. In the future let's add a checkbox that toggles showing the password.

Just for note, you don't need a password field for password managers, you instead can use the html attribute `autocomplete`. See: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

Which makes me think... maybe let's add this attribute to the login page?

server/migration/1610070934506-LocalUsers.ts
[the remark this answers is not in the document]
ankarhem: So the original migration lines where not abundant. Without them the migration was incomplete and caused existing users to show up as local users. I've readded the generated migration, which I've tested and it seems to work and preserves requests.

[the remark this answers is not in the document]
sct: Oh sorry, my mistake. I misread the migrations.

src/components/Login/LocalLogin.tsx
ankarhem: I think we can add it to the list of follow up stuff to be done, so that this gets out there and maybe we can get additional feedback from the community. After this is merged I'll create an issue for them.

sct: Sounds good.

sct (approved):
LGTM!

PR #597, opened 2021-01-07
feat: ability to edit user settings in bulk

### LINKED ISSUE #492: Bulk User Permisson Edit Button
With the recent addition of separated auto approve permissions for Movies/Series. I have to manually edit everyone permission to update them, so they have auto approve movies option. This takes a lot of time and can be easily fixed with a bulk edit button

Under the User tab, add a Bulk Edit Button. With this, the server admin can select everyone or a selected few and edit their permission.

Ombi User Management Page

**Additional context**

Screenshot_226

hmnd:
#### Description
Allow editing multiple users' settings at the same time in Users table.

35rf1Eii
tSfIds3l

#### Todos

- Fixes #492


hmnd:
Btw not sure whether it's alright to change key names in the localization files due to refactoring a component out, so let me know if that's wrong.

sct:
> Btw not sure whether it's alright to change key names in the localization files due to refactoring a component out, so let me know if that's wrong.

It _should_ be fine but weblate tends to get have issues with merge conflicts when this happens. Maybe just modify only the `en.json` file and let the translators migrate the translations over.

sct:
@hmnd This PR is really fantastic! I have a few comments I will leave in the review but I also wanted to point out a design choice that I don't necessarily agree with.

https://user-images.githubusercontent.com/234213/104017633-d1986a00-51fb-11eb-9217-6c4ae86320b8.mov

I really don't like how the whole table shifts here when clicking on this button. It can easily cause misclicks? And the effect is really jarring. What about moving the button to the right side of the table above the normal actions? You can even get away I think with using `buttonSize="sm"` there to make it fit pretty nicely!

Screen Shot 2021-01-08 at 21 54 10

What do you think?

src/components/UserList/index.tsx
```diff
@@ -175,6 +256,20 @@ const UserList: React.FC = () => {
         <Table.TBody>
           {data?.map((user) => (
             <tr key={`user-list-${user.id}`}>
+              <Table.TD>
+                {user.id !== currentUser?.id && (
```
sct, line 260: What do you think about not allowing modification of user id 1 here either? (And in the logic to select all users)

hmnd:
> I really don't like how the whole table shifts here when clicking on this button. It can easily cause misclicks? And the effect is really jarring. What about moving the button to the right side of the table above the normal actions? You can even get away I think with using `buttonSize="sm"` there to make it fit pretty nicely!

Totally agree. I couldn't think of a better place to put it that wouldn't move everything, but I think your idea is great!

hmnd:
@sct Hmm how about one of these maybe? I don't really love how the button looks in the table header on the right. Imo it doesn't align well with the bigger buttons below it and isn't noticeable unless you scroll over to the side on screen widths < ~1400px.
image
image

Magikarplvl4:
@hmnd Maybe "Bulk Edit" instead of only "edit"?

sct:
Screen Shot 2021-01-09 at 19 03 16

I think it looks fine on the right? It's also where all the other edit buttons are. Yes, at lower resolutions that area gets cut off but the expected place for the edit buttons remains the same.

sct:
I do like @Magikarplvl4's idea of having it say Bulk Edit:
Screen Shot 2021-01-09 at 20 11 48

hmnd:
@sct Yeah I think that works well! The button disappearing and reappearing was what put me off, but it works much better disabled like that. Will make those changes soon. :)

ankarhem:
Are you still working on this?

hmnd:
@ankarhem yup, working on rebasing now :)

hmnd:
@sct should be ready for your review again! Sorry for the wait.

sct (approved):

rcohen-unext:
@all-contributors please add @hmnd for code

PR #651, opened 2021-01-14
fix(user edit): fix user edit not being able to be saved

ankarhem:
The local user commit introduced a bug that made you unable to save edited users

also fixed another error that was on the user edit page

### DIFF
--- a/src/components/UserEdit/index.tsx
+++ b/src/components/UserEdit/index.tsx
@@ -79,7 +79,6 @@ const UserEdit: React.FC = () => {
       await axios.put(`/api/v1/user/${user?.id}`, {
         permissions: currentPermission,
         email: user?.email,
-        avatar: user?.avatar,
       });
 
       addToast(intl.formatMessage(messages.usersaved), {
--- a/overseerr-api.yml
+++ b/overseerr-api.yml
@@ -47,7 +47,6 @@ components:
             $ref: '#/components/schemas/MediaRequest'
       required:
         - id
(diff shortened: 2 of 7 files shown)


src/components/UserEdit/index.tsx
```diff
@@ -80,6 +80,7 @@ const UserEdit: React.FC = () => {
         permissions: currentPermission,
         email: user?.email,
         avatar: user?.avatar,
+        userType: user?.userType,
```
sct, line 83: I don't think we should pass user type. It shouldn't be editable by the user anyways. We should instead change the user route on the server end to merge the actual editable values being submitted.

sct (changes requested):

src/components/UserEdit/index.tsx
ankarhem: Don't the same apply for email and avatar though?

sct: Avatar yes, email will be editable for local users. So no.

At some point we could also make avatars editable but not very important.

ankarhem: Hm ok. I see your point then.

The issue is not that it isn't merging the result I think, but that the User schema lists userType as a required parameter.

sct: Oh, yeah. Ok let's get this merged for now and we can refactor it later.

sct (approved):

PR #657, opened 2021-01-17
fix(requests): handle when tvdbid is null

### LINKED ISSUE #562: No frontend handling of requests where tvdbId is null
**Describe the bug**
Sending a payload with tvdbId as null when requesting tvseries is unsupported and will result in a `400 Bad Request` but nothing is shown to the user in the frontend.

v1.14.0

**Expected behavior**
Should handle this and show something relevant in the frontend, so that the user knows what's up. Maybe we add a section in the docs that we can link to on how they can resolve it by adding the correct id to tmdb. This way we would encourage improving the resources as well.

ankarhem:
#### Todos

- [x] Handle missing tvdb id's errors and show them in frontend
- [x] Add get endpoint for looking up sonarr data with tmdb id
- [x] Add fallback posters
- [x] Add second step in requestmodal for media missing tvdbid, to confirm manual match
- [x] Translation Keys yarn i18n:extract
- [x] Successfully starts `yarn start` 😉

- Fixes #562


ankarhem:
About the fallback posters, I was thinking maybe we could add it to the actual returned data from the api's instead of optionally at each use case? That would however limit us to using one fallback poster, instead of different ones like it is now. (With / without text). Let me know what you think.

sct:
> About the fallback posters, I was thinking maybe we could add it to the actual returned data from the api's instead of optionally at each use case? That would however limit us to using one fallback poster, instead of different ones like it is now. (With / without text). Let me know what you think.

Should be up to the client to handle missing posters. Otherwise, it's difficult to determine on the client that a poster is actually missing.

sct:
The posters are not loading for me. The divs have 0 height:

image

sct:
Also, there is no loading state while the posters are being fetched. Maybe add the loading spinner into the modal while it's fetching from Sonarr?

src/components/RequestModal/SearchByNameModal/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+                handleClick(item.tvdbId);
+                e.preventDefault();
+              }
+            }}
+            role="link"
+            tabIndex={0}
+          >
+            <div
+              className={`border bg-gray-600 h-40 overflow-hidden w-full flex items-center p-2 rounded-xl shadow ${
+                tvdbId === item.tvdbId ? '' : 'border-transparent'
```
sct, line 91: Instead of using border, let's use rings. see: https://tailwindcss.com/docs/ring-color

Border of white doesn't look fantastic either? Ring color of indigo 500 or 400 might look better.

sct (approved):

PR #661, opened 2021-01-17
docs: Update LE/NGINX to remove CSP [skip ci]

### LINKED ISSUE #660: Local User Gravatar Image Doesn't Load
**Describe the bug**
Local User gravatar image doesn't load due to content security policies:

`Refused to load the image 'https://gravatar.com/avatar/131c92b4a38ed6810bb9984c0a493a92?default=mm&size=200' because it violates the following Content Security Policy directive: "img-src 'self' data: https://plex.tv https://assets.plex.tv https://secure.gravatar.com https://i2.wp.com https://image.tmdb.org".`

develop-d9919abb8998d28558ddec35b8e60ab2af75d5b7

**To Reproduce**

1. Add Local User

**Expected behavior**
Show default gravatar image

**Screenshots**
image

hirenshah:
#### Description
Adding https://gravatar.com to Content-Security-Policy

- Fixes #660

### DIFF
--- a/docs/extending-overseerr/reverse-proxy-examples.md
+++ b/docs/extending-overseerr/reverse-proxy-examples.md
@@ -112,8 +112,8 @@ server {
     add_header Referrer-Policy "no-referrer";
     # HTTP Strict Transport Security
     add_header Strict-Transport-Security "max-age=63072000; includeSubDomains" always;
-    # Reduce XSS risks (Content-Security-Policy) - add URLs whenever necessary
-    add_header Content-Security-Policy "default-src 'self'; connect-src 'self' https://plex.tv; style-src 'self' 'unsafe-inline' https://rsms.me/inter/inter.css; script-src 'self'; img-src 'self' data: https://plex.tv https://assets.plex.tv https://secure.gravatar.com https://gravatar.com https://i2.wp.com https://image.tmdb.org; font-src 'self' https://rsms.me/inter/font-files/" always;
+    # Reduce XSS risks (Content-Security-Policy) - uncomment to use and add URLs whenever necessary
+    # add_header Content-Security-Policy "default-src 'self'; connect-src 'self' https://plex.tv; style-src 'self' 'unsafe-inline' https://rsms.me/inter/inter.css; script-src 'self'; img-src 'self' data: https://plex.tv https://assets.plex.tv https://gravatar.com https://i2.wp.com https://image.tmdb.org; font-src 'self' https://rsms.me/inter/font-files/" always;
     # Prevent some categories of XSS attacks (X-XSS-Protection)
     add_header X-XSS-Protection "1; mode=block" always;
     # Provide clickjacking protection (X-Frame-Options)
--- a/docs/extending-overseerr/reverse-proxy-examples.md
+++ b/docs/extending-overseerr/reverse-proxy-examples.md
@@ -112,7 +112,7 @@ server {
     add_header Referrer-Policy "no-referrer";
     # HTTP Strict Transport Security
     add_header Strict-Transport-Security "max-age=63072000; includeSubDomains" always;
-    # Reduce XSS risks (Content-Security-Policy)
+    # Reduce XSS risks (Content-Security-Policy) - add URLs whenever necessary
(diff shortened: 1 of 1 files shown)


samwiseg0:
I don't like that we will have to keep this list up to date every time we pull from a new source? Let alone every user will have to update their config when this changes. 

I think we should remove it entirely and leave it up to the user if they want to spend the time figuring this out to add it.

hirenshah:
> I don't like that we will have to keep this list up to date every time we pull from a new source? Let alone every user will have to update their config when this changes.
> 
> I think we should remove it entirely and leave it up to the user if they want to spend the time figuring this out to add it.

Are you suggesting that the reverse proxy example be removed completely?

docs/extending-overseerr/reverse-proxy-examples.md
```diff
@@ -112,7 +112,7 @@ server {
     add_header Referrer-Policy "no-referrer";
     # HTTP Strict Transport Security
     add_header Strict-Transport-Security "max-age=63072000; includeSubDomains" always;
-    # Reduce XSS risks (Content-Security-Policy)
+    # Reduce XSS risks (Content-Security-Policy) - add URLs whenever necessary
     add_header Content-Security-Policy "default-src 'self'; connect-src 'self' https://plex.tv; style-src 'self' 'unsafe-inline' https://rsms.me/inter/inter.css; script-src 'self'; img-src 'self' data: https://plex.tv https://assets.plex.tv https://secure.gravatar.com https://gravatar.com https://i2.wp.com https://image.tmdb.org; font-src 'self' https://rsms.me/inter/font-files/" always;
```
samwiseg0, line 116: I think we can leave it up to date with:
```
add_header Content-Security-Policy "default-src 'self'; connect-src 'self' https://plex.tv; style-src 'self' 'unsafe-inline' https://rsms.me/inter/inter.css; script-src 'self'; img-src 'self' data: https://plex.tv https://assets.plex.tv https://secure.gravatar.com https://gravatar.com https://i2.wp.com https://image.tmdb.org; font-src 'self' https://rsms.me/inter/font-files/" always;
```
Maybe comment out the line so it is not enabled by default? 

Put a note in there such as:
This list is subject to change. If you enable this please add any missing entries.

samwiseg0:
> Are you suggesting that the reverse proxy example be removed completely?

No. I think we should not enable that line by default as it will break with future updates. If the user wants to enable it and understands they may have to update it periodically then I think that will be ok.

docs/extending-overseerr/reverse-proxy-examples.md
hirenshah: I had updated the comment to include "add URLs whenever necessary" which is effectively what you are suggesting.

I don't think this example should stop being updated because it is intended as a guide and therefore needs to be up to date.


You mentioned:
> every user will have to update their config when this changes

Surely that is a given? Those who are using a reverse proxy would **always** have to update their configuration when a new source is utilised.  Updating the guides helps to communicate that more effectively. Also, new URLs may be a common occurrence now but it should taper off as the project matures.

It raises another question though, why is both https://secure.gravatar.com and https://gravatar.com necessary? Should the local user gravatar be changed to point to https://secure.gravatar.com instead?

hirenshah:
> > Are you suggesting that the reverse proxy example be removed completely?
> 
> No. I think we should not enable that line by default as it will break with future updates. If the user wants to enable it and understands they may have to update it periodically then I think that will be ok.

The line wouldn't break future updates... it would just result in more errors than seen previously.

samwiseg0 (approved):
Thanks!

sct:
@all-contributors please add @hirenshah for docs

docs/extending-overseerr/reverse-proxy-examples.md
douglasparker: I authored the LE/NGINX configuration, and using `https://gravatar.com` should be fine. I've been using it this way in both Chromium and Firefox based browsers. It may have been that I originally had only `https://secure.gravatar.com`, which would have been an error on my part.

@hirenshah, what browser were you using when you encountered issue #660?

PR #682, opened 2021-01-18
feat(api): /request/count endpoint

### LINKED ISSUE #678: /request/count API endpoint
I use [Heimdall](https://heimdall.site/) as a landing page / dashboard for my domain, and it'd be nice if Overseerr had an API endpoint that returned the number of pending, approved, and available requests so that I could create an "Enhanced App" for Overseerr that displays live stats.

Ombi provides the following, and it would be great if Overseerr could do the same or something similar:
image

N/A

**Additional context**
N/A

TheCatLady:
#### Description
Adds `/request/count` endpoint which returns count of pending, approved, and available requests.

#### Screenshot
image

- Fixes #678

### DIFF
--- a/server/routes/request.ts
+++ b/server/routes/request.ts
@@ -236,14 +236,10 @@ requestRoutes.get('/count', async (_req, res, next) => {
     const approvedCount = await requestRepository.count({
       status: MediaRequestStatus.APPROVED,
     });
-    const availableCount = await requestRepository.count({
-      status: MediaRequestStatus.AVAILABLE,
-    });
 
     return res.status(200).json({
       pending: pendingCount,
       approved: approvedCount,
-      available: availableCount,
     });
   } catch (e) {
     next({ status: 500, message: e.message });
--- a/server/routes/request.ts
+++ b/server/routes/request.ts
@@ -221,10 +221,34 @@ requestRoutes.post(
 
       next({ status: 500, message: 'Invalid media type' });
     } catch (e) {
-      next({ message: e.message, status: 500 });
+      next({ status: 500, message: e.message });
     }
   }
 );
+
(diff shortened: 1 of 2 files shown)


server/routes/request.ts
```diff
@@ -221,11 +221,35 @@ requestRoutes.post(
...
+  try {
+    const pendingCount = await requestRepository.count({
+      status: MediaRequestStatus.PENDING,
+    });
+    const approvedCount = await requestRepository.count({
+      status: MediaRequestStatus.APPROVED,
+    });
+    const availableCount = await requestRepository.count({
+      status: MediaRequestStatus.AVAILABLE,
+    });
```
sct, line 241: This won't give you what you want. We don't actually use the request status of available for anything. You will need to do a manual query for requests that have a media relation thats available.

TheCatLady: Ah, maybe I should just remove this.  I was trying to match the Ombi API response, but I don't feel like the available request count is particularly useful anyway.

sct: You can also use the `results` from the pageinfo on the normal request list to get you this sorta data. But as of right now, that endpoint as the same problem with `available` items.

If you want to get the # of available items, regardless of whether or not there was a request, you can always hit the `/media` endpoint.

TheCatLady: I honestly wasn't going to use the available count anyway, and when you think about it, it's just an arbitrary number that is between 0 and the number of items in your Plex library.  Especially for those of us who used other request systems before.  I don't feel like the available count is particularly useful so I removed it in 469c31a.  Thanks for your quick responses & feedback!

sct (approved):
LGTM

sct:
@all-contributors please add @TheCatLady for code and translation

Issue #646, opened 2021-01-14
Follow up work on local users


ankarhem:
**Describe the solution you'd like**
- Improve create user modal. Hide/show password, move generate password checkbox etc.
- Add correct autocomplete to login page form fields. Maybe `new-password` to the user create modal?
- Add reset password button somewhere to the edit user page
- Somehow lessen the "importance" of local login, so that it has to be a deliberate action to choose it. (look at how tautulli handles it etc.)

LEFT TODO:

- [x] Reset password flow
~~- [ ] Welcome email template~~
~~- [ ] Improve create modal~~

sct:
A few other must-haves for a user system:

- Reset password flow on the login page. (Request new password reset. click link in email. etc)
- Password changes on user settings page, when we finally have a user settings page

Reset password button on the edit page is fine, but we can also just have a password field the admin can edit at any time if they like.

TRaSH-:
Thnx for this option,
but when creating the user where you need to add a username i'm forced to use a email.
after adding the user i can't edit the username (shown).
Would be nice if i could edit the username

costaht:
- Save language preference

hirenshah:
> Thnx for this option,
> but when creating the user where you need to add a username i'm forced to use a email.
> after adding the user i can't edit the username (shown).
> Would be nice if i could edit the username

Maybe have the email address as the username but then have a separate field as the Display Name?

costaht:
> > Thnx for this option,
> > but when creating the user where you need to add a username i'm forced to use a email.
> > after adding the user i can't edit the username (shown).
> > Would be nice if i could edit the username
> 
> Maybe have the email address as the username but then have a separate field as the Display Name?

I like the first option better. Regular username instead of email as username.

costaht:
It seems to be an inconsistency:

![image](https://user-images.githubusercontent.com/50637431/105082698-f1ccf200-5a61-11eb-9d35-a357a2b6b021.png)

ankarhem:
Extracting the last two points as I want this closed and the last two tasks seem less important and could be prioritized by themselves.

PR #709, opened 2021-01-20
refactor(login): redesign login page

ankarhem:
#### Description
add accordion component
redesign login page to use accordions with animation
one of the tasks on #646

Looks and functions like this, with the exception that `Sign in` is used everywhere instead of login on the local user submit button
https://streamable.com/sunpui

#### Todos


TheCatLady:
"Login" is a noun whereas "log in" is an action/verb.  Maybe you can correct that here as well?

ankarhem:
> "Login" is a noun whereas "log in" is an action/verb. Maybe you can correct that here as well?

Had no idea! Thanks :+1: 

I'll make the change

src/components/Common/Accordion/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+  AccordionContent: any;
+}
+
+export const AccordionContent: React.FC<{ isOpen: boolean }> = ({
+  isOpen,
+  children,
+}) => {
+  return (
+    <AnimateHeight height={isOpen ? 'auto' : 0}>
+      {children || <></>}
```
sct, line 25: Can we not just have this as the following?

```suggestion
      {children}
```

src/components/Login/LocalLogin.tsx
```diff
@@ -11,17 +11,15 @@ const messages = defineMessages({
   validationemailrequired: 'Not a valid email address',
   validationpasswordrequired: 'Password required',
   loginerror: 'Something went wrong when trying to sign in',
-  loggingin: 'Logging in...',
-  login: 'Login',
-  goback: 'Go back',
+  signingin: 'Signing in...',
```
sct, line 14: Use Unicode ellipsis:
```suggestion
  signingin: 'Signing in…',
```

src/components/PlexLoginButton/index.tsx
```diff
@@ -3,9 +3,9 @@ import PlexOAuth from '../../utils/plex';
 import { defineMessages, useIntl } from 'react-intl';
 
 const messages = defineMessages({
-  loginwithplex: 'Login with Plex',
+  signinwithplex: 'Sign in',
   loading: 'Loading...',
-  loggingin: 'Logging in...',
+  signingin: 'Signing in...',
```
sct, line 8:
```suggestion
  signingin: 'Signing in…',
```

sct (approved):

PR #711, opened 2021-01-21
feat: lookup movie by imdbid if tmdbid does not exits for plex movie agent

mmozeiko:
#### Description

In case movie for Plex movie agent library does not have tmdbid - try looking it up by imdbid. 

No UI changed

#### Todos

afaik, no issue for this

### DIFF
--- a/server/job/plexsync/index.ts
+++ b/server/job/plexsync/index.ts
@@ -89,6 +89,12 @@ class JobPlexSync {
             newMedia.tmdbId = Number(tmdbMatch);
           }
         });
+        if (newMedia.imdbId && !newMedia.tmdbId) {
+          const tmdbMovie = await this.tmdb.getMovieByImdbId({
+            imdbId: newMedia.imdbId,
+          });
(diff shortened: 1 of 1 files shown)


server/job/plexsync/index.ts
```diff
@@ -89,6 +89,12 @@ class JobPlexSync {
             newMedia.tmdbId = Number(tmdbMatch);
           }
         });
+        if (newMedia.imdbId && !newMedia.tmdbId) {
+          const tmdbMovie = await this.tmdb.getMovieByImdbId({
+            imdbId: newMedia.imdbId,
+          });
+          newMedia.tmdbId = tmdbMovie.id;
+        }
```
sct, line 97: What about one more check following this to make sure we actually got the tmdb id? Just in case we couldn't find a match for some reason?

mmozeiko: Added code to throw error in case tmdb id is still unknown after these checks.

sct (approved):
Great. Thanks!

PR #715, opened 2021-01-21
feat(users): add editable usernames

### LINKED ISSUE #264: Ability to set friendly names for users, Alias
It would be ideal to be able to set friendly names per user as it is difficult to remember every user's username.

The ability to set friendly names for each user.

ankarhem:
#### Description
Part of the follow up work for local users #646 

#### Todos

Closes #264

### DIFF
--- a/server/entity/User.ts
+++ b/server/entity/User.ts
@@ -6,6 +6,7 @@ import {
   UpdateDateColumn,
   OneToMany,
   RelationCount,
+  AfterLoad,
 } from 'typeorm';
 import { Permission, hasPermission } from '../lib/permissions';
 import { MediaRequest } from './MediaRequest';
@@ -25,14 +26,19 @@ export class User {
 
   static readonly filteredFields: string[] = ['plexToken', 'password'];
 
+  public displayName: string;
+
   @PrimaryGeneratedColumn()
   public id: number;
 
   @Column({ unique: true })
   public email: string;
 
-  @Column()
-  public username: string;
+  @Column({ nullable: true })
+  public plexUsername: string;
+
+  @Column({ type: 'text', nullable: true })
+  public username?: string;
 
   @Column({ nullable: true, select: false })
   public password?: string;
@@ -124,5 +130,10 @@ export class User {
         message: e.message,
       });
     }
+  }
+
+  @AfterLoad()
+  public setDisplayName(): void {
+    this.displayName = this.username || this.plexUsername;
   }
 }
--- a/server/entity/User.ts
+++ b/server/entity/User.ts
@@ -37,7 +37,7 @@ export class User {
   @Column({ nullable: true })
   public plexUsername: string;
 
-  @Column({ type: 'text', nullable: true })
+  @Column({ nullable: true })
   public username?: string;
 
   @Column({ nullable: true, select: false })
--- a/server/routes/user.ts
+++ b/server/routes/user.ts
@@ -138,7 +138,11 @@ router.put<{ id: string }>('/:id', async (req, res, next) => {
       });
     }
 
-    Object.assign(user, req.body);
+    Object.assign(user, {
+      username: req.body.username,
+      permissions: req.body.permissions,
+    });
+
     await userRepository.save(user);
 
     return res.status(200).json(user.filter());
@@ -214,19 +218,26 @@ router.post('/import-from-plex', async (req, res, next) => {
     for (const rawUser of plexUsersResponse.MediaContainer.User) {
       const account = rawUser.$;
       const user = await userRepository.findOne({
-        where: { plexId: account.id },
+        where: { email: account.email },
       });
       if (user) {
         // Update the users avatar with their plex thumbnail (incase it changed)
         user.avatar = account.thumb;
         user.email = account.email;
-        user.username = account.username;
+        user.plexUsername = account.username ?? account.email;
+
+        // in-case the user was previously a local account
+        if (user.userType === UserType.LOCAL) {
+          user.userType = UserType.PLEX;
+          user.username = '';
+          user.plexId = parseInt(account.id);
+        }
         await userRepository.save(user);
       } else {
         // Check to make sure it's a real account
         if (account.email && account.username) {
           const newUser = new User({
-            username: account.username,
+            plexUsername: account.username,
             email: account.email,
             permissions: settings.main.defaultPermissions,
             plexId: parseInt(account.id),
--- a/server/routes/user.ts
+++ b/server/routes/user.ts
@@ -226,7 +226,7 @@ router.post('/import-from-plex', async (req, res, next) => {
         // Update the users avatar with their plex thumbnail (incase it changed)
         user.avatar = account.thumb;
         user.email = account.email;
-        user.plexUsername = account.username ?? account.email;
+        user.plexUsername = account.username;
 
         // in-case the user was previously a local account
         if (user.userType === UserType.LOCAL) {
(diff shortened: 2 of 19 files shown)


server/routes/user.ts
```diff
@@ -108,6 +108,28 @@ router.put<{ id: string }>('/:id', async (req, res, next) => {
       });
     }
 
+    // To protect against sql injections via api
+    if (!req.body.username.match(/^[a-zA-Z0-9]+$/)) {
+      return next({
+        status: 500,
+        message: 'Invalid username format',
+      });
+    }
```
sct, line 117: Using the query builder will prevent SQL injection.

sct: I don't think we need to limit usernames so severely. Maybe a max character limit? We should allow spaces/dashes/other special characters if someone wants them.

sct:
Every time a user signs in with their plex account, the username will be overwritten, I believe.

server/routes/user.ts
ankarhem: Alright, I didn't know that. I think we should still limit it so that it doesn't destroy formatting in notifications if they use html/markdown characters in their name. Maybe `[a-zA-Z0-9\-\_]`?

ankarhem:
> Every time a user signs in with their plex account, the username will be overwritten, I believe.

Yeah, but plex users aren't allowed to change their username here. So it should be fine.

sct:
> > Every time a user signs in with their plex account, the username will be overwritten, I believe.
> 
> Yeah, but plex users aren't allowed to change their username here. So it should be fine.

Right, but we still want to handle https://github.com/sct/overseerr/issues/264 so what do you think about doing this in the same PR?

ankarhem:
> Right, but we still want to handle #264 so what do you think about doing this in the same PR?

I thought about it. I think the pros are that we need less migration files. The cons are the PR will grow pretty big as the alias/nickname will need to be conditionally inserted on to every place that use them. Notifications etc. What do you think?

sct:
> > Right, but we still want to handle #264 so what do you think about doing this in the same PR?
> 
> I thought about it. I think the pros are that we need less migration files. The cons are the PR will grow pretty big as the alias/nickname will need to be conditionally inserted on to every place that use them. Notifications etc. What do you think?

What do you think about adding the migration in this PR, so it's out of the way? Then we open a new PR for adding the actual friendly name functionality? We can add a new method to the user entity like a getter for `displayname` that will return the friendly name as a priority if it's set.

ankarhem:
> > > Right, but we still want to handle #264 so what do you think about doing this in the same PR?
> > 
> > 
> > I thought about it. I think the pros are that we need less migration files. The cons are the PR will grow pretty big as the alias/nickname will need to be conditionally inserted on to every place that use them. Notifications etc. What do you think?
> 
> What do you think about adding the migration in this PR, so it's out of the way? Then we open a new PR for adding the actual friendly name functionality? We can add a new method to the user entity like a getter for `displayname` that will return the friendly name as a priority if it's set.

Sure. That sounds reasonable. Getter seems like a better implementation for sure. I'll do it after work tomorrow.

server/routes/user.ts
sct: Spaces as well? `\s`

server/routes/user.ts
```diff
@@ -184,19 +188,26 @@ router.post('/import-from-plex', async (req, res, next) => {
...
         // Update the users avatar with their plex thumbnail (incase it changed)
         user.avatar = account.thumb;
         user.email = account.email;
-        user.username = account.username;
+        user.plexUsername = account.username ?? account.email;
+
+        // in-case the user was previously a local account
+        if (user.userType === UserType.LOCAL) {
+          user.userType = UserType.PLEX;
+          user.username = '';
```
sct, line 202: Do we need to reset the username? We should probably just leave it as whatever it was set to.

server/routes/user.ts
```diff
@@ -184,19 +188,26 @@ router.post('/import-from-plex', async (req, res, next) => {
     for (const rawUser of plexUsersResponse.MediaContainer.User) {
       const account = rawUser.$;
       const user = await userRepository.findOne({
-        where: { plexId: account.id },
+        where: { email: account.email },
```
sct, line 191: We need to keep looking up by `plexId` first, and then `email` if that fails. If the user changes their plex email, it should update their email in Overseerr.

server/routes/user.ts
ankarhem: If you don't want to reset this in all cases, we should at least reset it whenever the username has the same value as email. What do you think? Otherwise local users without usernames set, will not get an eventual username set in plex. Updates to plex username won't be used either.

ankarhem: I've pushed these changes and fixed the other concern you had + rebased. If you think it's fine you should be able to merge. If you want me to fix anything I'll probably be available a few hours more.

sct: I think it should be expected that if a user sets a username for themselves, it won't be overwritten. We can eventually make it clear through the UI that you can clear your username to just use plex's username.

ankarhem:
```          // reset data (as it now resides in plexUsername)
          // if an explicit username has not been set
          if (user.username === user.email) {
            user.username = '';
          }```

It only clears the username field if they haven't had their username edited since local user creation.

Comment (TheCatLady) [reply server/routes/user.ts]: I agree that once set, the username shouldn't be overwritten.  But can we add a reset button of some sort, in case a user wants their username to be synced with Plex again?  Maybe it could be like the button to regenerate the API key in General Settings.

Comment (ankarhem) [reply server/routes/user.ts]: It isn't overwritten and if we want to have this logic more complex I suggest we add it to the backlog.

Comment (TheCatLady) [reply server/routes/user.ts]: @ankarhem Couldn't the button just set the `username` to an empty string, if that signifies that the value should be read from `plexUsername`?

File: server/entity/User.ts [hunk 4]
```diff
@@ -25,14 +26,19 @@ export class User {
...
 
   @Column({ unique: true })
   public email: string;
 
-  @Column()
-  public username: string;
+  @Column({ nullable: true })
+  public plexUsername: string;
+
+  @Column({ type: 'text', nullable: true })
```
sct, line 40: Why the type of `text` here instead of `varchar`?

server/routes/user.ts
```diff
@@ -213,20 +217,38 @@ router.post('/import-from-plex', async (req, res, next) => {
...
-      const user = await userRepository.findOne({
-        where: { plexId: account.id },
-      });
+
+      const user =
+        (await userRepository.findOne({
+          where: { plexId: account.id },
+        })) ??
+        (await userRepository.findOne({
+          where: { email: account.email },
```
sct, line 226: I think you can do this in one query with an array for `where`?

```javascript
where: [
  { plexId: account.id },
  { email: account.email }
]
```

server/entity/User.ts
ankarhem: I don't recall why I wanted to try this. Removed it now.

server/routes/user.ts
```diff
@@ -213,20 +217,34 @@ router.post('/import-from-plex', async (req, res, next) => {
...
-        where: { plexId: account.id },
+        where: [{ plexId: account.id }, { email: account.email }],
       });
+
       if (user) {
         // Update the users avatar with their plex thumbnail (incase it changed)
         user.avatar = account.thumb;
         user.email = account.email;
-        user.username = account.username;
+        user.plexUsername = account.username ?? account.email;
```
sct, line 229: The account username is always set. We don't have to do the fallback ourselves. Although it doesn't really affect anything here to fall back on our own, it will actually never get called.

sct (approved):

PR #776, opened 2021-01-29
build(docker): Optimize Docker image layers/caching & bump node to 14.15

TheCatLady:
#### Description
Since each `RUN` command in `Dockerfile` creates an additional layer, we can combine the installation of `tzdata` and writing of `committag.json` to avoid an unnecessary extra image layer.  🙂

The following `COPY` statements may also be combinable, but I'm not sure if there's anything in `BUILD_IMAGE` which needs to be excluded... so someone else will have to provide some input on that.  Is there a reason why we can't just copy `/app` from `BUILD_IMAGE`?  Could we clean up the `/app` directory in the first stage so we only need one `COPY` command here?

```
COPY . /app
WORKDIR /app

# copy from build image
COPY --from=BUILD_IMAGE /app/dist ./dist
COPY --from=BUILD_IMAGE /app/.next ./.next
COPY --from=BUILD_IMAGE /app/node_modules ./node_modules
```

N/A

#### Todos

- [x] Sucessfully builds `docker build`


ankarhem:
While this is true, you will no longer cache the installation of the package, making it run each time. I think we want to optimize build time and not the (small?) difference it makes to have them share image layer.

TheCatLady:
> While this is true, you will no longer cache the installation of the package, making it run each time. I think we want to optimize build time and not the (small?) difference it makes to have them share image layer.

The `tzdata` package is only 431.7 kB, however (3.36 MB installed).  It would only be a difference of a few seconds, no?

ankarhem:
Yeah, but not cacheing it would create a new layer that includes that space for every new build. Instead of just once? So I don't think that's an argument that would be in favor for this PR? Now that I think of it's probably not cached now anyway since commit tag is over and would be changing?

Dockerfile
```diff
@@ -11,22 +11,25 @@ RUN yarn --frozen-lockfile && \
...
 
+RUN apk add --no-cache tzdata
+
 ARG COMMIT_TAG
 ENV COMMIT_TAG=${COMMIT_TAG}
 
-RUN apk add tzdata
-
-COPY . /app
+COPY [^src]* /app/
```
sct, line 28: Whats the point of doing this if we don't do it for the `server` directory also?

Dockerfile
```diff
@@ -11,24 +11,19 @@ RUN yarn --frozen-lockfile && \
...
 RUN yarn install --production --ignore-scripts --prefer-offline
-RUN yarn cache clean
 
-FROM node:12.18-alpine
+RUN echo "{\"commitTag\": \"${COMMIT_TAG}\"}" > committag.json && \
+  rm -rf src && \
+  rm -rf server
 
-ARG COMMIT_TAG
-ENV COMMIT_TAG=${COMMIT_TAG}
```
sct, line 19: We can't remove these.

Dockerfile
```diff
@@ -11,24 +11,19 @@ RUN yarn --frozen-lockfile && \
 
 # remove development dependencies
 RUN yarn install --production --ignore-scripts --prefer-offline
-RUN yarn cache clean
 
-FROM node:12.18-alpine
+RUN echo "{\"commitTag\": \"${COMMIT_TAG}\"}" > committag.json && \
+  rm -rf src && \
+  rm -rf server
```
sct, line 17: I don't like the idea of combining two entirely different operations into a single step.

TheCatLady: Split into two `RUN` statements now!

Dockerfile
TheCatLady: Ah okay, added them back.

sct (approved):

PR #797, opened 2021-01-31
docs: Update LE/NGINX CSP to fix loading of resources

douglasparker:
There is an issue with the CSP which results in Firefox blocking inline scripts.

I also noticed that Gravatar images were also broken due to a change made in PR #661 to fix issue #660.

Both of these issues are resolved with this PR.


sct:
@all-contributors please add @douglasparker for docs

PR #825, opened 2021-02-03
feat(ui): Add custom title functionality

### LINKED ISSUE #404: Customize the logo / Change the name of Overseerr
One of the simpliest way to customize Ombi as a User nowadays is to replace the Logo and Name of the application (showing up in the tab next to the favicon). I would be nice if the admin could do that in Overseerr as well.

Within the "General Settings" Section, create a subsection named "Customization" (or else) to upload a .png or .jpg for the new logo that will overwrite the existing one. Same thing could be done for a text input where you could write a new name for the application and rename Overseerr.

TheCatLady:
#### Description

This PR adds the option to customize the application title.  In addition, it makes the title reflect the current page (currently, only the movie, series, and collection detail pages have titles w/ more than just "Overseerr").

image

#### Todos

- [x] Successful build
- [x] Translation keys
- [x] Add `PageTitle` shared component

- Partially fixes #404


sct:
Instead of having to add `useSettings`/format the title on every page, why not a shared component that does this? Something like `<PageTitle title="Search" />` which will on its own import `useSettings` and format the title correctly?

TheCatLady:
> Instead of having to add `useSettings`/format the title on every page, why not a shared component that does this? Something like `<PageTitle title="Search" />` which will on its own import `useSettings` and format the title correctly?

Ooh, good point!  I was just copying what existed already on the movie/series/collection detail pages, but it does make sense to have a shared component for this!

src/components/CollectionDetails/index.tsx
```diff
@@ -103,112 +102,38 @@ const CollectionDetails: React.FC<CollectionDetailsProps> = ({
...
-      <Transition
-        enter="opacity-0 transition duration-300"
-        enterFrom="opacity-0"
-        enterTo="opacity-100"
-        leave="opacity-100 transition duration-300"
-        leaveFrom="opacity-100"
-        leaveTo="opacity-0"
-        show={requestModal}
+    <>
+      <PageTitle title={data.name} />
```
sct, line 106: Instead of dropping it here just drop it into the existing JSX so we don't cause every file to be completely indented.

src/components/MovieDetails/MovieSimilar.tsx
```diff
@@ -77,6 +78,11 @@ const MovieSimilar: React.FC = () => {
 
   return (
     <>
+      <PageTitle
+        title={`${intl.formatMessage(messages.similar)}${
+          movieData && ' - ' + movieData.title
```
sct, line 83: This will print undefined if `movieData` evaluates to undefined. Although it shouldn't be undefined, we should build the string outside of this return.

sct: This applies to other files using the same pattern.

src/pages/_app.tsx
```diff
@@ -117,7 +117,6 @@ const CoreApp: Omit<NextAppComponentType, 'origGetInitialProps'> = ({
             <InteractionProvider>
               <ToastProvider components={{ Toast }}>
                 <Head>
-                  <title>Overseerr</title>
```
sct, line 120: This is the default fallback. Why did we get rid of it?

TheCatLady: Ah whoops, I removed it initially to make it easier for me to test something, but I'll add that back.

src/components/MovieDetails/MovieSimilar.tsx
TheCatLady: How about the following?  Or is it still better to define the string outside of the return?
```
      <PageTitle
        title={`${intl.formatMessage(messages.similar)}${
          movieData ? ' - ' + movieData.title : ''
        }`}
      />
```

sct: I would build PageTitle to accept an array OR string. if its an array, it would do something like `title.join(' - ')` and solve this problem :D

src/components/Settings/SettingsMain.tsx
```diff
@@ -160,6 +163,21 @@ const SettingsMain: React.FC = () => {
...
+                    <div className="flex max-w-lg rounded-md shadow-sm">
+                      <Field
+                        id="applicationTitle"
+                        name="applicationTitle"
+                        type="text"
+                        placeholder="Overseerr"
+                      />
+                    </div>
+                  </div>
+                </div>
```
sct, line 180: We should probably add validation so this can't be empty

sct (approved):

PR #826, opened 2021-02-03
feat(docker): Check for /app/config volume mount during setup

TheCatLady:
#### Description

A common issue for users who aren't familiar with Docker is that they find their Overseerr configuration/data wiped after an update/restart of the container.

This PR adds the creation of a zero-byte file to the Docker image in `/app/config`, which would only be visible to the container if the volume mount was not properly configured.

image

#### Todos

- [x] Successful build
- [x] Translation keys

### DIFF
--- a/src/components/Setup/index.tsx
+++ b/src/components/Setup/index.tsx
@@ -11,6 +11,8 @@ import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
 import Badge from '../Common/Badge';
 import LanguagePicker from '../Layout/LanguagePicker';
 import PageTitle from '../Common/PageTitle';
+import { existsSync } from 'fs';
+import Alert from '../Common/Alert';
 
 const messages = defineMessages({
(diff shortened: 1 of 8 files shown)


src/components/Setup/index.tsx
```diff
@@ -23,6 +24,9 @@ const messages = defineMessages({
   tip: 'Tip',
   syncingbackground:
     'Syncing will run in the background. You can continue the setup process in the meantime.',
+  dockerVolumeMissing: 'Docker Volume Mount Missing',
+  dockerVolumeMissingDescription:
+    'The <code>/app/config</code> volume mount was not configured properly. All data will be cleared when the container is stopped or restarted.',
```
sct, line 29: Forgot to remove this?

TheCatLady: Ah whoops, good catch!

sct (approved):

PR #848, opened 2021-02-04
refactor(ui): Display user avatars in request list

TheCatLady:
#### Description

Adds avatars to the request list, just like the ones recently added to request cards!

image

#### Todos

- [x] Successful build
- [x] Translation keys

### DIFF
--- a/src/components/RequestList/RequestItem/index.tsx
+++ b/src/components/RequestList/RequestItem/index.tsx
@@ -27,7 +27,6 @@ import { useToasts } from 'react-toast-notifications';
 import RequestModal from '../../RequestModal';
 
 const messages = defineMessages({
-  requestedby: 'Requested by {username}',
   seasons: 'Seasons',
   notavailable: 'N/A',
   failedretry: 'Something went wrong while retrying the request.',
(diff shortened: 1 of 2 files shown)


src/components/RequestList/RequestItem/index.tsx
```diff
@@ -163,10 +162,15 @@ const RequestItem: React.FC<RequestItemProps> = ({
...
             </Link>
-            <div className="text-sm">
-              {intl.formatMessage(messages.requestedby, {
-                username: requestData.requestedBy.displayName,
-              })}
+            <div className="flex items-center">
+              <img
+                src={requestData.requestedBy.avatar}
+                alt=""
+                className="w-4 mr-1 rounded-full sm:mr-2 sm:w-5"
```
sct, line 169: Do we need responsive sizes here?

src/components/RequestList/RequestItem/index.tsx
```diff
@@ -216,18 +220,24 @@ const RequestItem: React.FC<RequestItemProps> = ({
...
-                  value={Math.floor(
-                    (new Date(requestData.updatedAt).getTime() - Date.now()) /
-                      1000
-                  )}
-                  updateIntervalInSeconds={1}
+              <div className="flex items-center">
+                <img
+                  src={requestData.modifiedBy.avatar}
+                  alt=""
+                  className="w-4 mr-1 rounded-full sm:mr-2 sm:w-5"
```
sct, line 227: Here also. Probably doesn't need to change the avatar size.

src/components/RequestList/RequestItem/index.tsx
TheCatLady: Ah you're right, we don't!  I just copied what you had from the request card, haha.  I'll remove that :)

sct (approved):

PR #851, opened 2021-02-05
fix(ui): Dynamically generate path to config in warning message

TheCatLady:
#### Description

Currently, the path to `/app/config` is hardcoded in the warning message if the volume mount was not configured properly for a Docker install.

This PR dynamically generates the path instead, so that if the path changes in the future or third-party images use a different path, the warning message is accurate/helpful to the user.

#### Todos

- [x] Successful build
- [x] Translation keys

### DIFF
--- a/server/utils/appDataVolume.ts
+++ b/server/utils/appDataVolume.ts
@@ -6,3 +6,7 @@ const DOCKER_PATH = path.join(__dirname, '../../config/DOCKER');
 export const appDataStatus = (): boolean => {
   return !existsSync(DOCKER_PATH);
 };
+
+export const appDataPath = (): string => {
+  return path.join(__dirname, '../../config');
+};
--- a/server/utils/appDataVolume.ts
+++ b/server/utils/appDataVolume.ts
@@ -1,12 +1,16 @@
 import { existsSync } from 'fs';
 import path from 'path';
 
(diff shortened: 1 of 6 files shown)


server/utils/appDataVolume.ts
```diff
@@ -6,3 +6,7 @@ const DOCKER_PATH = path.join(__dirname, '../../config/DOCKER');
 export const appDataStatus = (): boolean => {
   return !existsSync(DOCKER_PATH);
 };
+
+export const appDataPath = (): string => {
+  return path.join(__dirname, '../../config');
```
sct, line 11: I guess an issue here is the assumption the config folder is always located at the root of the app but it can be anywhere, technically. We have an environment variable `CONFIG_DIRECTORY` to change where the config folder is located. Maybe we should check that also here?

TheCatLady: Oh yeah, I wasn't aware of the env var, but that sounds great.  I'll look into this tomorrow or Saturday 😺

TheCatLady: Okay nevermind, my brain wouldn't let me leave this alone for so long.  Fixed!  Also found another location that wasn't using `CONFIG_DIRECTORY` so I fixed that as well.

sct (approved):

Issue #366, opened 2020-12-17
Sonarr v3 Language Profile


Panzer1119:
**Is your feature request related to a problem? Please describe.**
It is not related to a problem.

**Describe the solution you'd like**
I'd like to be able to choose a Sonarr Language Profile when setting up an Instance in the Services Tab.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

soonic6:
I am looking for the same. Sonarr v3 have separate profiles for languages and quality. Request are only for ENG.

johnpyp:
I'll add this for the settings page, though it will may also be desirable to have this be a per-movie setting in the future as well, not sure.

Panzer1119:
So is there any Update on this?

PR #860, opened 2021-02-06
feat(requests): add language profile support

ankarhem:
#### Description
Adds support for default language profiles per sonarr server. Can be overridden in the advanced section on the request modal.

#### Todos

- #366


src/components/RequestModal/AdvancedRequester/index.tsx
```diff
@@ -21,12 +21,15 @@ const messages = defineMessages({
   loadingprofiles: 'Loading profiles…',
   loadingfolders: 'Loading folders…',
   requestas: 'Request As',
+  languageprofile: 'Language Profile',
+  loadinglanguages: 'Loading folders…',
```
TheCatLady, line 25:
```suggestion
  loadinglanguages: 'Loading language profiles…',
```

TheCatLady: Or, we can just use the `loadingProfiles` string that already exists.  I think we should either use the same string for both, or change `loadingProfiles` to `'Loading quality profiles…'`

src/components/Settings/SonarrModal/index.tsx
```diff
@@ -559,6 +577,54 @@ const SonarrModal: React.FC<SonarrModalProps> = ({
                     )}
                   </div>
                 </div>
+                <div className="form-row">
+                  <label
+                    htmlFor="activeLanguageProfileId"
+                    className="text-label"
+                  >
+                    {intl.formatMessage(messages.languageprofile)}
+                    <span className="text-red-500">*</span>
```
sct, line 586: This is marked as required but there is no actual validation on this field.

sct (approved):

PR #875, opened 2021-02-07
docs(api): Fix API endpoint tags

TheCatLady:
#### Description

Currently, the `/auth/me` endpoint is listed twice, and the `auth` endpoints relating to password resets are under the `users` tag.

I thought this was odd and that it must be a mistake, but if it's intentional please just close this PR.

**Before:**
image

**After:**
image

#### Todos

- [x] Successful build


sct:
Intended.

PR #878, opened 2021-02-08
feat(ui): added content ratings for tv shows and movie ratings

### LINKED ISSUE #685: Add content rating of movie and TV series within the details.
No

It would be useful for parents, that like to request media on behalf of their children, to see the content rating of a movie and TV series so they can determine if it is age-appropriate.  I noticed TMDB and IMDb place the content rating underneath the title of the media but maybe it would be best placed on the right where the release date and network studio are located. 

Of course, this is not a big issue as users can easily just search for the content rating on the web, but I think it would be quite a nice feature. 

**Additional context**
https://developers.themoviedb.org/3/tv/get-tv-content-ratings - I found this on TMDB for TV series but nothing for movies as far as I can tell.

danshilm:
#### Description
This adds content ratings for tv shows and movie ratings. The ratings default to either the US when available, or the first one in the response.
I added a border to the rating since it can be unclear as to what '15' is; for example: 
image

This also includes a small fix where the `|` between the runtime and genres on the movie detail page would be displayed even when there are no genres available.

On desktop:
image

On mobile:
image

#### Todos
None

- Fixes #685

### DIFF
--- a/src/components/MovieDetails/index.tsx
+++ b/src/components/MovieDetails/index.tsx
@@ -131,6 +131,16 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
 
   const movieAttributes = [];
 
+  if (data.releases.results.length > 0) {
+    movieAttributes.push(
+      <span className="p-0.5 py-0 border rounded-md">
+        {data.releases.results.find((r) => r.iso_3166_1 === 'US')
+          ?.release_dates[0].certification ||
+          data.releases.results[0].release_dates[0].certification}
+      </span>
+    );
+  }
+
   if (data.runtime) {
     movieAttributes.push(
       intl.formatMessage({ ...messages.runtime }, { minutes: data.runtime })
@@ -369,7 +379,13 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
             )}
           </h1>
           <span className="mt-1 text-xs lg:text-base lg:mt-0">
-            {movieAttributes.join(' | ')}
+            {movieAttributes
+              .map<React.ReactNode>((t, k) => <span key={k}>{t}</span>)
+              .reduce((prev, curr) => (
+                <>
+                  {prev} | {curr}
+                </>
+              ))}
           </span>
         </div>
         <div className="relative z-10 flex flex-wrap justify-center flex-shrink-0 mt-4 sm:justify-end sm:flex-nowrap lg:mt-0">
--- a/src/components/MovieDetails/index.tsx
+++ b/src/components/MovieDetails/index.tsx
@@ -129,9 +129,14 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
     revalidate();
   };
 
-  const movieAttributes = [];
+  const movieAttributes: React.ReactNode[] = [];
 
-  if (data.releases.results.length > 0) {
+  if (
+    data.releases.results.length &&
+    (data.releases.results.find((r) => r.iso_3166_1 === 'US')?.release_dates[0]
+      .certification ||
+      data.releases.results[0].release_dates[0].certification)
+  ) {
     movieAttributes.push(
       <span className="p-0.5 py-0 border rounded-md">
         {data.releases.results.find((r) => r.iso_3166_1 === 'US')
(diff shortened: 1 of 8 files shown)


TheCatLady:
@danshilm You need to rebase. 😛  The movie attributes are now an array that is constructed outside of the `return`.  See 7c0ddad653393327226a877692f046d8693ddc66

sct:
Also, let's display the content rating first in that row :)

src/components/TvDetails/index.tsx
```diff
@@ -393,9 +393,19 @@ const TvDetails: React.FC<TvDetailsProps> = ({ tv }) => {
               </span>
             )}
           </h1>
-          <span className="mt-1 text-xs lg:text-base lg:mt-0">
+          <div className="mt-1 text-xs lg:text-base lg:mt-0">
             {data.genres.map((g) => g.name).join(', ')}
-          </span>
+            {data.contentRatings.results.length > 0 && (
```
TheCatLady, line 398: Create a `seriesAttributes` array similar to `movieAttributes` in `MovieDetails`

src/components/MovieDetails/index.tsx
```diff
@@ -356,18 +356,30 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
             {data.title}{' '}
             <span className="text-2xl">({data.releaseDate.slice(0, 4)})</span>
           </h1>
-          <span className="mt-1 text-xs lg:text-base lg:mt-0">
+          <div className="mt-1 text-xs lg:text-base lg:mt-0">
```
TheCatLady, line 359: Why are we changing this to a `div`?

src/components/TvDetails/index.tsx
danshilm: Got it ;)

src/components/MovieDetails/index.tsx
danshilm: I changed it to a `div` because I was building several `span`s under it and thus felt like a `div` was a more appropriate element. I'm changing it back to a `span` anyway because I'll be using the attributes array you made.

TheCatLady: I think that since `div` is a block element and `span` is inline, it makes more sense semantically to use `span` here since this is basically just a label.  No problem with nested `span` elements regardless 😉  Make sure to revert the change in `TvDetails` as well so it's consistent!

danshilm: On it!

src/components/MovieDetails/index.tsx
```diff
@@ -133,6 +133,16 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
 
   const movieAttributes = [];
 
+  if (data.releases.results.length > 0) {
```
TheCatLady, line 136: Just FYI... since `0` is a falsy value, the `> 0` is unnecessary here.

src/components/MovieDetails/index.tsx
```diff
@@ -133,6 +133,16 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
 
   const movieAttributes = [];
```
sct, line 134:
```suggestion
  const movieAttributes: React.ReactNode[] = [];
```

If you mark this as an array of React nodes here, you won't need to specify it in the map below.

danshilm: Thanks!

src/components/MovieDetails/index.tsx
```diff
@@ -131,9 +131,14 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
     revalidate();
   };
 
-  const movieAttributes = [];
+  const movieAttributes: React.ReactNode[] = [];
 
-  if (data.releases.results.length > 0) {
+  if (
+    data.releases.results.length &&
+    (data.releases.results.find((r) => r.iso_3166_1 === 'US')?.release_dates[0]
```
TheCatLady, line 138: Are these additional `if` conditions necessary?  If so, do we need them in `TvDetails` as well?

TheCatLady: Also, is it the desired behavior to default to the US release dates?  I'm located in the US and personally don't have an issue with it, but I know we have users all around the globe.

danshilm: Unfortunately yes, I've come across a few instances where the `certification` is just an empty string, but only for movies. It might also be the case for tv shows, so I can add this for tv shows as well.
On TMDB's side, there is now validation that prevents this (empty string) from being added to new shows/movies, so the empty strings are just remnants from before that validation was implemented.

sct (approved):

PR #891, opened 2021-02-08
fix(requests): correctly clean up when tvdbid is missing

### LINKED ISSUE #873: Manually matched series by a user with auto-approve permissions does not send the tvdbId to Sonarr
**Describe the bug**
If a (non-admin) user has auto-approve enabled, and has to manually match the series, the tvdbId from the new series is not forwarded to Sonarr, causing Sonarr to fail to add the series. In turn, the `media` object in the request object is `null`, which breaks the Overseerr UI/DB.

v1.19.1

**To Reproduce**

1. Create (or update) a user with the auto-approve permission
2. Using the above user, request a series that requires matching (Details about the content in my scenario will be below)
3. A toast will be displayed in the top right saying an error has occurred, browser logs show HTTP 500 (internal server error)

The discover and requests (specifically approved/all filtered) page for that user and the admin will now be broken, will continue to work for other users and for the pending list of movies/series.

Content in my scenario:
- **Name**: [Mobile Suit Gundam Unicorn](https://www.themoviedb.org/tv/45500-mobile-suit-gundam-unicorn)
- **Matched To**: [Mobile Suit Gundam Unicorn RE:0096](https://www.thetvdb.com/series/mobile-suit-gundam-uc-re-0096)

**Expected behavior**
The request to be added to Overseerr and Sonarr without breaking the instance.

**Additional context**

Overseerr log when that content is added: ```2021-02-07T14:05:11.618Z [error][Media Request]: Request failed to send to Sonarr {"errorMessage":"Request failed to send to sonarr: Series was missing tvdb id"}```

Was debugged and pinpointed in Discord, start of the chain of messages: https://discord.com/channels/783137440809746482/789002037529804833/807957890987851837

ankarhem:
#### Description
cleans up media and mediarequest when request to sonarr has failed because of tvdbid

#### Todos

- Fixes #873

### DIFF
--- a/server/entity/MediaRequest.ts
+++ b/server/entity/MediaRequest.ts
@@ -538,7 +538,9 @@ export class MediaRequest {
         const tvdbId = series.external_ids.tvdb_id ?? media.tvdbId;
 
         if (!tvdbId) {
-          this.handleRemoveParentUpdate();
+          const requestRepository = getRepository(MediaRequest);
+          await mediaRepository.remove(media);
+          await requestRepository.remove(this);
           throw new Error('Series was missing tvdb id');
         }
 


server/entity/MediaRequest.ts
```diff
@@ -538,7 +538,9 @@ export class MediaRequest {
         const tvdbId = series.external_ids.tvdb_id ?? media.tvdbId;
 
         if (!tvdbId) {
-          this.handleRemoveParentUpdate();
+          const requestRepository = getRepository(MediaRequest);
+          await mediaRepository.remove(media);
```
sct, line 542: By deleting the parent media, it should cascade and delete all related requests.

ankarhem: This isn't the case. If we don't remove it from the reqest repository as well, it results in the media having a pending status whilst also not listing itself in requests. So some sort of failed state.

sct: How can the media have a pending state if it's deleted? There are foreign keys that tie the requests to the media so you can't delete the media without deleting its relations. That's what the on delete cascade should be doing?

ankarhem: I don't know. But you can try it yourself.

sct (approved):

PR #892, opened 2021-02-08
build(docker): Leverage GitHub cache using actions/cache

TheCatLady:
#### Description

Since we recently made changes to the `Dockerfile`, I decided to look through some of the GitHub Actions logs today and discovered that we actually aren't caching anything at all. 😅

This PR leverages [GitHub cache](https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows) using [`actions/cache`](https://github.com/actions/cache), as described here: https://github.com/docker/build-push-action#leverage-github-cache


sct (approved):
Sure why not!

PR #899, opened 2021-02-09
feat(ui): Add 'Available' filter to request list and update request status to AVAILABLE where appropriate

TheCatLady:
#### Description

Update `MediaRequestStatus` to `AVAILABLE` when appropriate, and add the "Available" filter option to the request list.

This is an alternative to #905 for implementing #656.  This PR tries to keep request statuses in sync with media statuses.

**Note:** 4K content doesn't get picked up until it is detected during a Radarr/Sonarr/Plex sync.  When we add availability notifications for 4K content, we can fix this.

#### Screenshot (if UI related)

image

#### Todos

- [x] Successful build
- [x] Translation keys
- [x] Update request status to `AVAILABLE` or `PARTIALLY_AVAILABLE` as new media becomes available
- [x] Update existing available request statuses during Radarr/Sonarr/Plex sync

#### Issues Fixed or Closed by this PR

- Fixes #656

PR #905, opened 2021-02-10
feat(ui): Add 'Available' filter to request list and remove unused MediaRequestStatus.AVAILABLE enum value

### LINKED ISSUE #656: Sort by Requested only
No

I'd like to be able to filter the requested list by 'Requested' only and not shows that are 'Available' as I manually find some of my TV series/Movies rather than using Radarr/Sonarr,

I have to use the 'Approved' filter and ignore the large number of 'Available' shows

TheCatLady:
#### Description

Remove unused `MediaRequestStatus` enum value `AVAILABLE` , and add "Available" and "Processing" filter options to the request list page.

This is an alternative to #899 for implementing #656, using a more advanced query for the `/request` endpoint instead of trying to keep request statuses in sync with media statuses.

Currently still trying to find an elegant way to filter on and display availability for partial series requests (e.g., when a user only requests one season of a series and that becomes available, we should treat that as an available request even if other seasons are missing).

image

#### Todos

- [x] Successful build
- [x] Translation keys
- [x] Add filters to request list
- [x] Update `/request/count` endpoint

- Fixes #656

### DIFF
--- a/server/routes/request.ts
+++ b/server/routes/request.ts
@@ -22,8 +22,9 @@ requestRoutes.get('/', async (req, res, next) => {
     let statusFilter: MediaRequestStatus[];
 
     switch (req.query.filter) {
-      case 'available':
       case 'approved':
+      case 'processing':
+      case 'available':
         statusFilter = [MediaRequestStatus.APPROVED];
         break;
       case 'pending':
@@ -49,8 +50,7 @@ requestRoutes.get('/', async (req, res, next) => {
       case 'available':
         mediaStatusFilter = [MediaStatus.AVAILABLE];
         break;
-      case 'approved':
-      case 'pending':
+      case 'processing':
       case 'unavailable':
         mediaStatusFilter = [
           MediaStatus.UNKNOWN,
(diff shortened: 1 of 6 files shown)


server/routes/request.ts
```diff
@@ -282,9 +319,23 @@ requestRoutes.get('/count', async (_req, res, next) => {
...
+      .leftJoinAndSelect('request.requestedBy', 'requestedBy')
+      .where('request.status = :requestStatus', {
+        requestStatus: MediaRequestStatus.APPROVED,
+      })
+      .andWhere(
+        '(request.is4k = false AND media.status != :availableStatus) OR (request.is4k = true AND media.status4k != :availableStatus)',
+        {
+          availableStatus: MediaStatus.AVAILABLE,
+        }
+      )
```
sct, line 337: Why does the approved count filter out available titles? Just because they become available doesn't make them not approved?

TheCatLady: Yes, but the point here was to separate available requests from approved, still-processing requests, if you take a look at #656.  Do we want to rename the filter option to something else?

sct: This is the count endpoint and it says "approved count" so it should be the number of approved requests, regardless of their availability. It doesn't say "available count"?

sct: We shouldn't change it either. If you want to add a new count for available items, sure? But I don't see why we would change the behavior for "approved".

TheCatLady: I think that an "approved" count that includes available requests isn't particularly useful, however.  Wouldn't we want to know how many requests are approved and still pending?  That's actually what the Ombi endpoint returns and it was the original intention of this endpoint.  It's far more useful this way, since the "approved" count will otherwise just be a counter that keeps growing/increasing.

sct: Then we should add another field like "processing" because approved to me means returning the count of approved items.

TheCatLady: Sure!  That sounds like a good idea.

TheCatLady: I'll make the change to both endpoints and add the additional filter option to the request list.

sct (approved):

PR #928, opened 2021-02-14
fix(frontend): changed plex, request, and cog buttons to align proper…

OwsleyJr:
…ly on smaller mobile UIs

#### Description
Play on plex/watch trailer, request, and settings cog buttons would break on a smaller mobile device such as an iPhone X. Also aligns media status with the poster.
Before:
before
After:
new version

#### Todos

### DIFF
--- a/src/components/MovieDetails/index.tsx
+++ b/src/components/MovieDetails/index.tsx
@@ -394,11 +394,11 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
                 ))}
           </span>
         </div>
-        <div className="relative z-10 flex flex-wrap justify-center flex-shrink-0 mt-4 sm:justify-end sm:flex-nowrap lg:mt-0">
+        <div className="relative z-10 flex flex-wrap justify-center flex-shrink-0 mt-2 space-y-2 sm:space-y-0 sm:justify-end sm:flex-nowrap sm:mt-4">
           {(trailerUrl ||
             data.mediaInfo?.plexUrl ||
             data.mediaInfo?.plexUrl4k) && (
-            <div className="mb-3 sm:mb-0">
+            <div className="mt-2 sm:mt-0">
               <ButtonWithDropdown
                 buttonType="ghost"
                 text={
--- a/src/components/MovieDetails/index.tsx
+++ b/src/components/MovieDetails/index.tsx
@@ -352,7 +352,7 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
         <div className="flex flex-col flex-1 mt-4 text-center text-white lg:mr-4 lg:mt-0 lg:text-left">
           <div className="mb-2">
             {data.mediaInfo && data.mediaInfo.status !== MediaStatus.UNKNOWN && (
-              <span className="md:mr-2">
+              <span className="lg:mr-2">
                 <StatusBadge
                   status={data.mediaInfo?.status}
                   inProgress={(data.mediaInfo.downloadStatus ?? []).length > 0}
@@ -398,7 +398,7 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
           {(trailerUrl ||
             data.mediaInfo?.plexUrl ||
             data.mediaInfo?.plexUrl4k) && (
-            <div className="mb-3 md:mb-0">
+            <div className="mb-3 sm:mb-0">
               <ButtonWithDropdown
                 buttonType="ghost"
                 text={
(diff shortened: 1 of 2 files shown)


src/components/MovieDetails/index.tsx
```diff
@@ -352,7 +352,7 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
         <div className="flex flex-col flex-1 mt-4 text-center text-white lg:mr-4 lg:mt-0 lg:text-left">
           <div className="mb-2">
             {data.mediaInfo && data.mediaInfo.status !== MediaStatus.UNKNOWN && (
-              <span className="mr-2">
+              <span className="md:mr-2">
```
TheCatLady, line 355: This should be `lg:mr-2`.

src/components/MovieDetails/index.tsx
```diff
@@ -398,89 +398,91 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
...
-                </>
-              }
-              onClick={() => {
-                if (data.mediaInfo?.plexUrl) {
-                  window.open(data.mediaInfo?.plexUrl, '_blank');
-                } else if (data.mediaInfo?.plexUrl4k) {
-                  window.open(data.mediaInfo?.plexUrl4k, '_blank');
-                } else if (trailerUrl) {
-                  window.open(trailerUrl, '_blank');
+            <div className="mb-3 md:mb-0">
```
TheCatLady, line 401: Instead of wrapping `ButtonWithDropdown` in a `div`, you can simply give it a `className` attribute.  Also, it should be `mb-3 sm:mb-0` instead of `mb-3 md:mb-0`.

src/components/TvDetails/index.tsx
```diff
@@ -378,7 +378,7 @@ const TvDetails: React.FC<TvDetailsProps> = ({ tv }) => {
         <div className="flex flex-col flex-1 mt-4 text-center text-white lg:mr-4 lg:mt-0 lg:text-left">
           <div className="mb-2">
             {data.mediaInfo && data.mediaInfo.status !== MediaStatus.UNKNOWN && (
-              <span className="mr-2">
+              <span className="md:mr-2">
```
TheCatLady, line 381: This should be `lg:mr-2`.

src/components/TvDetails/index.tsx
```diff
@@ -422,93 +422,95 @@ const TvDetails: React.FC<TvDetailsProps> = ({ tv }) => {
...
-                </>
-              }
-              onClick={() => {
-                if (data.mediaInfo?.plexUrl) {
-                  window.open(data.mediaInfo?.plexUrl, '_blank');
-                } else if (data.mediaInfo?.plexUrl4k) {
-                  window.open(data.mediaInfo?.plexUrl4k, '_blank');
-                } else if (trailerUrl) {
-                  window.open(trailerUrl, '_blank');
+            <div className="mb-3 md:mb-0">
```
TheCatLady, line 429: Instead of wrapping `ButtonWithDropdown` in a `div`, you can simply give it a `className` attribute.  Also, it should be `mb-3 sm:mb-0` instead of `mb-3 md:mb-0`.

TheCatLady:
What is the reasoning behind the changes in 59611a2?  There are other elements that would also need to be updated as a result of those new changes.

sct:
> What is the reasoning behind the changes in [59611a2](https://github.com/sct/overseerr/commit/59611a21c1f41ce811f47e4c266dcd0e5c0f1076)? There are other elements that would also need to be updated as a result of those new changes.

I asked him to change it to this because I didn't like the buttons going to the edge of the screen. They should respect the same margins the rest of the site has. What other components would need to change?

TheCatLady:
> > What is the reasoning behind the changes in [59611a2](https://github.com/sct/overseerr/commit/59611a21c1f41ce811f47e4c266dcd0e5c0f1076)? There are other elements that would also need to be updated as a result of those new changes.
> 
> I asked him to change it to this because I didn't like the buttons going to the edge of the screen. They should respect the same margins the rest of the site has. What other components would need to change?

Ah I see.  We also need to apply the same margins from the "Play on Plex" button to the `div` surrounding the request button and the cog button, right?  Otherwise we'll run into the same issue depending if the elements fit on one line or not?

OwsleyJr:
> > > What is the reasoning behind the changes in [59611a2](https://github.com/sct/overseerr/commit/59611a21c1f41ce811f47e4c266dcd0e5c0f1076)? There are other elements that would also need to be updated as a result of those new changes.
> > 
> > 
> > I asked him to change it to this because I didn't like the buttons going to the edge of the screen. They should respect the same margins the rest of the site has. What other components would need to change?
> 
> Ah I see. We also need to apply the same margins from the "Play on Plex" button to the `div` surrounding the request button and the cog button, right? Otherwise we'll run into the same issue depending if the elements fit on one line or not?

I changed the margin on the div wrapping the three buttons so they will stay the same on one line.

TheCatLady:
> I changed the margin on the div wrapping the three buttons so they will stay the same on one line.

I just pulled your changes to get a better look.  I see what you did now.  LGTM!

sct (approved):

PR #936, opened 2021-02-15
feat(ui): Add user requests page

TheCatLady:
#### Description

Adds a request list view at `/users/[userId]/requests` that only displays one user's requests.  The "Recent Requests" slider heading links to this page.

image
image

#### Todos

- [x] Successful build
- [x] Translation keys

### DIFF
--- a/server/routes/request.ts
+++ b/server/routes/request.ts
@@ -17,6 +17,9 @@ requestRoutes.get('/', async (req, res, next) => {
   try {
     const pageSize = req.query.take ? Number(req.query.take) : 10;
     const skip = req.query.skip ? Number(req.query.skip) : 0;
+    const requestedBy = req.query.requestedBy
+      ? Number(req.query.requestedBy)
+      : 0;
 
     let statusFilter: MediaRequestStatus[];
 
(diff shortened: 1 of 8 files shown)


server/routes/request.ts
```diff
@@ -17,6 +17,9 @@ requestRoutes.get('/', async (req, res, next) => {
   try {
     const pageSize = req.query.take ? Number(req.query.take) : 10;
     const skip = req.query.skip ? Number(req.query.skip) : 0;
+    const requestedBy = req.query.requestedBy
+      ? Number(req.query.requestedBy)
+      : 0;
```
sct, line 22: Any reason we don't just set it to null here if there isn't a query value?

TheCatLady: Will change that! 👍🏻

src/pages/users/[userId]/requests.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { NextPage } from 'next';
+import RequestList from '../../../components/RequestList';
+import useRouteGuard from '../../../hooks/useRouteGuard';
+import { Permission } from '../../../hooks/useUser';
+
+const UserRequestsPage: NextPage = () => {
+  useRouteGuard(Permission.MANAGE_REQUESTS);
```
Comment (sct) [inline src/pages/users/[userId]/requests.tsx:8]: This guard will prevent people who can view requests from access this also?

Comment (TheCatLady) [reply src/pages/users/[userId]/requests.tsx:8]: Ah right, will add the View Requests perm as well.  Good catch, thanks 😄

sct (approved):

PR #941, opened 2021-02-15
fix(ui): Automatically disable and uncheck user permissions with unmet requirements

TheCatLady:
#### Description

Some permissions are dependent on other, non-parent permissions.  For example, "Auto-Approve 4K" is dependent on both "Request" (general request permission) and "Request 4K" (4K content request permission).  This PR adds a check for requirements for a particular permission, and disables and unchecks a permission if that condition is not satisfied.

#### Todos

- [x] Successful build

### DIFF
--- a/src/components/PermissionOption/index.tsx
+++ b/src/components/PermissionOption/index.tsx
@@ -20,16 +20,9 @@ const hasRequirement = (
   requirement: PermissionRequirement,
   currentPermission: number
 ): boolean => {
-  switch (requirement.type) {
-    case 'or':
-      return requirement.permissions.some(
-        (permission) => !!(permission & currentPermission)
-      );
-    default:
-      return requirement.permissions.every(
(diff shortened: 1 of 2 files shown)


src/components/PermissionOption/index.tsx
```diff
@@ -8,8 +8,30 @@ export interface PermissionItem {
...
+    case 'or':
+      return requirement.permissions.some(
+        (permission) => !!(permission & currentPermission)
+      );
+    default:
+      return requirement.permissions.every(
+        (permission) => !!(permission & currentPermission)
+      );
+  }
+};
```
sct, line 33: Why are we not just using `hasPermission` to check this?

TheCatLady: Fixed!

sct (approved):

PR #942, opened 2021-02-15
feat(regions): add region setting for filtering Discover (#732)

### LINKED ISSUE #564: Upcoming + Trending for TV shows

We got for movies Upcoming, Trending, Popular
And for TV shows only Popular .
I would like to see this also for TV Shows

Ombi v4 

**Additional context**

image

### LINKED ISSUE #732: Ability to filter "Discover" results by language or set a default language server side
Currently "Discover" lists items in all languages by default.

Would like the ability to filter "Discover" items by the language they are in.

Instead of filtering on the "Discover" page, could also just have a server side setting with a drop down where you could pick the language of the moves/shows are populated. Only issue with doing that server side would be if a user wants a movie in a different language. So would still need normal searches to return movies in all languages.

**Additional context**
This would allow "Discover" to deliver all results in the users preferred language so the results are more relevant. If done server side, it would keep users from accidentally adding movies from "Discover" without realizing it's in the wrong language.

dancarter:
#### Description

Adds a Default Region setting to the settings page that filters the Discover page

Screen Shot 2021-02-16 at 0 48 28

#### Todos

- [x] Add global region/language selector to settings
- [x] Add user-specific region/language selector to user settings
- [x] Add upcoming series discover slider

- Fixes #732
- Closes #564


server/api/themoviedb/index.ts
```diff
@@ -351,6 +356,8 @@ class TheMovieDb extends ExternalAPI {
           page,
           include_adult: includeAdult,
           language,
+          with_release_type: 3,
```
dancarter, line 359: Release type 3 is 'Theatrical.' Would it make sense to add any other types as well? Specifically 1 Premiere and 2
Theatrical (limited)

server/api/themoviedb/index.ts
```diff
@@ -364,13 +371,16 @@ class TheMovieDb extends ExternalAPI {
...
+    region = '',
   }: DiscoverTvOptions = {}): Promise<TmdbSearchTvResponse> => {
     try {
       const data = await this.get<TmdbSearchTvResponse>('/discover/tv', {
         params: {
           sort_by: sortBy,
           page,
           language,
+          with_release_type: 6,
+          region: region,
```
dancarter, line 383: These are the correct parameters to filter `/discover/movie` by region, but doesn't seem to work for `/discover/tv`. I haven't seen anything in the API docs for it either. I am wondering if anyone has any idea?

server/api/themoviedb/index.ts
```diff
@@ -594,6 +607,24 @@ class TheMovieDb extends ExternalAPI {
...
+      }
+
+      return regions;
+    } catch (e) {
+      throw new Error(`[TMDB] Failed to fetch countries: ${e.message}`);
+    }
+  }
 }
 
+let regions: TmdbRegion[] | undefined;
```
dancarter, line 628: The first time the regions are retrieved from TMDB they will be stored here. I wanted to check if this makes sense, or if there is somewhere better?

server/index.ts
```diff
@@ -47,6 +48,9 @@ app
     // Load Settings
     const settings = getSettings().load();
 
+    // Load regions from tmdb
+    await new TheMovieDb().getRegions();
```
dancarter, line 52: `getRegions` is called once here so they will be stored for any later calls. Does this approach make sense for retrieving and storing them when the server starts up?

src/components/Settings/SettingsMain.tsx
```diff
@@ -50,6 +52,9 @@ const SettingsMain: React.FC = () => {
   const { data, error, revalidate } = useSWR<MainSettings>(
     '/api/v1/settings/main'
   );
+  const { data: regions, error: regionsError } = useSWR<Region[]>(
+    '/api/v1/discover/regions'
+  );
```
dancarter, line 57: I felt like it didn't make sense for the main settings endpoint to return the regions so I added a second API call here. Please let me know if there is a way to improve this.

src/components/Settings/SettingsMain.tsx
```diff
@@ -219,6 +226,28 @@ const SettingsMain: React.FC = () => {
...
+                      <Field as="select" id="region" name="region">
+                        <option value="">
+                          {intl.formatMessage(messages.regionDefault)}
+                        </option>
+                        {regions?.map((region: Region) => (
+                          <option
+                            key={region.iso_3166_1}
+                            value={region.iso_3166_1}
+                          >
+                            {region.english_name}
```
TheCatLady, line 244: We will want to display the region names in whatever language the user currently has set.

src/components/Settings/SettingsMain.tsx
```diff
@@ -219,6 +226,28 @@ const SettingsMain: React.FC = () => {
                     )}
                   </div>
                 </div>
+                <div className="form-row">
+                  <label htmlFor="region" className="text-label">
+                    {intl.formatMessage(messages.region)}
+                  </label>
+                  <div className="form-input">
+                    <div className="flex max-w-lg rounded-md shadow-sm">
+                      <Field as="select" id="region" name="region">
```
TheCatLady, line 235: This probably should be a multi-select field, so that users can select more than just one region, or no region at all if they want all results.

src/components/Settings/SettingsMain.tsx
```diff
@@ -219,6 +226,28 @@ const SettingsMain: React.FC = () => {
                     )}
                   </div>
                 </div>
+                <div className="form-row">
```
TheCatLady, line 229: This option/row should also probably be immediately above/below the "Hide Available Media" option.  We should try to group/order the options in some logical way, and IMO the application URL, proxy, and CSRF settings should be together.

src/components/Settings/SettingsMain.tsx
sct: Unfortunately, I don't think that TMDB will return different values for this. We have the same issue with `Original Language` displayed on Movie/TV detail pages.

src/components/Settings/SettingsMain.tsx
sct: No region is possible but does TMDB accept multiple regions?

dancarter: Based on the api docs, it doesn't seem like it. I think we would have to make multiple requests to get results for more than one region.

src/components/Settings/SettingsMain.tsx
dancarter: Yeah, the region endpoint only seems to return the names in English unfortunately.

TheCatLady: Ah, that's super unfortunate.  Guess we can't really do anything about this.

server/api/themoviedb/index.ts
sct: This... should work? We should define it above the class though. I wonder if we can just set a very long cache time for this request instead?

dancarter: Oh yeah, caching the request seems like it might be a more elegant solution.

TheCatLady:
Should this be a user setting instead of a global one?  Or if it is a global setting, do we want to allow users the ability to override the setting?

sct:
> Should this be a user setting instead of a global one?  Or if it is a global setting, do we want to allow users the ability to override the setting?

Initially it will be global and we will allow users to override the value as well.

src/i18n/locale/en.json
```diff
@@ -671,7 +677,11 @@
   "components.UserProfile.UserSettings.UserGeneralSettings.displayName": "Display Name",
   "components.UserProfile.UserSettings.UserGeneralSettings.generalsettings": "General Settings",
   "components.UserProfile.UserSettings.UserGeneralSettings.localuser": "Local User",
+  "components.UserProfile.UserSettings.UserGeneralSettings.originallanguage": "Discover Language",
+  "components.UserProfile.UserSettings.UserGeneralSettings.originallanguageTip": "Filter content by original language (this will only apply to the \"Popular\" and \"Upcoming\" categories)",
   "components.UserProfile.UserSettings.UserGeneralSettings.plexuser": "Plex User",
+  "components.UserProfile.UserSettings.UserGeneralSettings.region": "Discover Region",
+  "components.UserProfile.UserSettings.UserGeneralSettings.regionTip": "Filter content by region (this will only apply to the \"Popular\" and \"Upcoming\" categories)",
```
TheCatLady, line 684:
```suggestion
  "components.UserProfile.UserSettings.UserGeneralSettings.originallanguageTip": "Filter content by original language (only applies to \"Popular\" and \"Upcoming\" categories)",
  "components.UserProfile.UserSettings.UserGeneralSettings.plexuser": "Plex User",
  "components.UserProfile.UserSettings.UserGeneralSettings.region": "Discover Region",
  "components.UserProfile.UserSettings.UserGeneralSettings.regionTip": "Filter content by region (only applies to \"Popular\" and \"Upcoming\" categories)",
```
Realized my original suggestion was unnecessarily wordy 😄

sct:
@all-contributors please add @dancarter for code

PR #945, opened 2021-02-15
feat(requests): add ability to limit users requests with quotas

### LINKED ISSUE #263: Ability to limit requests per user, quotas
The ability to limit the amount of requests a particular user can make. This could be done on a per user basis or another thought would be to create permission groups that would have the limits set. That would make it easier to manage multiple users at once vs changing it per user individually. 

To have limits that are imposed per user/group. User feedback of their limits would also be necessary. Currently Ombi does a rolling 7 day request rate. Not sure if you would want to implement the same or have that be configurable. Other options could be x day rolling and x number per month.

ankarhem:
#### Todos

- Fixes #263

### DIFF
--- a/overseerr-api.yml
+++ b/overseerr-api.yml
@@ -2975,6 +2975,51 @@ paths:
                     type: array
                     items:
                       $ref: '#/components/schemas/MediaRequest'
+  /user/{userId}/quota:
+    get:
+      summary: Get quotas for a specific user
+      description: |
+        Returns quota details for a user in a JSON object. Requires `MANAGE_USERS` permission if viewing other users.
+      tags:
+        - users
+      parameters:
+        - in: path
+          name: userId
+          required: true
+          schema:
+            type: number
+      responses:
+        '200':
+          description: User quota details in JSON
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  movie:
+                    type: object
+                    properties:
+                      quotaPeriod:
+                        type: number
+                        example: 7
+                      quotaQuantity:
+                        type: number
+                        example: 10
+                      quotaRequestCount:
+                        type: number
+                        example: 5
+                  tv:
+                    type: object
+                    properties:
+                      quotaPeriod:
+                        type: number
+                        example: 7
+                      quotaQuantity:
+                        type: number
+                        example: 10
+                      quotaRequestCount:
+                        type: number
+                        example: 5
   /user/{userId}/settings/main:
     get:
       summary: Get general settings for a user
--- a/server/lib/settings.ts
+++ b/server/lib/settings.ts
@@ -61,12 +61,21 @@ export interface SonarrSettings extends DVRSettings {
   enableSeasonFolders: boolean;
 }
 
+interface Quota {
+  quotaQuantity: number | undefined;
+  quotaPeriod: number | undefined;
+}
+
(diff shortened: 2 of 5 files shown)


src/components/Settings/SettingsMain.tsx
```diff
@@ -41,6 +41,8 @@ const messages = defineMessages({
   validationApplicationTitle: 'You must provide an application title',
   validationApplicationUrl: 'You must provide a valid URL',
   validationApplicationUrlTrailingSlash: 'URL must not end in a trailing slash',
+  quotaPeriod: 'Default Quota Period (days)',
+  quotaQuantity: 'Default Quota Quantity (requests)',
```
TheCatLady, line 45:
```suggestion
  quotaQuantity: 'Default Request Limit',
```

I think that while "Quota" makes sense for us as devs, "Request Limit" would be clearer and less confusing for the end user.  We should also avoid parentheses in labels / field names; if parentheses are necessary, it probably needs to be reworded/renamed or have a "tip"/description added.

Also, I think that this should appear as one setting/option/row to the end user even though it's really two values, like:
```
Default Request Limit            [      ] requests per [      ] days
```

We also need `Yup` validation on these fields ("You must enter a number of requests" / "You must enter a number of days").

server/lib/settings.ts
```diff
@@ -59,6 +59,8 @@ export interface MainSettings {
   hideAvailable: boolean;
   localLogin: boolean;
   trustProxy: boolean;
+  quotaQuantity?: number;
+  quotaPeriod?: number;
```
TheCatLady, line 63: Nitpicking here, but I don't think "quantity" is the right descriptor.  It kind of sounds like the number of quotas.  Maybe either `quotaLimit` or `quotaRequests`?  And `quotaDays` so that it's clear to anyone looking at the code that it's a number of days?

overseerr-api.yml
```diff
@@ -2818,6 +2818,39 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/User'
+  /user/{userId}/quota:
+    get:
+      summary: Get quotas for a specific user
+      description: |
+        Retrieves quota details for a user in a JSON object.. Requires the `MANAGE_USERS` permission.
```
TheCatLady, line 2825:
```suggestion
        Retrieves quota details for a user in a JSON object. Requires the `MANAGE_USERS` permission.
```

src/components/Settings/SettingsMain.tsx
TheCatLady: On second thought, the `quotaPeriod` setting should probably be a dropdown with options for day/week/month (we can still store the value as 1/7/30 days).

```
Default Request Limit            [      ] requests per [ day   v ]
                                                       [ week    ]
                                                       [ month   ]
```

Then, we don't need validation on that field, and the `quotaQuantity` setting can be optional (so quotas only apply if a value is set).

Inrego:
I may be a bit late to the party, but another great addition would be the option to have separate limits for TV and Movies.

ankarhem:
> I may be a bit late to the party, but another great addition would be the option to have separate limits for TV and Movies.

I haven't updated this remotely but my rebased local branch does seperate this logic. You set one quota for movies, and one quota for how many seasons of TV you can request.

PR #946, opened 2021-02-16
refactor(ui): Create PlayButton component

TheCatLady:
#### Description

Create a PlayButton component which accepts an array of links, to make it easier to add additional links in the future (e.g., for Jellyfin or Emby).

Also fixed a bug where sometimes a "0" is still shown instead of a missing rating, and replaced remaining occurrences of the `FormattedMessage` element with `intl.formatMessage()` in both `MovieDetails` and `TvDetails`.

#### Todos

### DIFF
--- a/src/components/Common/PlayButton/index.tsx
+++ b/src/components/Common/PlayButton/index.tsx
@@ -1,0 +1,68 @@
+import React from 'react';
+import ButtonWithDropdown from '../ButtonWithDropdown';
+
+interface PlayButtonProps {
+  links: PlayButtonLink[];
+}
+
+export interface PlayButtonLink {
+  text: string;
+  url: string;
+}
+
+const PlayButton: React.FC<PlayButtonProps> = ({ links }) => {
+  if (!links || !links.length) {
+    return null;
+  }
+
+  return (
+    <ButtonWithDropdown
+      buttonType="ghost"
+      text={
+        <>
+          <svg
+            className="w-5 h-5 mr-1"
+            fill="none"
+            stroke="currentColor"
+            viewBox="0 0 24 24"
+            xmlns="http://www.w3.org/2000/svg"
+          >
+            <path
+              strokeLinecap="round"
+              strokeLinejoin="round"
+              strokeWidth={2}
+              d="M14.752 11.168l-3.197-2.132A1 1 0 0010 9.87v4.263a1 1 0 001.555.832l3.197-2.132a1 1 0 000-1.664z"
+            />
+            <path
+              strokeLinecap="round"
+              strokeLinejoin="round"
+              strokeWidth={2}
+              d="M21 12a9 9 0 11-18 0 9 9 0 0118 0z"
+            />
+          </svg>
+          <span>{links[0].text}</span>
+        </>
+      }
+      onClick={() => {
+        window.open(links[0].url, '_blank');
+      }}
+    >
+      {links.length > 1 &&
+        links.slice(1).map((link, i) => {
+          return (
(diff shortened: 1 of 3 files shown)


ankarhem:
Maybe you could add `as="a"` support to this component similar to what I did for the `Button` component? It's good so that we follow correct html syntax for what we actual do.

TheCatLady:
> Maybe you could add `as="a"` support to this component similar to what I did for the `Button` component? It's good so that we follow correct html syntax for what we actual do. Or do you think it's out of scope? We should try to prioritize it anyway imo.

Yeah we can totally add it here!  I'll take a look at your PR.  I agree that we should follow correct HTML syntax :)

src/components/Common/ButtonWithDropdown/index.tsx
```diff
@@ -39,21 +38,37 @@ const DropdownItem: React.FC<DropdownItemProps> = ({
...
-}
+  // Had to do declare this manually as typescript would assume e was of type any otherwise
+  onClick?: (
+    e: React.MouseEvent<P extends 'a' ? HTMLAnchorElement : HTMLButtonElement>
+  ) => void;
+};
+
+type ButtonWithDropdownProps<P extends React.ElementType> = {
+  as?: P;
+} & MergeElementProps<P, ButtonWithDropdownBaseProps<P>>;
```
ankarhem, line 61: We could just use any types that are identical from `Button` and import them.

src/components/Common/ButtonWithDropdown/index.tsx
```diff
@@ -39,21 +38,37 @@ const DropdownItem: React.FC<DropdownItemProps> = ({
...
+function ButtonWithDropdown<P extends ElementTypes = 'button'>({
+  as,
   text,
   children,
   dropdownIcon,
   className,
   buttonType = 'primary',
   ...props
-}) => {
+}: ButtonWithDropdownProps<P>): JSX.Element {
```
ankarhem, line 71: This would mean that a dropdown could only have one type, as in only buttons, or only anchors. I'm not sure we want that? I haven't thought it through.

src/components/Common/PlayButton/index.tsx
```diff
@@ -43,18 +46,14 @@ const PlayButton: React.FC<PlayButtonProps> = ({ links }) => {
...
         links.slice(1).map((link, i) => {
           return (
             <ButtonWithDropdown.Item
               key={`play-button-dropdown-item-${i}`}
-              onClick={() => {
-                window.open(link.url, '_blank');
-              }}
+              href={link.url}
+              target="_blank"
               buttonType="ghost"
```
ankarhem: I feel like this is the component that should support a-tags?

src/components/Common/ButtonWithDropdown/index.tsx
TheCatLady: The dropdown is actually only anchors already.  Only the button itself can be one or the other...

src/components/Common/PlayButton/index.tsx
TheCatLady: It actually already is an `a` tag.

src/components/Common/ButtonWithDropdown/index.tsx
TheCatLady: But they aren't exactly the same 🤔

TheCatLady:
Reverted the `as="a"` support, as it probably actually is out of the scope of this PR.  Will need to update the request dropdown as well for that.

sct (approved):
💯

PR #955, opened 2021-02-17
fix(plex): Check Plex server access on user import

### LINKED ISSUE #475: Change import all users from Plex to only import users associated with the current Plex server
**Problem**
I believe the import all users button, should only import users that are currently 'assigned/shared' to the Plex server that Overseerr is hooked up too. This will help with multi-server cases where you then have lots of users who may never log in, due to being on a different Plex server.

**Solution**
The import function could only import the users that have a share to a Plex server that matches the one that Overseerr is connected too.

TheCatLady:
#### Description

Check for Plex server access on user import, so that only users with access are imported.

#### Todos

- [x] Successful build

- Fixes #475

### DIFF
--- a/server/api/plextv.ts
+++ b/server/api/plextv.ts
@@ -212,7 +212,7 @@ class PlexTvAPI {
     return parsedXml;
   }
 
-  public async checkUserAccess(authUser: PlexUser): Promise<boolean> {
+  public async checkUserAccess(userId: number): Promise<boolean> {
     const settings = getSettings();
 
     try {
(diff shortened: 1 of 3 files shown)


server/routes/user.ts
```diff
@@ -265,7 +265,11 @@ router.post('/import-from-plex', async (req, res, next) => {
         await userRepository.save(user);
       } else {
         // Check to make sure it's a real account
-        if (account.email && account.username) {
+        if (
+          account.email &&
+          account.username &&
+          (await mainPlexTv.checkUserAccess(Number(account.id)))
```
sct, line 271: Will this not throw and break the loop? Preventing users from being added?

TheCatLady: Wouldn't it only throw if Plex API access is unavailable, in which case we wouldn't want to import users?

TheCatLady: I mean, we can add a `try`/`catch` here but I feel like exiting is likely the desired behavior?  If the check throws once, isn't it likely to throw for any following user checks as well?

sct: checkUserAccess throws when it can't find the user in the list. Not just when it can't connect to the api.

TheCatLady: Ahh I see, I'll fix that!

sct (approved):

PR #956, opened 2021-02-17
fix(frontend): request and cog button would be misaligned without pla...

OwsleyJr:
…y on plex/watch trailer button

#### Description
Fixed a bug that I missed with my last PR. The request and settings cog margins would be off if the play on plex or watch trailer button was missing. This also fixes the z-index of the watch trailer dropdown in certain scenarios.
Before:
image
image
After:
image
image
#### Todos

### DIFF
--- a/src/components/TvDetails/index.tsx
+++ b/src/components/TvDetails/index.tsx
@@ -422,11 +422,11 @@ const TvDetails: React.FC<TvDetailsProps> = ({ tv }) => {
                 ))}
           </span>
         </div>
-        <div className="flex flex-wrap justify-center flex-shrink-0 mt-4 sm:flex-nowrap sm:justify-end lg:mt-0">
+        <div className="relative z-10 flex flex-wrap justify-center flex-shrink-0 mt-4 sm:justify-end sm:flex-nowrap lg:mt-0">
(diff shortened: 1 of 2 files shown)


src/components/TvDetails/index.tsx
```diff
@@ -422,11 +422,11 @@ const TvDetails: React.FC<TvDetailsProps> = ({ tv }) => {
                 ))}
           </span>
         </div>
-        <div className="flex flex-wrap justify-center flex-shrink-0 mt-2 space-y-2 sm:space-y-0 sm:flex-nowrap sm:justify-end sm:mt-4">
+        <div className="flex flex-wrap justify-center flex-shrink-0 mt-4 sm:flex-nowrap sm:justify-end lg:mt-0">
```
TheCatLady, line 425:
```suggestion
        <div className="relative z-10 flex flex-wrap justify-center flex-shrink-0 mt-4 sm:justify-end sm:flex-nowrap lg:mt-0">
```

Shouldn't this be the same as it is in `MovieDetails`?

OwsleyJr: Good catch. I updated it

sct (approved):

PR #957, opened 2021-02-17
feat(ui): Add 'Page Size' setting for request/user list pages

### LINKED ISSUE #355: Change # of Requests shown on each page
It is not related to a problem.

I'd like to see the ability to increase the number of requests visible on each page above 10. Perhaps an option for 10, 25, 50, and 100 (and maybe even an 'all' option) Whether or not this is browser-specific via a cookie or user-specific is up to the dev, but I think browser-specific would be more graceful.

n/a

**Additional context**
n/a

TheCatLady:
#### Description

Adds a setting to the Requests & Users pages to allow adjustment of the number of items shown per page.  (Also adds pagination to the "User List" page.)

**Desktop view:**
image

**Mobile view:**
image

#### Todos

- [x] Successful build
- [x] Translation keys

- Fixes #355


sct:
I think this should be a dropdown on the request/user lists with some preset values like 10, 20, 30, 50. We can pass any take size to the API based on that choice.

server/routes/user/index.ts
```diff
@@ -9,46 +9,60 @@ import logger from '../../logger';
...
+      .getManyAndCount();
+
+    return res.status(200).json({
+      pageInfo: {
+        pages: Math.ceil(userCount / pageSize),
+        pageSize,
+        results: userCount,
+        page: Math.ceil(skip / pageSize) + 1,
+      },
+      results: User.filterMany(users),
```
sct, line 61: You need the second parameter here for `filterMany`

TheCatLady: Ahh, good catch!  Fixing now~

sct (approved):

PR #966, opened 2021-02-18
fix(ui): Add tip & validation for Discord ID input

TheCatLady:
#### Description

"Discord ID" is not a user's Discord "tag," which may be confusing for users unfamiliar with Discord.

This PR adds a "tip" with a link to the Discord support article detailing how to find one's user ID number, as well as validation to ensure an 18-digit numeric value is inputted.

image

#### Todos

- [x] Successful build
- [x] Translation keys


src/components/UserProfile/UserSettings/UserNotificationSettings/index.tsx
```diff
@@ -97,7 +111,46 @@ const UserNotificationSettings: React.FC = () => {
...
+                          'ja',
+                          'ko',
+                          'nl',
+                          'pl',
+                          'pt-br',
+                          'ru',
+                          'sv',
+                          'tr',
+                          'zh-tw',
+                        ];
```
sct, line 133: This is really thoughtful but also not something we should maintain.

TheCatLady: Is there a better way to do this?  Ideally the Discord support website would just redirect if an invalid language code is used, but it just fails.  NBD if we want to just link to the English article, I suppose.

sct: Not that I am aware of. It's on discord to handle it though.

TheCatLady: Okay, will remove this.

src/components/UserProfile/UserSettings/UserNotificationSettings/index.tsx
```diff
@@ -32,6 +36,13 @@ const UserNotificationSettings: React.FC = () => {
     user ? `/api/v1/user/${user?.id}/settings/notifications` : null
   );
 
+  const UserNotificationSettingsSchema = Yup.object().shape({
+    discordId: Yup.string().matches(
+      /^\d{18}$/,
+      intl.formatMessage(messages.validationDiscordId)
+    ),
+  });
```
sct, line 44: I think this will block saving for people who never fill it in. The default value is null for `discordId` which I believe will fail the string check here.

sct: We had a similar issue where people couldn't edit users without setting a displayname.

sct (approved):

PR #968, opened 2021-02-18
feat(ui): Add support for requesting collections in 4K

### LINKED ISSUE #639: 4k request support for "Request Collection"
Love the 4k request feature!!! Is it possible to support request collection as well with 4k request?

I like to have the possibility to do a 4k request for a whole collection :)

**Additional context**

TheCatLady:
#### Description

Adds support for requesting collections in 4K.

~~Also made one small/minor change to remove the icon from the regular "request" button dropdown children on movie/series detail pages, since there are no icons in other dropdowns (e.g., the "play" button dropdown children) and we should be consistent.  (If we want to add the icons to the request dropdowns, then we should add them everywhere else too!)~~

**Edit:** While testing, found that request buttons were being displayed on title cards even if users lacked the "Request" permission, so this is now fixed here.  Also, we were showing 4K status in some places even if a user lacked the relevant 4K request permissions; fixed that in this PR as well.

image
image

#### Todos

- [x] Successful build
- [x] Translation keys

- Fixes #639


src/components/RequestButton/index.tsx
```diff
@@ -550,7 +550,6 @@ const RequestButton: React.FC<RequestButtonProps> = ({
                 onClick={button.action}
                 key={`request-option-${button.id}`}
               >
-                {button.svg}
```
sct, line 553: Why is this removed?

TheCatLady: As noted above:

> Also made one small/minor change to remove the icon from the regular "request" button dropdown children on movie/series detail pages, since there are no icons in other dropdowns (e.g., the "play" button dropdown children) and we should be consistent. (If we want to add the icons to the request dropdowns, then we should add them everywhere else too!)

TheCatLady: I'm totally cool with putting it back, but then I feel we should update the "play" button dropdown children to have icons as well.  Lmk what you think!

sct: I disagree with this change. Consistency doesn't directly always translate to clarity. The request dropdown has a lot of mixed options that do different things. The icons help quickly understand what you are clicking on at a glance.

sct: All of the actions for the `PlayButton` are "watching" something. They would all have the same icon. It would look silly.

TheCatLady: Hmmm okay.  I'll put it back 😄

sct (approved):
LGTM w/ tiny comment for future reference

src/components/CollectionDetails/index.tsx
```diff
@@ -73,6 +117,7 @@ const CollectionDetails: React.FC<CollectionDetailsProps> = ({
           await axios.post<MediaRequest>('/api/v1/request', {
             mediaId: part.id,
             mediaType: 'movie',
+            is4k: is4k,
```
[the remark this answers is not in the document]
TheCatLady: I appreciate the nitpicking, and will change that!

PR #979, opened 2021-02-19
fix(ui/notif): Custom application title in password-related emails and UI messages

TheCatLady:
#### Description

The application title is not being passed when sending password-related emails, which results in "[undefined]" appearing in the subject and a weird "Open" link (which should be "Open {applicationTitle}") in the message body.

Also, the application title should be shown instead of "Overseerr" in the alert shown on the user password settings page, since that alert is visible to non-admin users who might not know/understand what Overseerr is.  Additionally, added a check to only show this alert only if (1) local user sign-in is enabled _OR_ (2) the user has the ability to modify settings to enable local user sign-in, because it does not make sense to allow users to set a password if they can't use it.

#### Todos

- [x] Successful build (`yarn build`)
- [x] Translation keys (`yarn i18n:extract`)


src/components/UserProfile/UserSettings/index.tsx
```diff
@@ -73,6 +75,14 @@ const UserSettings: React.FC = ({ children }) => {
     regex: RegExp;
     isMobile?: boolean;
   }> = ({ children, route, regex, isMobile = false }) => {
+    if (
+      route === '/settings/password' &&
+      !settings.currentSettings.localLogin &&
+      !hasPermission(Permission.MANAGE_SETTINGS)
+    ) {
+      return null;
+    }
```
sct, line 84: Let's try to keep the PR's scope focused on one thing if we can. Just keep it in mind for next time.

sct (approved):

PR #985, opened 2021-02-20
fix(ui): Notification-related string/UI edits and field validation

TheCatLady:
#### Description

Edit some notification-related language strings for clarity and consistency.  It was confusing to users that notifications were not sent for automatically approved requests, so hopefully these changes will help alleviate some confusion around that.

Also added validation to the Pushover application token and user key fields (which were also renamed to match the official language used by Pushover).

**Edit:** Added validation to the Telegram chat ID field as well.

#### To-Dos

#### Issues Fixed or Closed

N/A


src/components/Settings/SettingsNotifications.tsx
```diff
@@ -38,107 +40,106 @@ interface SettingsRoute {
...
+      ),
+      route: '/settings/notifications/telegram',
+      regex: /^\/settings\/notifications\/telegram/,
+    },
+    {
+      text: intl.formatMessage(messages.webhook),
+      content: (
+        <span className="flex items-center">
+          <Bolt className="h-4 mr-2" />
+          {intl.formatMessage(messages.email)}
```
sct, line 134: Wrong message.

TheCatLady: Ahhh good catch!

src/components/Settings/Notifications/NotificationsPushover/index.tsx
```diff
@@ -38,12 +38,18 @@ const NotificationsPushover: React.FC = () => {
...
-    accessToken: Yup.string().required(
-      intl.formatMessage(messages.validationAccessTokenRequired)
-    ),
-    userToken: Yup.string().required(
-      intl.formatMessage(messages.validationUserTokenRequired)
-    ),
+    accessToken: Yup.string()
+      .required(intl.formatMessage(messages.validationAccessTokenRequired))
+      .matches(
+        /^a[A-Za-z0-9]{29}$/,
```
sct, line 44: Are we absolutely certain this will always be 29 characters? I mean, do we really need strict validation like this here? This is on the user to enter the correct string, I feel.

TheCatLady: It is always 30 characters, the first of which is `a` (user/group tokens always begin with `u` or `g`).  I'm a Pushover user myself and this is clearly documented.

TheCatLady: Documentation:
https://pushover.net/api#registration
https://pushover.net/api#identifiers

src/components/NotificationTypeSelector/index.tsx
```diff
@@ -5,15 +5,17 @@ import NotificationType from './NotificationType';
 const messages = defineMessages({
   mediarequested: 'Media Requested',
   mediarequestedDescription:
-    'Sends a notification when new media is requested. For certain agents, this will only send the notification to admins or users with the "Manage Requests" permission.',
+    'Sends a notification when media is requested and requires approval.',
   mediaapproved: 'Media Approved',
-  mediaapprovedDescription: 'Sends a notification when media is approved.',
+  mediaapprovedDescription:
+    'Sends a notification when media is approved.\
+    By default, automatically approved requests will not trigger notifications.',
```
sct, line 12: What's happening here?

TheCatLady: We get a lot of questions from users about why they aren't receiving notifications, so I wanted to clarify that the default behavior is that auto-approved requests don't trigger notifications.  Perhaps I need to rethink the wording here?

sct: No, I mean what's up with the slash in the middle of it? Haha

TheCatLady: Oh it's just to break it into multiple lines for readability.

sct (approved):

PR #1004, opened 2021-02-23
fix(ui): add missing string for default Discover Language & edit string for default Discover Region

TheCatLady:
#### Description

Change strings to "All Languages" and "All Regions."  (String for default Discover Language was missing previously.)

#### To-Dos

#### Issues Fixed or Closed

N/A


src/components/Settings/SettingsMain.tsx
```diff
@@ -48,6 +48,7 @@ const messages = defineMessages({
   validationApplicationTitle: 'You must provide an application title',
   validationApplicationUrl: 'You must provide a valid URL',
   validationApplicationUrlTrailingSlash: 'URL must not end in a trailing slash',
+  originalLanguageDefault: 'All Languages',
```
sct, line 51: I think you need to do this for the user settings page also.

TheCatLady: Oh right, I forgot about that!  Fixing now

sct (approved):

PR #1006, opened 2021-02-23
fix(frontend): use region settings instead of hardcoded 'US' value for movie/TV ratings

TheCatLady:
#### Description

Movie/TV ratings are associated with releases, and we currently return the rating associated with a US release if one exists.

This changes the behavior to use the global or user region setting, and to fall back to the US region only if those are both set to "All Regions."

N/A

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/src/components/MovieDetails/index.tsx
+++ b/src/components/MovieDetails/index.tsx
@@ -82,20 +82,28 @@ interface MovieDetailsProps {
 
 const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
   const settings = useSettings();
-  const { hasPermission } = useUser();
+  const { user, hasPermission } = useUser();
   const router = useRouter();
   const intl = useIntl();
   const { locale } = useContext(LanguageContext);
   const [showManager, setShowManager] = useState(false);
+
   const { data, error, revalidate } = useSWR<MovieDetailsType>(
     `/api/v1/movie/${router.query.movieId}?language=${locale}`,
     {
       initialData: movie,
     }
   );
+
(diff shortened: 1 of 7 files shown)


src/components/MovieDetails/index.tsx
```diff
@@ -82,21 +82,29 @@ interface MovieDetailsProps {
...
+
   const { data: ratingData } = useSWR<RTRating>(
     `/api/v1/movie/${router.query.movieId}/ratings`
   );
 
+  const { data: userSettings } = useSWR<{
+    username?: string;
+    region?: string;
+    originalLanguage?: string;
+  }>(user ? `/api/v1/user/${user?.id}/settings/main` : null);
```
sct, line 106: This is not necessary. The settings object will be attached to the user that comes back from `useUser`. We just need to update the type.

TheCatLady: Oooh that makes it a lot nicer.  Will fix that shortly

src/hooks/useUser.ts
```diff
@@ -6,6 +6,7 @@ import {
 } from '../../server/lib/permissions';
 import { UserType } from '../../server/constants/user';
 import { mutateCallback } from 'swr/dist/types';
+import { UserSettings } from '../../server/entity/UserSettings';
```
sct, line 9: Let's not cross import this. We don't cross import `User` because it can cause compilation issues. So let's keep rebuilding the user type here manually.

TheCatLady: Got it!

sct (approved):

PR #1008, opened 2021-02-23
feat(ui): in Settings > Services, make Radarr/Sonarr server names and logos clickable links

TheCatLady:
#### Description

I personally would appreciate an easier way to navigate directly to the Radarr/Sonarr instance from the settings page. 😉  Also display the full address, including scheme (http/https) and port, and removed usage of `FormattedMessage`.

image

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/src/components/Settings/SettingsServices.tsx
+++ b/src/components/Settings/SettingsServices.tsx
@@ -117,7 +117,7 @@ const ServerInstance: React.FC<ServerInstanceProps> = ({
           <img
             className="flex-shrink-0 w-10 h-10"
             src={`/images/${isSonarr ? 'sonarr' : 'radarr'}_logo.svg`}
-            alt=""
+            alt={isSonarr ? 'Sonarr' : 'Radarr'}
           />
         </a>
       </div>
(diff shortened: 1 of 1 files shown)


src/components/Settings/SettingsServices.tsx
```diff
@@ -68,40 +74,48 @@ const ServerInstance: React.FC<ServerInstanceProps> = ({
...
-        />
+        {externalUrl && (
+          <a href={externalUrl} className="opacity-50 hover:opacity-100">
+            <img
+              className="flex-shrink-0 w-10 h-10"
+              src={`/images/${isSonarr ? 'sonarr' : 'radarr'}_logo.svg`}
+              alt=""
+            />
+          </a>
+        )}
```
sct, line 111: External URL isn't required for links to sonarr/radarr to work. Some people don't need them in their setup because everything is only internal anyways 🤷

TheCatLady: @sct Ah true.  I guess I can build the URL outside of the `return` and use the internal if external isn't set.

sct (approved):
💯

PR #1027, opened 2021-02-25
feat(telegram): add support for individual chat notifications

### LINKED ISSUE #499: Telegram - Notification to Requester only
The issue #365 led to a PR that will enable notifications using Telegram. According to the developer, the notification will work the same way it works on Ombi, sending all the messages to a single chatid.

My proposal is to add telegram chatid or username (for convenience, Overseerr would get the id in background) to userprofiles. (I don't know if the last is possible). With that information in the database, Overseerr would send the notifications to the chatid on the user profile (if it exists) that the request originated from.

This way only the people involved in the request (requester and users with "Manage Requests" profile) would be notified.

ankarhem:
#### Description

add botUsername to telegram notification settings (so that users can click link to start chat with the admins bot)
add telegram chat Id to userprofile to send notifications to the requestUser as well as the admin
add silent option to userprofile

#### To-Dos

#### Issues Fixed or Closed

- Fixes #499


TheCatLady:
Looks good!  I think that we also need to add some clarification around the difference between the admin notifications and the individual user notifications, but that doesn't need to happen in this PR.

src/components/UserProfile/UserSettings/UserNotificationSettings/index.tsx
```diff
@@ -19,6 +19,15 @@ const messages = defineMessages({
...
   validationDiscordId: 'You must provide a valid Discord user ID',
+  telegramChatId: 'Telegram Chat ID',
+  telegramChatIdTip:
+    'The Chat ID can be aquired by adding <GetIdBotLink>@get_id_bot</GetIdBotLink> to the chat.',
+  telegramChatIdTipLong:
+    'Start a chat by clicking <TelegramBotLink>here</TelegramBotLink>.\
+    Then get the group Chat ID by adding <GetIdBotLink>@get_id_bot</GetIdBotLink> to that chat and send /my_id to the chat',
+  sendSilently: 'Send Silently',
+  sendSilentlyDescription: 'Send telegram notifications silently',
+  validationTelegramChatId: 'You must provide a valid Telegram Chat ID',
```
TheCatLady, line 30:
```suggestion
  telegramChatId: 'Chat ID',
  telegramChatIdTip:
    'Can be found by adding <GetIdBotLink>@get_id_bot</GetIdBotLink> to the chat',
  telegramChatIdTipLong:
    '<TelegramBotLink>Start a chat</TelegramBotLink>, add <GetIdBotLink>@get_id_bot</GetIdBotLink>, and issue the /my_id command',
  sendSilently: 'Send Silently',
  sendSilentlyDescription: 'Send notifications with no sound',
  validationTelegramChatId: 'You must provide a valid chat ID',
```

We probably just need a separate section for these Telegram settings, and `telegramChatIdTipLong` really feels like it belongs in a info box/bubble like in Settings >Notifications.  I think it is too long to be a form label tip.

Otherwise, we can't have a field just named "Send Silently" (and "Telegram Send Silently" / "Send Telegram Silently" just don't sound right), even if there's a tip that explains it is for Telegram.  Perhaps we can use the same formatting/layout from the system-wide/admin notifications?  I imagine that this page will get kinda unwieldy as we add user-specific notification support for other agents if we don't add that type of tab structure to the page.

sct: We can work on refactoring the display of the user notification settings to look more like the notification settings for the app in the future. For this PR's purposes, this is fine.

server/lib/notifications/agents/telegram.ts
```diff
@@ -134,6 +134,19 @@ class TelegramAgent
         disable_notification: this.getSettings().options.sendSilently,
       } as TelegramPayload);
 
+      if (
+        payload.notifyUser.settings?.enableNotifications &&
+        payload.notifyUser.settings?.telegramChatId
```
sct, line 139: What about a check here to see if the `telegramChatId` matches the global chat Id? To prevent duplicate notifications.

sct (approved):

PR #1036, opened 2021-02-26
feat(ci): build arm64 and amd64 container images for ci images

onedr0p:
#### Description

I know this is only for testing images, but having it build `amd64` as well as `arm64` would be nice when people on ARM devices want to test changes.

I'll look at opening a PR for the building of the release images in `arm64` as well.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


samwiseg0:
Nice! You also need to make changes to the release workflow. This only builds develop images.

onedr0p:
Right @samwiseg0 I'll look into it, but it's a little abstracted because we're using sematic-release + npm. Hope I can find a solution that works.

onedr0p:
> Nice!
> 
> Edit: just saw your note about releases as well.
> 
> Has this been tested on the latest changes?

Testing the workflow is not easy, but I can say I do this in multiple other projects and there's no issue using this method.

onedr0p:
@samwiseg0 maybe it's better to move that logic out of the `package.json` and create a build job like you do for `build-snap`?

samwiseg0:
> > Nice!
> > Edit: just saw your note about releases as well.
> > Has this been tested on the latest changes?
> 
> Testing the workflow is not easy, but I can say I do this in multiple other projects and there's no issue using this method.

I understand it’s hard as I have had to do it for snap and based off my experience we need to test this before making changes as the arm builds can be tricky as they have slightly different depends with node. At least they did with snap.

samwiseg0:
> @samwiseg0 maybe it's better to move that logic out of the `package.json` and create a build job like you do for `build-snap`?

That would be up to @sct. I do think it’s a good idea to break out the build workflows.

sct:
> @samwiseg0 maybe it's better to move that logic out of the `package.json` and create a build job like you do for `build-snap`?

That logic exists in package.json because we need the updated `package.json` version that semantic release creates _before_ it commits the changes.

sct:
This won't actually build because certain packages like `sqlite` do not have pre-built binaries on arm. We need to manually install the binaries for this to work.

onedr0p:
> This won't actually build because certain packages like sqlite do not have pre-built binaries on arm. We need to manually install the binaries for this to work.

@sct good call. I built Overseerr locally with `docker buildx` and there was some issues, especially around `node-gyp` ( no stranger to this in my DevOps career xD ). 

I updated the `Dockerfile` with some `buildx` magic and installed `python make g++` only on `arm/v7` and `arm64` build matrixees. I can see that Overseerr is now building successfully around these architectures 👍

For now, I am pretty confident this PR shouldn't break any releases going forward and can be merged if you think everything is good.

My steps:

- `docker buildx create --use`
- `docker buildx build -t overseerr:test --platform linux/amd64,linux/arm64,linux/arm/v7 -f Dockerfile .`

> That logic exists in package.json because we need the updated package.json version that semantic release creates before it commits the changes.

What do you suggest for then if we want to have release images be multi-arch? I am not familiar with wrapping docker build around semantic release but there has to be a way.

Dockerfile
```diff
@@ -1,11 +1,20 @@
...
 ARG COMMIT_TAG
 ENV COMMIT_TAG=${COMMIT_TAG}
 
 COPY . /app
 WORKDIR /app
 
+RUN \
+  case "${TARGETPLATFORM}" in \
+    'linux/arm64') apk add --update --no-cache python make g++ ;; \
+    'linux/arm/v7') apk add --update --no-cache python make g++ ;; \
```
TheCatLady, line 15:
```suggestion
    'linux/arm64') apk add --no-cache python make g++ ;; \
    'linux/arm/v7') apk add --no-cache python make g++ ;; \
```
I was previously advised that the `--update` flag is not desirable if you want consistently reproducible builds.

onedr0p: IIRC `--no-cache` implies `--update`, so the package cache is up-to-date for the packages you want to install.

I'll remove it since it's redundant.

TheCatLady: @onedr0p Ahhh you're right 🤔🤷🏻‍♀️

sct:
As long as we still prepare the runner environment with the same steps we do in the develop CI, we can probably just execute a direct buildx command in the semantic-release config? I think?

onedr0p:
Maybe... but that's going to require some testing in another repo though because I cannot test it here.

sct:
> Maybe... but that's going to require some testing in another repo though because I cannot test it here.

Fair enough. We can start with this PR, and if all works well, I will work on getting it into the release build.

sct (approved):

PR #1042, opened 2021-02-27
fix(ui): for server default options, display "All" region/language option instead of empty string

TheCatLady:
#### Description

When the server default is not set, the option is currently "Default ()" instead of "Default (All <Regions/Languages>)."

Also, I think the application title is unnecessary here?  It can also result in the text not being fully visible if the user-set custom title is too long, so I feel it is best to just omit it.

- **Before:**
  image
- **After:**
  image

#### To-Dos

#### Issues Fixed or Closed

N/A


src/components/RegionSelector/index.tsx
```diff
@@ -77,12 +82,11 @@ const RegionSelector: React.FC<RegionSelectorProps> = ({
                         }) ?? selectedRegion.english_name
                       : isUserSetting && selectedRegion?.iso_3166_1 !== 'all'
                       ? intl.formatMessage(messages.regionServerDefault, {
-                          applicationTitle: currentSettings.applicationTitle,
                           region: currentSettings.region
                             ? intl.formatDisplayName(currentSettings.region, {
                                 type: 'region',
                                 fallback: 'none',
-                              }) ?? currentSettings.region
+                              })
```
sct, line 89: You still need the `currentSettings.region` for the fallback when formatDisplayName cant find the localization of the region.

src/components/RegionSelector/index.tsx
```diff
@@ -125,24 +129,27 @@ const RegionSelector: React.FC<RegionSelectorProps> = ({
...
-                              applicationTitle:
-                                currentSettings.applicationTitle,
                               region: currentSettings.region
                                 ? intl.formatDisplayName(
                                     currentSettings.region,
                                     {
                                       type: 'region',
                                       fallback: 'none',
                                     }
-                                  ) ?? currentSettings.region
```
sct, line 145: You still need the `currentSettings.region` for the fallback when formatDisplayName cant find the localization of the region.

src/components/UserProfile/UserSettings/UserGeneralSettings/index.tsx
```diff
@@ -167,15 +167,17 @@ const UserGeneralSettings: React.FC = () => {
...
                         {intl.formatMessage(messages.languageServerDefault, {
-                          applicationTitle: currentSettings.applicationTitle,
-                          language:
-                            intl.formatDisplayName(
-                              currentSettings.originalLanguage,
-                              {
-                                type: 'language',
-                                fallback: 'none',
-                              }
-                            ) ?? currentSettings.originalLanguage,
```
sct, line 178: Same thing here

src/components/RegionSelector/index.tsx
sct: Actually its probably easier to set `fallback` in this case to `code`. Since we would be falling back to the code anyways in this rare case.

sct: Actually, on iOS this is an even bigger problem because the intl support for display names is really bad. I think we need to actually search through the tmdb regions for a matching code and use the `english_name` provided by them.

sct:
Are we missing some changes here now in the region selector?

src/components/RegionSelector/index.tsx
```diff
@@ -82,12 +87,15 @@ const RegionSelector: React.FC<RegionSelectorProps> = ({
...
                           region: currentSettings.region
                             ? intl.formatDisplayName(currentSettings.region, {
                                 type: 'region',
                                 fallback: 'none',
-                              }) ?? currentSettings.region
+                              }) ??
+                              regions?.find(
+                                (region) =>
+                                  region.iso_3166_1 === currentSettings.region
+                              )?.english_name
```
sct, line 98: We can clean this up by finding the fallback name outside of the return instead of inlining this find.

src/components/UserProfile/UserSettings/UserGeneralSettings/index.tsx
```diff
@@ -167,15 +167,17 @@ const UserGeneralSettings: React.FC = () => {
...
+                          language: currentSettings.originalLanguage
+                            ? intl.formatDisplayName(
+                                currentSettings.originalLanguage,
+                                {
+                                  type: 'language',
+                                  fallback: 'none',
+                                }
+                              ) ?? currentSettings.originalLanguage
```
sct, line 177: I think we will need the same type of fallback here. Otherwise, this would just say `en`.

sct (approved):
Looks great. :D

PR #1043, opened 2021-02-27
feat(lang): localize job names

TheCatLady:
#### Description

Instead of displaying a hardcoded job name, use a localized job name when available.

N/A

#### To-Dos

#### Issues Fixed or Closed

N/A


overseerr-api.yml
```diff
@@ -2087,9 +2087,6 @@ paths:
                     id:
                       type: string
                       example: job-name
-                    name:
-                      type: string
-                      example: A Job Name
```
sct, line 2092: No reason to remove this. It's also a breaking API change. Let's just leave it in.

src/components/Settings/SettingsJobsCache/index.tsx
```diff
@@ -66,7 +77,7 @@ const SettingsJobs: React.FC = () => {
     await axios.post(`/api/v1/settings/jobs/${job.id}/run`);
     addToast(
       intl.formatMessage(messages.jobstarted, {
-        jobname: job.name,
+        jobname: intl.formatMessage(messages[job.id]),
```
sct, line 80: We should design this to check if the message actually exists. If it doesn't exist there should be a fallback string like `Unknown Job.`

overseerr-api.yml
TheCatLady: I didn't think it was being used anywhere, but that's a good point.  I'll add it back!

sct: Other people might be using it though. Never know.

TheCatLady: @sct Yeah you're totally right :)

server/job/schedule.ts
```diff
@@ -7,9 +7,9 @@ import downloadTracker from '../lib/downloadtracker';
 
 interface ScheduledJob {
   id: string;
-  job: schedule.Job;
   name: string;
   type: 'process' | 'command';
+  job: schedule.Job;
```
sct, line 12: Don't really need this change?

src/components/Settings/SettingsJobsCache/index.tsx
```diff
@@ -11,7 +16,7 @@ import Badge from '../../Common/Badge';
 import { CacheItem } from '../../../../server/interfaces/api/settingsInterfaces';
 import { formatBytes } from '../../../utils/numberHelpers';
 
-const messages = defineMessages({
+const messages: { [messageName: string]: MessageDescriptor } = defineMessages({
```
sct, line 19: Why are we typing this ourselves? `defineMessages` should already be returning the correct type?

server/job/schedule.ts
TheCatLady: Oh whoops, that was unintentional!

src/components/Settings/SettingsJobsCache/index.tsx
TheCatLady: It doesn't like it when you do `messages[someString]` :(

sct (approved):

PR #1067, opened 2021-03-02
feat: add genre/studio/network view to Discover results

### LINKED ISSUE #493: Make Studio and Genre clickable and searchable
This feature is not related to a problem, but a request for usability consistency.

It's currently possible to click on actors and other cast members of items. It would be awesome to be able to also click on 2 other tags that are listed in the item detail page: each individual genre, as well as the network/studio. Clicking on each genre would load top items for that genre, and clicking the studio would give all media that network/studio has produced. 

Perhaps, instead of making the genres and studios hyperlinks, there can be a row in addition to Recommendations and Similar titles for other movies/series the studios or networks have produced and maybe a row for other movies/series in those genres (inclusive).

**Additional context**
Understandably, Genre is pretty broad, so it would return many results. So that one may be already covered by Similar titles and Recommendations. 
For the networks and studios, if using hyperlinks, it would follow the same UI as clicking on the names of Director, Producer, Screenplay, etc.

TheCatLady:
#### Description

Adds filtered "Discover" views by genre, studio, and network at:
- `/discover/movies/genre/{genreId}`
- `/discover/tv/genre/{genreId}`
- `/discover/movie/studio/{studioId}`
- `/discover/tv/network/{networkId}`

Links to these views added to movie & TV detail pages.

image

#### To-Dos

#### Issues Fixed or Closed

- Fixes #493


server/api/themoviedb/index.ts
```diff
@@ -368,17 +378,18 @@ class TheMovieDb extends ExternalAPI {
           page,
           include_adult: includeAdult,
           language,
-          with_release_type: '3|2',
```
sct, line 371: Why are we removing this?

TheCatLady: It was removing valid results (e.g., items which only have a digital release).  I'm unsure why this was added in the first place?

sct: It ignores the region without it, from what we saw.

TheCatLady: @sct It didn't seem to be the case when I tested, but let me do some more testing to verify.

TheCatLady: @sct I verified that it does respect the region without `with_release_type` being set.  `with_release_type` is just an additional filter so that returned results must have at least one release of the specified types in the region.  Without `with_release_type`, results are just required to have a release of any type for the region.

Setting this value to `3|2` just made the region setting appear to do more, since it limited the results to items with a theatrical release in the specified region.  I don't think this is desirable, since it filters out items that are exclusive to streaming services, for example.

sct (approved):

PR #1072, opened 2021-03-03
Added Docker compose installation example

NuroDev:
#### Description

The current installation documentation replies on a manual deployment method. Docker-compose allows a more hands off managed system.
This PR simply adds an example to Docker installation methods to get up and running using a `docker-compose.yml` file


docs/getting-started/installation.md
```diff
@@ -25,6 +25,25 @@ docker run -d \
...
+  container_name: overseerr
+  restart: unless-stopped
+  hostname: overseerr
+  ports:
+    - 5055:5055
+  environment:
+    - TZ=Asia/Tokyo
+    - LOG_LEVEL=info
+  volumes:
+    - $PWD/config:/app/config/
```
TheCatLady, line 42:
```suggestion
overseerr:
  image: sctx/overseerr:latest
  container_name: overseerr
  environment:
    - LOG_LEVEL=info
    - TZ=Asia/Tokyo
  ports:
    - 5055:5055
  volumes:
    - /path/to/appdata/config:/app/config
  restart: unless-stopped
```
Let's be consistent with the other exampes.

docs/getting-started/installation.md
```diff
@@ -25,6 +25,25 @@ docker run -d \
 
 {% endtab %}
 
+
+{% tab title="Compose" %}
```
TheCatLady, line 29: Add an additional line break here to match the formatting of the rest of the document.

sct:
Let's include all of what's required to get running in docker-compose, not just a single service

TheCatLady:
> Let's include all of what's required to get running in docker-compose, not just a single service

I think this is a good call.  @NuroDev you just need to add:
```
---

services:
```
and then indent everything.  (Do we want to specify a Compose file version?)

sct:
> > Let's include all of what's required to get running in docker-compose, not just a single service
> 
> I think this is a good call. @NuroDev you just need to add:
> 
> ```
> ---
> 
> services:
> ```
> 
> and then indent everything. (Do we want to specify a Compose file version?)

Yes, we should.

NuroDev:
Was just working on adding this.
I feel it would be better to specify any docker-compose version. That way the docs has a complete example. EG: If you want to copy paste the snippet that is all you need from nothing.

One other thing to note: Would it be better to add a comment prefix on the first line to specify that it is a file / the file name?
EG:
```yaml
# docker-compose.yml
version: "3"

services:
# ...
```

TheCatLady:
Compose file version is not required, though.  And @NuroDev, the YAML should begin with `---` to denote the beginning of the document.  And no, I don't think a comment is appropriate.  You can add the filename above the code block, though.

TheCatLady:
And if we do specify a Compose version, we should figure out the minimum required version that works and use that.  Not everyone has an updated version of Compose.

NuroDev:
@TheCatLady This is true. Having a quick look at the Dockerfile I would **_guess_** that version 2.x would be sufficient, but I do agree that it would be better to work out an exact minimum required version.

I'll also go ahead and add the file name above the code block

sct:
> @TheCatLady This is true. Having a quick look at the Dockerfile I would **_guess_** that version 2.x would be sufficient, but I do agree that it would be better to work out an exact minimum required version.
> 
> I'll also go ahead and add the file name above the code block

I think let's just stick with the current standard version, 3.

TheCatLady:
> I think let's just stick with the current standard version, 3.

Since the version defined in the Compose file is only informative, I think we should use a version 2.x, whatever the minimum required version is.  More info: https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-top-level-element

docs/getting-started/installation.md
```diff
@@ -25,6 +25,29 @@ docker run -d \
 
 {% endtab %}
 
+{% tab title="Compose" %}
+
+docker-compose.yml
```
TheCatLady, line 30:
```suggestion
**docker-compose.yml:**

```

We bold headings in tabs elsewhere, so seems appropriate to do the same for the filename here.

sct:
> https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-top-level-element

The docs you link though go against what you said?

> but prefer the most recent schema at the time it has been designed.

TheCatLady:
> The docs you link though go against what you said?

I guess I interpreted that differently. 😅  I guess it's not much of an issue to specify v3.x.  Do note that if a minor version number isn't specified, it will default to 3.0 (it doesn't prefer the latest version).

sct (approved):
Looks good! Get this rebased and we can merge it.

sct:
@all-contributors please add @NuroDev for docs

PR #1079, opened 2021-03-04
fix(ui): correct language usage re: "sync" vs. "scan"

TheCatLady:
#### Description

Since we do not sync Overseerr status/data back to services, these tasks are more accurately described as "scan" rather than "sync" jobs. (In the future, when syncing is added, it will be a separate job independent of the current "scan" jobs.)

**Note:** This PR updates the descriptions of Plex "sync" endpoints to reflect that they are actually triggering scans, however the endpoints themselves were not renamed to avoid yet another breaking API change.

#### To-Dos

#### Issues Fixed or Closed

N/A


server/job/schedule.ts
```diff
@@ -32,63 +32,65 @@ export const startJobs = (): void => {
...
     cancelFn: () => jobSonarrSync.cancel(),
   });
 
-  // Run download sync
+  // Run download scan
   scheduledJobs.push({
-    id: 'download-sync',
-    name: 'Download Sync',
+    id: 'download-scan',
+    name: 'Download Scan',
```
sct, line 79: This is actually in sync!

TheCatLady: Doesn't "sync" infer it is two-way?  Don't we just scan and display the download status? 🤔

sct: Not necessarily? It just means the state is synchronized. Which it is.

TheCatLady: I suppose that's true.  But I still feel like it'd be clearer to have these current tasks which just fetch the current status of things be called "scans," and future jobs which do two-way synchronization to be called "syncs"?

(Just thinking ahead here.  IMO we should try to be deliberate and clear with language to minimize confusion around what these jobs actually do.)

sct (approved):

PR #1081, opened 2021-03-04
fix(lang): use localized genre names

TheCatLady:
#### Description

TMDb supports the `language` parameter on their genres endpoints, so we should pass it to get localized genre names on the new genre-filtered Discover pages.  (The companies and networks endpoints unfortunately do not support this.)

#### To-Dos

#### Issues Fixed or Closed

N/A


sct:
No longer needed!

PR #1095, opened 2021-03-05
fix(ui): add alt prop to studio/network logos & fix blinking text cursor

TheCatLady:
#### Description

There is no `alt` text for the studio/network logos on the filtered Discover pages, so this PR adds that.

Also noticed that a blinking text cursor would appear either before/after the logos, so fixed that as well.

**Blinking text cursor:**
image

#### To-Dos

#### Issues Fixed or Closed

N/A


TheCatLady:
Also a little torn as to whether the page heading (not the page title) should simply be the name of the studio/network, to match the meaning/behavior when there is no logo.  I feel like it definitely works for studios, but network names are sometimes really short and it can look odd.  🤔

sct (approved):

PR #1105, opened 2021-03-05
fix(ui): improve responsive design on new request list UI

TheCatLady:
#### Description

The new request list UI is amazing, but needed a few tweaks 😸

Notable changes:

- media titles are not truncated as much as before
- season badges scrolling like on the Discover request cards
- consistent styling w/ Discover request cards
- make better use of available horizontal space on medium-size screens

Updated view at varying window/screen sizes (only depicts changes from the first commit in this PR  because I don't want to take new screenshots):
image

#### To-Dos

#### Issues Fixed or Closed

N/A


src/components/RequestCard/index.tsx
```diff
@@ -100,52 +104,64 @@ const RequestCard: React.FC<RequestCardProps> = ({ request, onTitleData }) => {
...
+                ? intl.formatMessage(globalMessages.declined)
+                : intl.formatMessage(globalMessages.failed)}
+            </Badge>
+          ) : (
             <StatusBadge
               status={
-                requestData.is4k
-                  ? requestData.media.status4k
-                  : requestData.media.status
+                requestData.media[requestData.is4k ? 'status4k' : 'status']
```
sct, line 147: I don't understand why this was changed?

src/components/RequestCard/index.tsx
```diff
@@ -215,26 +231,25 @@ const RequestCard: React.FC<RequestCardProps> = ({ request, onTitleData }) => {
...
+      >
+        <a className="flex-shrink-0 w-20 sm:w-28">
           <img
             src={
               title.posterPath
                 ? `//image.tmdb.org/t/p/w600_and_h900_bestv2${title.posterPath}`
                 : '/images/overseerr_poster_not_found.png'
             }
-            alt=""
+            alt={isMovie(title) ? title.title : title.name}
```
sct, line 248: This alt is not necessary. We already show the title. Screen readers would say the movie title twice in this case.

src/components/RequestList/RequestItem/index.tsx
```diff
@@ -130,114 +132,122 @@ const RequestItem: React.FC<RequestItemProps> = ({
...
-                status={
-                  requestData.media[requestData.is4k ? 'status4k' : 'status']
-                }
-                inProgress={
-                  (
-                    requestData.media[
-                      requestData.is4k ? 'downloadStatus4k' : 'downloadStatus'
-                    ] ?? []
-                  ).length > 0
+              alt={isMovie(title) ? title.title : title.name}
```
sct, line 150: Same with the other case, alt can be empty.

src/components/RequestCard/index.tsx
TheCatLady: I realized that the badge was not being displayed for failed requests, so I copied the status badge code/logic that we have in the request list.  I think it's best to be consistent?  And I prefer this since it's more readable/compact.

sct: More compact does not necessarily mean more readable. I prefer the previous style, but it's subjective. You also lose auto-completion writing it this way.

TheCatLady: That's true.  I'll update it in both locations then?

TheCatLady: I was previously under the impression that this way was preferable since we deal with the download status in the same way (but I suppose that is also more complex since it's an array and not as simple to read as this case).

sct: It's useful for inline conditional values, but this is just using it as a simple ternary statement. I don't really think it's super important. If you do change it, we should update the RequestItem also. (I copied the code from the old component, which is why it's unchanged)

I think the main thing about this is we are changing it here just to change it. There really isn't a point other than "I like it the other way?" You can argue consistency but code is organic and changes all the time. We can't maintain a specific and consistent style across the code base and we would be crazy to try to. We've already spent too much time over something very insignificant :D

TheCatLady: @sct So I actually did copy this from `RequestItem`.  I only changed it because I noticed they were different but the badges should really be the same.  Whichever way you want to keep it, I do prefer that it's the same in both places!  That was my original intention.

sct: You will notice things are different all the time, but sometimes we gotta say, "it doesn't matter, it works!" We have tons of contributors, coding styles, preferences. Adding extra changes to a PR to keep things consistent (if that's not the PR's purpose) increases the cognitive load of the PR with no real gain.

I think it's best to leave it for now, since it's the same, but let's keep this in mind in the future.

TheCatLady: You got it, boss!

sct (approved):

PR #1106, opened 2021-03-06
fix(ui): fix request list UI behavior when season list is too long

TheCatLady:
#### Description

Removal of the `overflow-hidden` class in a couple places when the poster link was added broke the request item display when the season list is too long.

Fixed that and also made user avatars behave in a similar way to the posters on link hover.

Before:
image

After:
image

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -63,6 +63,7 @@ module.exports = {
     margin: ['first', 'last', 'responsive'],
     boxShadow: ['group-focus'],
     opacity: ['disabled', 'hover', 'group-hover'],
+    scale: ['group-hover'],
     zIndex: ['hover', 'responsive'],
   },
   plugins: [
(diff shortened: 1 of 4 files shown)


tailwind.config.js
```diff
@@ -63,6 +63,7 @@ module.exports = {
     margin: ['first', 'last', 'responsive'],
     boxShadow: ['group-focus'],
     opacity: ['disabled', 'hover', 'group-hover'],
+    scale: ['group-hover'],
```
sct, line 66: You need to add the defaults here, such as `hover` or any other scaling hovers not using group-hover won't work.

TheCatLady: Hm.  I was literally testing the poster hovers though, and they didn't stop working with this addition?

sct: Better to assume it won't work in some situation and add it to be safe. Like if it's compiled away or removed as a default by tailwind or something. No reason not to specify it in the variants.

sct (approved):

PR #1109, opened 2021-03-07
fix(frontend): status, requested by, and modified alignment fix

OwsleyJr:
#### Description
Status, requested by, and modified were not aligned in the request item on mobile. Plus I removed status in the request card on mobile.
image
image
#### To-Dos

### DIFF
--- a/src/components/RequestCard/index.tsx
+++ b/src/components/RequestCard/index.tsx
@@ -159,7 +159,7 @@ const RequestCard: React.FC<RequestCardProps> = ({ request, onTitleData }) => {
         </div>
         {request.seasons.length > 0 && (
           <div className="card-field">
-            <span className="card-field-name">
+            <span className="hidden sm:block card-field-name">
               {intl.formatMessage(messages.seasons)}
             </span>
             {!isMovie(title) &&
--- a/src/components/RequestCard/index.tsx
+++ b/src/components/RequestCard/index.tsx
@@ -130,7 +130,7 @@ const RequestCard: React.FC<RequestCardProps> = ({ request, onTitleData }) => {
           </Link>
         </div>
         <div className="card-field">
-          <span className="card-field-name">
+          <span className="hidden sm:block card-field-name">
             {intl.formatMessage(messages.status)}
           </span>
           {requestData.media[requestData.is4k ? 'status4k' : 'status'] ===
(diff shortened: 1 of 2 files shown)


TheCatLady:
@OwsleyJr can you also fix the button alignment on the request cards?  Just need to add `pt-0.5 sm:pt-1` to line 185.  Thanks! ❤️

Also, I think we should keep the "Seasons" text on mobile since we do have horizontal scroll on the season number list, and it looks sorta odd/confusing to just have a numbered badge.

src/components/RequestCard/index.tsx
```diff
@@ -159,7 +159,7 @@ const RequestCard: React.FC<RequestCardProps> = ({ request, onTitleData }) => {
         </div>
         {request.seasons.length > 0 && (
           <div className="card-field">
-            <span className="card-field-name">
+            <span className="hidden sm:block card-field-name">
```
TheCatLady, line 162:
```suggestion
            <span className="card-field-name">
```

TheCatLady: And maybe we can move "Seasons" above the status (just switch the order), to be consistent with the ordering on the request list items?  I feel like it makes more sense that way too, since the seasons are a key attribute of the request.

OwsleyJr: I added just the seasons text but not sure if it looks good. What do you think @sct?
image

TheCatLady: @OwsleyJr Just FYI, all of your screenshots in this PR display a personal e-mail address.  You might want to black those out? 😅

I don't get to make these decisions, but I personally think the "Seasons" text looks just fine 😸  It definitely looks better than just a single numbered badge when a request is only for one season!

OwsleyJr: Oh it's just a throwaway email that I made years ago that I use for testing so not a big deal. And I do think having just the badge looked a little odd especially if it's only one season.

TheCatLady: Btw, do we want to modify the "Seasons" strings on the request list & cards to reflect the correct plurality?

```
seasons: '{seasonCount, plural, one {Season} other {Seasons}}',
```

```
{intl.formatMessage(messages.seasons, { seasonCount: request.seasons.length })}
```

sct:
Let's revert to now having seasons, for now, to fix the alignment issues first. I need to think about the mobile design more and how to best fit seasons.

sct (approved):

PR #1111, opened 2021-03-07
feat: add language-filtered Discover pages

TheCatLady:
#### Description

Adds Discover pages for results of specific languages, which are linked to from the movie & TV detail pages.

Also use `intl.formatDisplayName()` to localize language names on movie & TV detail pages, since the `name` property being returned by TMDb is not localized, but in the language itself.  (This was the original goal of this PR, haha.)

#### To-Dos

#### Issues Fixed or Closed

N/A


server/routes/discover.ts
```diff
@@ -63,6 +63,45 @@ discoverRoutes.get('/movies', async (req, res) => {
...
+discoverRoutes.get<{ language: string }>(
+  '/movies/language/:language',
+  async (req, res) => {
+    const tmdb = createTmdbWithRegionLanaguage(req.user);
+
+    const languages = await tmdb.getLanguages();
+
+    const language = languages.find(
+      (lang) => lang.iso_639_1 === req.params.language
+    );
```
sct, line 75: We should probably throw an error here if there is no matching language?

TheCatLady: @sct Do we want to also throw errors for the genre/studio/network Discover routes if we can't find matches for the IDs?

server/routes/discover.ts
```diff
@@ -106,10 +153,18 @@ discoverRoutes.get<{ genreId: string }>(
...
-  async (req, res) => {
+  async (req, res, next) => {
     const tmdb = createTmdbWithRegionLanaguage(req.user);
 
-    const studio = await tmdb.getStudio(Number(req.params.studioId));
+    const studio = await tmdb
+      .getStudio(Number(req.params.studioId))
+      .catch(async () => {
+        return undefined;
+      });
```
sct, line 163: Not sure I like this. Might be better to wrap this in a try/catch?

TheCatLady: So that's what I tried initially, but it wasn't working :|  I'll play around with it some more and see if I can figure out another way, though!

TheCatLady: @sct Okay, not sure what was going on before but I switched it back to the try-catch block and it just worked this time?  I must have done something wrong earlier haha

server/routes/discover.ts
```diff
@@ -202,23 +253,25 @@ discoverRoutes.get('/tv', async (req, res) => {
...
 
+    if (!language) {
+      return next({ status: 404, message: 'Unable to retrieve language' });
+    }
+
     const data = await tmdb.getDiscoverTv({
       page: Number(req.query.page),
       language: req.query.language as string,
-      genre: Number(req.params.genreId),
+      originalLanguage: req.params.language as string,
```
sct, line 274: Not sure why we need this type assertion here?

TheCatLady: Ah you're right, whoops!  Removed from both places. 😺

sct (approved):

PR #1113, opened 2021-03-07
feat(notif): add setting for Discord bot username

TheCatLady:
#### Description

Add a setting to allow for customization of the Discord bot username.

image

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/src/components/Settings/Notifications/NotificationsDiscord.tsx
+++ b/src/components/Settings/Notifications/NotificationsDiscord.tsx
@@ -13,6 +13,7 @@ const messages = defineMessages({
   save: 'Save Changes',
   saving: 'Saving…',
   agentenabled: 'Enable Agent',
+  botUsername: 'Bot Username',
   webhookUrl: 'Webhook URL',
   webhookUrlPlaceholder: 'Server Settings → Integrations → Webhooks',
   discordsettingssaved: 'Discord notification settings saved successfully!',
@@ -45,6 +46,7 @@ const NotificationsDiscord: React.FC = () => {
       initialValues={{
         enabled: data.enabled,
         types: data.types,
+        botUsername: data?.options.botUsername,
         webhookUrl: data.options.webhookUrl,
       }}
       validationSchema={NotificationsDiscordSchema}
@@ -54,6 +56,7 @@ const NotificationsDiscord: React.FC = () => {
             enabled: values.enabled,
             types: values.types,
             options: {
+              botUsername: values.botUsername,
               webhookUrl: values.webhookUrl,
             },
           });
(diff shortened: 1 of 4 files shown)


sct:
No need to refactor that yet.

Also, do we really need an option for the Discord username? It can be set to anything in Discord itself. 🤔

TheCatLady:
> No need to refactor that yet.

👍🏻

> Also, do we really need an option for the Discord username? It can be set to anything in Discord itself. 🤔

Currently, we set the username to the application title, which overrides the setting within Discord.  I think that some users may find this useful.  We can also add another field to set the bot avatar.

If we don't want to add this though, then I would like to remove `username` from the JSON payload we send to Discord.  I actually want my bot to be named something different from the application title, since it's kind of silly to have the bot username and the author in the embed be the same text IMO haha.

sct:
> Currently, we set the username to the application title, which overrides the setting within Discord.  I think that some users may find this useful.  We can also add another field to set the bot avatar.
> 
> 
> 
> If we don't want to add this though, then I would like to remove `username` from the JSON payload we send to Discord.  I actually want my bot to be named something different from the application title, since it's kind of silly to have the bot username and the author in the embed be the same text IMO haha.

Oh I wasn't aware that it overwrote the setting on Discord. I guess that makes sense though. Okay yeah then I get why we need this 😄

TheCatLady:
I'll also a setting to configure the bot avatar URL 😺

src/components/Settings/Notifications/NotificationsDiscord.tsx
```diff
@@ -13,14 +13,16 @@ const messages = defineMessages({
...
+  botAvatarUrl: 'Bot Avatar URL',
   webhookUrl: 'Webhook URL',
   webhookUrlPlaceholder: 'Server Settings → Integrations → Webhooks',
   discordsettingssaved: 'Discord notification settings saved successfully!',
   discordsettingsfailed: 'Discord notification settings failed to save.',
   testsent: 'Test notification sent!',
   test: 'Test',
   notificationtypes: 'Notification Types',
-  validationWebhookUrl: 'You must provide a valid URL',
+  validationUrl: 'You must provide a valid URL',
```
sct, line 25: This change seems unnecessary? Translators will have to revisit the string.

TheCatLady: I wanted to use the same string for both URL fields, instead of adding a new string.  I could keep it as `validationWebhookUrl`, but I was afraid that a translator might see the name and translate it as "You must provide a valid webhook URL," which wouldn't work for the avatar URL field.

TheCatLady: If you think it's better to add a new string `valdationBotAvatarUrl`, we can do that as well.  Just let me know what you think is the better option here!

sct: Oh I see we are sharing it. Carry on then.

PR #1127, opened 2021-03-08
fix(ui): display "Season" vs. "Seasons" as appropriate, and fix request block "Seasons" formatting

TheCatLady:
#### Description

Fixes cards to display "Season" if only one season is requested (if "ALL" badge is displayed, still shows "Seasons").

Also fixes formatting on request block; "Seasons" text was not the same color as other text in the request block.

image

#### To-Dos

#### Issues Fixed or Closed

N/A


sct:
I actually don't like the plurality here. I think it's fine to say Seasons for both a single selection and multiple.

TheCatLady:
> I actually don't like the plurality here. I think it's fine to say Seasons for both a single selection and multiple.

Ah really?  I think it's really odd and looks like a mistake. 😕  We display the correct plurality for networks/studios on the detail pages, and I really think we should fix this here as well.

sct:
> Ah really?  I think it's really odd and looks like a mistake. 😕  We display the correct plurality for networks/studios on the detail pages, and I really think we should fix this here as well.

I don't agree. Networks/Studios are a different case. Seasons is a list of badges representing each season. Feels better to me to just have it say one thing in both cases. 🤷‍♂️

TheCatLady:
> I don't agree. Networks/Studios are a different case. Seasons is a list of badges representing each season. Feels better to me to just have it say one thing in both cases. 🤷‍♂️

I think it would be particularly bad in cases where a single season is requested and it isn't the first season.  Because then you'd have text saying: "Seasons: [2]" for example, and it could be interpreted as a season count rather than a list of season numbers.

sct (approved):

PR #1137, opened 2021-03-09
feat(notif): add "Media Automatically Approved" notification type

### LINKED ISSUE #1133: Media Requested notifications for auto-approved media
#### Description

It would be nice to be able to be notified of all media requests regardless of auto-approval setting. There are cases where something might be auto-approved but unable to be grabbed automatically, requiring manual intervention. Without a request notification, admin may not even be aware of a problem until the user follows up as to why media is missing.

#### Desired Behavior

Provide an option to send Media Requested notifications for auto-approved media.

TheCatLady:
#### Description

Replaces "Enable Notifications for Automatic Approvals" setting, which was confusing to users despite our attempts to clarify the notification types.

This PR separates notifications for automatically approved requests from notifications for manually approved requests.

Also fixes Discord and Telegram notification agents so that they don't mention/notify users for "Media Requested" and "Media Failed" notification types, since those notifications are intended for admin users.

Additionally, adds media type (i.e., either "movie" or "series") to the subject of notifications.

image

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1133


server/lib/notifications/agents/agent.ts
```diff
@@ -27,3 +27,9 @@ export interface NotificationAgent {
   shouldSend(type: Notification, payload: NotificationPayload): boolean;
   send(type: Notification, payload: NotificationPayload): Promise<boolean>;
 }
+
+export const userNotificationTypes: Notification[] = [
+  Notification.MEDIA_APPROVED,
+  Notification.MEDIA_DECLINED,
+  Notification.MEDIA_AVAILABLE,
+];
```
sct, line 35: What about adding this as a static method/value in the `BaseAgent` since we extend that for every agent anyways? Then you don't need to import this into the other files.

TheCatLady: That definitely sounds better!  Will fix shortly

sct:
It looks like admins will get emails for all of their own requests as well with this. I wonder if that's something people will want?

TheCatLady:
> It looks like admins will get emails for all of their own requests as well with this. I wonder if that's something people will want?

That's no different form the current behavior, however.

server/lib/notifications/agents/discord.ts
```diff
@@ -122,6 +122,7 @@ class DiscordAgent
         });
         break;
       case Notification.MEDIA_APPROVED:
+      case Notification.MEDIA_AUTO_APPROVED:
```
sct, line 125: This will cause a double notification? Right?

sct: Oh, I see. We will only send one type.

TheCatLady: Haha yeah, it's either one or the other.

sct (approved):

PR #1143, opened 2021-03-10
refactor(ui): move user-related settings out of "General Settings"

TheCatLady:
#### Description

The "General Settings" page is getting really long, and user-related settings belong in their own tab.

Also, ~~further reduce page length in some cases by hiding~~ **_disable_** 4K-related permissions if the required 4K servers are not present.

#### To-Dos

#### Issues Fixed or Closed

N/A


sct:
I don't like the idea of hiding permission fields unless certain things are clicked. People won't even know the possibility for those options is there without first interacting with a different option. Even if it's related, I would prefer the possible permissions a user can select are always _visible_. Disabling them is fine, of course.

TheCatLady:
> I don't like the idea of hiding permission fields unless certain things are clicked. People won't even know the possibility for those options is there without first interacting with a different option. Even if it's related, I would prefer the possible permissions a user can select are always _visible_. Disabling them is fine, of course.

This doesn't hide permissions fields on click.  It simply removes permissions for which an Overseerr instance isn't set up to accommodate.  If there are no 4K servers linked, I don't see the reason why we should allow for setting of 4K permissions?  Similar to how the "Request 4K" buttons aren't displayed if there is no 4K server.

sct:
> > I don't like the idea of hiding permission fields unless certain things are clicked. People won't even know the possibility for those options is there without first interacting with a different option. Even if it's related, I would prefer the possible permissions a user can select are always _visible_. Disabling them is fine, of course.
> 
> This doesn't hide permissions fields on click. It simply removes permissions for which an Overseerr instance isn't set up to accommodate. If there are no 4K servers linked, I don't see the reason why we should allow for setting of 4K permissions? Similar to how the "Request 4K" buttons aren't displayed if there is no 4K server.

Because without someone setting up a 4K server, they may never even know it's _possible_ to set those permissions. They can stay disabled, but people should be aware they exist.

src/components/PermissionOption/index.tsx
```diff
@@ -67,7 +70,17 @@ const PermissionOption: React.FC<PermissionOptionProps> = ({
...
+          ((option.permission === Permission.REQUEST_4K ||
+            option.permission === Permission.AUTO_APPROVE_4K) &&
+            (!settings.currentSettings.movie4kEnabled ||
+              !settings.currentSettings.series4kEnabled)) ||
+          ((option.permission === Permission.REQUEST_4K_MOVIE ||
+            option.permission === Permission.AUTO_APPROVE_4K_MOVIE) &&
+            !settings.currentSettings.movie4kEnabled) ||
+          ((option.permission === Permission.REQUEST_4K_TV ||
+            option.permission === Permission.AUTO_APPROVE_4K_TV) &&
+            !settings.currentSettings.series4kEnabled)
```
sct, line 83: This is very difficult to understand. I think it's time we break out these checks from being inline. Instead of repeating ourselves constantly in this file.

TheCatLady: Done, and added comments to clarify each check 😺

sct (approved):

PR #1156, opened 2021-03-11
refactor(css): add new global classes & card border hover styling

TheCatLady:
#### Description

Cleaned up the slider header CSS classes a bit, and created global classes for those as well as the `div` elements wrapping form text fields.

Also added a hover effect to the new card borders.

#### To-Dos

#### Issues Fixed or Closed

N/A


sct:
Hover styles applied to cards should be done with the `isHovered` (or whatever the relevant name is in the component) condition.

TheCatLady:
> Hover styles applied to cards should be done with the `isHovered` (or whatever the relevant name is in the component) condition.

Fixed!

sct (approved):

PR #1160, opened 2021-03-12
feat: add tagline, episode runtime, genres list to media details & clean/refactor CSS into globals

TheCatLady:
#### Description

- Add taglines to movie and series detail pages
- Add episode runtime to series detail pages
- Add genres to collection detail pages
- Move external links block into "media facts" table
- Apply slider CSS global classes to cast slider
- Clean up and refactor CSS used in movie, series, and collection detail pages into `globals.css`
- Use consistent date formatting throughout the application

image

#### To-Dos

#### Issues Fixed or Closed

N/A


sct:
image
Introduced this issue

TheCatLady:
> image
> Introduced this issue

Ah, will fix that!

sct:
image

Not sure what happened, but I am unable to click on these arrows in this PR.

sct:
image
Badge is off center

src/components/PersonDetails/index.tsx
```diff
@@ -15,8 +15,8 @@ import { groupBy } from 'lodash';
 import PageTitle from '../Common/PageTitle';
 
 const messages = defineMessages({
-  appearsin: 'Appears in',
-  crewmember: 'Crew Member',
+  appearsin: 'Acting',
```
sct, line 18: Not everything under here is "acting." There are talk shows, documentaries, etc.

src/components/PersonDetails/index.tsx
```diff
@@ -15,8 +15,8 @@ import { groupBy } from 'lodash';
...
+  crewmember: 'Production',
```
sct, line 19: The same goes here I think. Not everything is always specifically "Production"?

src/components/RequestModal/AdvancedRequester/index.tsx
```diff
@@ -18,12 +18,10 @@ const messages = defineMessages({
   qualityprofile: 'Quality Profile',
   rootfolder: 'Root Folder',
   animenote: '* This series is an anime.',
-  default: '(Default)',
-  loadingprofiles: 'Loading profiles…',
-  loadingfolders: 'Loading folders…',
+  default: ' (Default)',
```
sct, line 21: We should leave formatting like this out of strings.

TheCatLady: It's not formatting, however; it's actually a part of the string.  And in some languages (like Chinese), not having this be a part of the string would result in whitespace where there shouldn't be any.

sct: But then you can argue that default shouldn't be a separate string at all? I think there can be some compromise here. I don't really like having spacing like this decided by the translation strings.

It would be better to have a string that is for default that takes the default value as variable.

"{value} (Default)" for example.

TheCatLady: Yes, I definitely think that's the better option! I thought about doing that, but figured it might be easier for our translators to just edit the space.  Will fix shortly!

sct (approved):

PR #1170, opened 2021-03-14
feat(frontend): add loading bar indicator

ankarhem:
#### Description

Adds a loading indicator for navigation

#### To-Dos

### DIFF
--- a/src/components/LoadingBar/index.tsx
+++ b/src/components/LoadingBar/index.tsx
@@ -1,0 +1,86 @@
+import { NProgress } from '@tanem/react-nprogress';
+import React, { useEffect, useState } from 'react';
+import ReactDOM from 'react-dom';
+import { useRouter } from 'next/router';
+
+interface BarProps {
+  progress: number;
+  isFinished: boolean;
+}
+
+const domAvailable = !!(
+  typeof window !== 'undefined' &&
+  window.document &&
+  window.document.createElement
+);
+
+const Bar = ({ progress, isFinished }: BarProps) => {
+  return (
+    <div
+      className={`fixed top-0 left-0 z-50 w-full transition-opacity ease-out duration-400 ${
+        isFinished ? 'opacity-0' : 'opacity-100'
+      }`}
+    >
+      <div
+        className="duration-300 bg-indigo-400 transition-width"
+        style={{
+          height: '3px',
+          width: `${progress * 100}%`,
+        }}
+      />
+    </div>
+  );
+};
+
+/* eslint-disable */
+const MemoizedNProgress = React.memo(({ loading }: { loading: boolean }) => (
+  <NProgress isAnimating={loading}>
+    {({ isFinished, progress }) => (
+      <Bar progress={progress} isFinished={isFinished} />
+    )}
+  </NProgress>
+));
+/* eslint-enable */
+
+const LoadingBar = (): React.ReactPortal | null => {
+  const [mounted, setMounted] = useState(false);
+  const [loading, setLoading] = useState(false);
+  const router = useRouter();
+  useEffect(() => {
+    setMounted(true);
+  }, []);
+
+  useEffect(() => {
+    const handleLoading = () => {
+      setLoading(true);
+    };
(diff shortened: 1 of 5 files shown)


src/components/Layout/Sidebar/index.tsx
```diff
@@ -175,9 +175,11 @@ const Sidebar: React.FC<SidebarProps> = ({ open, setClosed }) => {
...
                     <div className="flex items-center flex-shrink-0 px-4">
                       <span className="text-xl text-gray-50">
-                        <a href="/">
-                          <img src="/logo.png" alt="Logo" />
-                        </a>
+                        <Link href="/">
+                          <a>
+                            <img src="/logo.png" alt="Logo" />
+                          </a>
+                        </Link>
```
sct, line 182: This was intentional to force reload the page if the logo is clicked. Helps with apps that are bookmarked on mobile devices where there is no refresh button.

sct (changes requested):

src/components/Layout/Sidebar/index.tsx
```diff
@@ -237,9 +239,11 @@ const Sidebar: React.FC<SidebarProps> = ({ open, setClosed }) => {
...
               <div className="flex items-center flex-shrink-0 px-4">
                 <span className="text-2xl text-gray-50">
-                  <a href="/">
-                    <img src="/logo.png" alt="Logo" />
-                  </a>
+                  <Link href="/">
+                    <a>
+                      <img src="/logo.png" alt="Logo" />
+                    </a>
+                  </Link>
```
sct, line 246: Same thing here. Not mobile but serves the same purpose.

src/components/LoadingBar/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+        style={{
+          height: '3px',
+          width: `${progress * 100}%`,
+        }}
+      />
+    </div>
+  );
+};
+
+/* eslint-disable */
```
sct, line 35: Why are we disabling eslint here?

src/components/Layout/Sidebar/index.tsx
ankarhem: Alright, I'll change it back.

src/components/LoadingBar/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  }, []);
+
+  return mounted && domAvailable
+    ? ReactDOM.createPortal(
+        <MemoizedNProgress loading={loading} />,
+
+        // This will always exist :)
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        document.getElementById('__next')
```
sct, line 81: Any reason we can't just use `document.body` here and avoid having to do ts-ignores?

src/components/LoadingBar/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+    return () => {
+      router.events.off('routeChangeStart', handleLoading);
+      router.events.off('routeChangeComplete', handleFinishedLoading);
+      router.events.off('routeChangeError', handleFinishedLoading);
+    };
+
+    // Only want this to run once on mount
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
sct, line 72: You want `router` as a dependency here. Not having it could introduce bugs in the future if the router ever changes for any reason.

src/components/LoadingBar/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  return mounted && domAvailable
```
sct, line 74: Since you are already setting `mounted` to true in a `useEffect`, it shouldn't be necessary to check if the dom is available. It _must_ be available because `useEffect` only runs once the component is mounted on the client.

src/components/LoadingBar/index.tsx
sct: I assume because the component doesn't have a displayName? Use a named function instead to get around this.

sct (approved):
💯

PR #1172, opened 2021-03-15
feat(notif): allow users to enable/disable specific agents

### LINKED ISSUE #993: Specify notification agent per user
#### Description
It would be nice to have the flexibility to specify which notification to use for each user. When having multiple notification agents enabled it will send notifications to all notification agents to all users.

#### Desired Behavior
Specify for each user which notification agent to use as a primary one.

### LINKED ISSUE #1176: Increase verboseness of notification logs
#### Description
The logs related to notification today are very basic and aren't very useful when it comes to troubleshooting
That's an example:
```
2021-03-14T19:20:17.356Z [info][Notifications]: Sending notification for MEDIA_AVAILABLE 
2021-03-14T19:20:17.357Z [debug][Notifications]: Sending Telegram notification 
2021-03-14T19:20:18.202Z [error][Notifications]: Request failed with status code 403 
```
I can't answer who this notification was sent to, what movie was available, and what 403 means.

#### Desired Behavior
It would be nice if we had something similar to this:

```
2021-03-14T19:20:17.356Z [info][Notifications]: Sending notification for MEDIA_AVAILABLE 
2021-03-14T19:20:17.357Z [debug][Notifications]: Sending {notification agent} notification to {user} - {movie} - {status}
2021-03-14T19:20:18.202Z [error][Notifications]: Request failed with status code 403
dump here the agent error message
```
Resulting in:
```
2021-03-14T19:20:17.356Z [info][Notifications]: Sending notification for MEDIA_AVAILABLE 
2021-03-14T19:20:17.357Z [debug][Notifications]: Sending Telegram notification to ankarhem - To Sir, with Love (1967) - Available
2021-03-14T19:20:18.202Z [error][Notifications]: Request failed with status code 403
{"ok":false,"error_code":400,"description":"Bad Request: chat not found"}
```
#### Additional Context

I requested movies in the name of some users and now I can't tell which of them wasn't notified.

TheCatLady:
#### Description

- Allows users to enable/disable specific agents
- Refactors setting navigation tabs into a new `SettingsTabs` component
- Adds tabbed user notification settings and hides tabs for settings which cannot be enabled (e.g., don't allow user to configure email settings if the owner/admin has not set up and enabled email notifications)
- Adds new shared settings tabs component for settings navigation elements
- Modifies `Yup` validation to only enforce required fields when the notification agent is enabled
- Refactors email notification agent to reduce duplicated logic/code
- Improves notification logging
- Removes redundant "Enable Notifications" setting, since each agent can be enabled/disabled on its own
- Fixes test function for email notification agent to use form values instead of saved settings

image
image
image

#### To-Dos

#### Issues Fixed or Closed

- Closes #993
- Closes #1176


src/components/Settings/Notifications/NotificationsTelegram.tsx
```diff
@@ -86,6 +86,8 @@ const NotificationsTelegram: React.FC = () => {
               botUsername: values.botUsername,
             },
           });
+          mutate('/api/v1/settings/public');
```
sct, line 89: Why are we calling mutate here?

TheCatLady: The global/system Telegram notifications need to be configured in order for users to configure personal notifications, since user notifications are dependent on the bot username & bot API key.

sct: I don't think we need to provide any of that data into the public settings endpoint? It can just be returned with the API calls from the settings page itself.

TheCatLady: Oh hm, good point.  I'll change that!

src/components/UserProfile/UserSettings/index.tsx
```diff
@@ -48,12 +44,12 @@ const UserSettings: React.FC = ({ children }) => {
   const settingsRoutes: SettingsRoute[] = [
     {
       text: intl.formatMessage(messages.menuGeneralSettings),
-      route: '/settings/main',
+      route: `/users/${user?.id}/settings/main`,
```
sct, line 47: The changes to the routes here break the profile routes.

server/entity/UserSettings.ts
```diff
@@ -20,24 +24,28 @@ export class UserSettings {
   @JoinColumn()
   public user: User;
 
-  @Column({ default: true })
-  public enableNotifications: boolean;
-
   @Column({ nullable: true })
-  public discordId?: string;
+  public region?: string;
```
sct, line 29: Please don't reorganize things unless it serves a purpose. This makes the PR really hard to read.

server/lib/notifications/agents/discord.ts
```diff
@@ -177,6 +181,7 @@ class DiscordAgent
         ...(payload.extra ?? []).map((extra) => ({
           name: extra.name,
           value: extra.value,
+          inline: true,
```
sct, line 184: Because this can be any arbitrary data, I felt inlining it wouldn't make sense if we decided to attach longer messages.

server/migration/1616699371961-UserSettingsRemoveEnableNotifications.ts
```diff
@@ -0,0 +1,32 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
sct, line 1: This file needs to be regenerated since we have had another migration in the meantime.

server/migration/1616699467321-UserSettingsAddNotificationAgents.ts
```diff
@@ -0,0 +1,32 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
sct, line 1: Why do we have two migrations?

server/routes/settings/notifications.ts
```diff
@@ -1,36 +1,16 @@
...
-
-  Object.assign(settings.notifications, {
-    enabled: req.body.enabled,
-  });
-  settings.save();
-
-  return res.status(200).json({
-    enabled: settings.notifications.enabled,
-  });
-});
```
sct, line 32: I think I misunderstood you before. I didn't want to remove the global notification disable option. I just wanted to remove it for the users.

src/components/Settings/Notifications/NotificationsEmail.tsx
```diff
@@ -64,21 +62,62 @@ const NotificationsEmail: React.FC = () => {
...
+        .when('pgpPassword', {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          is: (value: any) => !!value,
+          then: Yup.string()
+            .nullable()
+            .required(intl.formatMessage(messages.validationPgpPrivateKey)),
+          otherwise: Yup.string().nullable(),
+        })
+        .matches(
+          /^-----BEGIN PGP PRIVATE KEY BLOCK-----[\s\w]+-----END PGP PRIVATE KEY BLOCK-----$/,
```
sct, line 107: Hrm. I am not sure how entirely strict the spacing format is for this block. Do we really need validation here? People should know what they are doing.

What about a placeholder example in the text area instead?

src/components/Settings/Notifications/NotificationsEmail.tsx
```diff
@@ -64,21 +62,62 @@ const NotificationsEmail: React.FC = () => {
...
+          intl.formatMessage(messages.validationPgpPrivateKey)
+        ),
+      pgpPassword: Yup.string().when('pgpPrivateKey', {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        is: (value: any) => !!value,
```
sct, line 112: Better to not disable eslint if we can avoid it. Instead make the type here `unknown`:

```suggestion
        is: (value: unknown) => !!value,
```

src/components/Settings/Notifications/NotificationsEmail.tsx
```diff
@@ -64,21 +62,62 @@ const NotificationsEmail: React.FC = () => {
...
+          is: true,
+          then: Yup.number().required(
+            intl.formatMessage(messages.validationSmtpPortRequired)
+          ),
+          otherwise: Yup.number().nullable(),
+        }),
+      pgpPrivateKey: Yup.string()
+        .when('pgpPassword', {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          is: (value: any) => !!value,
```
sct, line 100: Same thing here with type `unknown` instead of any.

src/components/UserProfile/UserSettings/index.tsx
```diff
@@ -67,50 +63,20 @@ const UserSettings: React.FC = ({ children }) => {
     },
     {
       text: intl.formatMessage(messages.menuNotifications),
-      route: '/settings/notifications',
+      route: data?.emailEnabled
+        ? `/users/${user?.id}/settings/notifications/email`
+        : `/users/${user?.id}/settings/notifications/discord`,
```
sct, line 68: What if neither are enabled? Should we show the discord notification page if discord notifications aren't enabled?

src/components/Settings/Notifications/NotificationsWebhook/index.tsx
```diff
@@ -60,14 +59,26 @@ const NotificationsWebhook: React.FC = () => {
...
   const NotificationsWebhookSchema = Yup.object().shape({
     webhookUrl: Yup.string()
-      .required(intl.formatMessage(messages.validationWebhookUrl))
+      .when('enabled', {
+        is: true,
+        then: Yup.string()
+          .nullable()
+          .required(intl.formatMessage(messages.validationWebhookUrl)),
+        otherwise: Yup.string().nullable(),
+      })
```
sct, line 68: You made this sort of change in many places, but I am unsure how I feel about it UX-wise.

Users could actually save invalid data, but when they went to enable it, suddenly errors would spring up, telling them what they entered was wrong. But it just let them save it before? I think the validation should always be on if the inputs are visible. If we want to "remove" user input when the notification agent isn't enabled, it would be better to have something that hides all the inputs unless the type is actually enabled (which would make this toggled validation make sense)—something to think about for another PR.

src/components/UserProfile/UserSettings/index.tsx
TheCatLady: The user can always enter a Discord ID, even if Discord notifications are not enabled. Some people may be using Requestrr and not have our built-in Discord notifications enabled.

src/components/Settings/Notifications/NotificationsWebhook/index.tsx
TheCatLady: So this doesn't actually allow them to save invalid data. Rather, it just makes them optional (doesn't apply `required()`).  If a user enters a value in the field, it still must pass the valdation after/outside the `when()`.

src/components/Settings/Notifications/NotificationsEmail.tsx
TheCatLady: According to my research, the PGP private key must always begin with `-----BEGIN PGP PRIVATE KEY BLOCK-----` and end with `-----END PGP PRIVATE KEY BLOCK-----`. The spacing and characters and length of the block in between can vary, however, which is why I went with `[\s\w]+` in between (to match 1 or more space/word characters).

A placeholder might be confusing since we'd need to include both the header/footer of the key, and it would disappear once the user inputted a value into the field. When we add the help system, we should probably have some clarification in there about the format of this field.

src/components/UserProfile/UserSettings/index.tsx
TheCatLady: Could you clarify? I didn't find/notice anything broken as a result.  The paths just had to change since I updated it to use a shared component for the navigation.

src/components/UserProfile/UserSettings/index.tsx
sct: Ah yes, Requestrr! Makes sense!

src/components/UserProfile/UserSettings/index.tsx
sct: When a user navigates to their own settings from their own profile or from the user dropdown, the URL is different:

`/profile/settings/notifications`

TheCatLady: @sct Fixed!

src/pages/profile/settings/notifications/discord.tsx
```diff
@@ -0,0 +1,20 @@
+import { NextPage } from 'next';
+import React from 'react';
+import UserSettings from '../../../../components/UserProfile/UserSettings';
+import UserNotificationSettings from '../../../../components/UserProfile/UserSettings/UserNotificationSettings';
+import UserNotificationsDiscord from '../../../../components/UserProfile/UserSettings/UserNotificationSettings/UserNotificationsDiscord';
+import useRouteGuard from '../../../../hooks/useRouteGuard';
+import { Permission } from '../../../../hooks/useUser';
+
+const NotificationsPage: NextPage = () => {
+  useRouteGuard(Permission.MANAGE_USERS);
```
sct, line 10: How do users get to their own management page with this guard?

sct: Same goes for the other profile pages

TheCatLady: Ah whoops, copy-paste error.

sct (approved):

PR #1197, opened 2021-03-16
refactor(ui): tweak genre card colors

TheCatLady:
#### Description

Makes some small adjustments to reduce the numbers of adjacent cards with the same/similar colors.

image
image

#### To-Dos

#### Issues Fixed or Closed

N/A


sct:
This also depends on your screen size, right? You could have adjacent colors just the same with a wider screen.

sct:
> This also depends on your screen size, right? You could have adjacent colors just the same with a wider screen.

I did some testing. Similar colors still line up but just with really wide screens and there really isn't much we can do about that I suppose 😂

sct (approved):

PR #1202, opened 2021-03-16
fix(lang): UI string edits, round 2

TheCatLady:
#### Description

More language string fixes, for things I didn't catch the last time or which were introduced since that first round of string edits.  Apologies @sct and translators 👀

Also in this PR:
- set more descriptive page titles for settings, user settings, and error pages
- add links back to the media details page from the "Recommendations" and "Similar <Movies/Series>" pages
- fixed `Yup` validation on port number fields to not show errors like "port must be a `number` type, but the final value was: `NaN` (cast from the value `...`)."
- removed all remaining instances of `FormattedMessage` and `FormattedNumber`
- swapped `ReactMarkdown` for `ReactMarkdownWithHtml` so that images defined using HTML `img` tags in release notes get displayed properly

#### To-Dos

#### Issues Fixed or Closed

N/A


src/components/Settings/SettingsAbout/Releases/index.tsx
```diff
@@ -18,13 +18,14 @@ const messages = defineMessages({
...
+  runningDevelop: 'Development Version',
   runningDevelopMessage:
-    'The changes in your version will not be available below. Please see the <GithubLink>GitHub repository</GithubLink> for latest updates.',
+    'The latest changes to the <code>develop</code> branch of Overseerr are not shown below.\
+    Please see the commit history for this branch on <GithubLink>GitHub</GithubLink> for details.',
 });
 
 const REPO_RELEASE_API =
-  'https://api.github.com/repos/sct/overseerr/releases?per_page=20';
+  'https://api.github.com/repos/sct/overseerr/releases?per_page=10';
```
sct, line 28: Why 10?

src/components/Settings/SettingsAbout/Releases/index.tsx
```diff
@@ -95,7 +96,9 @@ const Release: React.FC<ReleaseProps> = ({
           }}
         >
           <div className="prose">
-            <ReactMarkdown>{release.body}</ReactMarkdown>
+            <ReactMarkdownWithHtml allowDangerousHtml>
+              {release.body}
+            </ReactMarkdownWithHtml>
```
sct, line 101: Was something not working before that we need to parse HTML and allow dangerous HTML?

src/components/Settings/SettingsAbout/index.tsx
```diff
@@ -86,7 +94,7 @@ const SettingsAbout: React.FC = () => {
               rel="noreferrer"
               className="text-indigo-500 hover:underline"
             >
-              {intl.formatMessage(messages.clickheretojoindiscord)}
+              https://discord.gg/PkCWJSeCk7
```
sct, line 97: What was wrong with the descriptive string?

src/components/Settings/SettingsAbout/Releases/index.tsx
TheCatLady: Yes, there were images in some release notes that were included using HTML.

src/components/Settings/SettingsAbout/index.tsx
TheCatLady: It doesn't match the pattern on that page:
image

src/components/Settings/SettingsAbout/Releases/index.tsx
TheCatLady: I looked at how often we release and didn't think it was useful to display that many.  The page looks cleaner and less intimidating with a more reasonable number of entries IMO.  There are currently only 19 entries _total_, so we aren't even hitting that limit of 20, and I think that just covers a really long period of time that wouldn't be useful to the user... so much has changed in that time!

src/components/Settings/SettingsAbout/index.tsx
sct: There is no set-in-stone pattern for this page, though. I don't like the invite URL being just the plain invite code because technically, it can change. Unless we get level 3 boosting going on and have a static URL, I like the other string better.

sct: I suppose that could happen? Either way, it's not important. No need to bikeshed over the specific text of a link. If you like the URL more, we can go with that.

src/components/Settings/SettingsLogs/index.tsx
```diff
@@ -9,9 +9,11 @@ import Error from '../../../pages/_error';
 import Badge from '../../Common/Badge';
 import Button from '../../Common/Button';
 import LoadingSpinner from '../../Common/LoadingSpinner';
+import PageTitle from '../../Common/PageTitle';
 import Table from '../../Common/Table';
 
 const messages = defineMessages({
+  settings: 'Settings',
```
sct, line 16: Maybe a global message for `Settings` is better here.

src/components/Settings/SettingsMain.tsx
```diff
@@ -24,29 +27,27 @@ const messages = defineMessages({
   applicationTitle: 'Application Title',
   applicationurl: 'Application URL',
   region: 'Discover Region',
-  regionTip:
-    'Filter content by region (only applies to the "Popular" and "Upcoming" categories)',
```
sct, line 28: Are these tips no longer relevant?

TheCatLady: I didn't remove the tip, just the parts in parentheses.  We are also filtering results for genres now, and possibly more areas in the future.  We can't possibly list them all, and I think we only filter where it makes sense, so I didn't feel the category clarification was necessary anymore.

I did change the region tip to say "filter content by **regional availability**" though, since that is a bit more clear on what the region setting actually does.

sct (approved):

PR #1208, opened 2021-03-17
fix(ui): use appropriate cursor type for readonly input fields

TheCatLady:
#### Description

Makes it clearer that `readonly` input fields cannot be interacted with, since otherwise they appear like regular text boxes.  (Currently, there are only two such fields... the API key in General Settings, and the server name in Plex Settings.)

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/src/components/Settings/SettingsMain.tsx
+++ b/src/components/Settings/SettingsMain.tsx
@@ -181,7 +181,7 @@ const SettingsMain: React.FC = () => {
                         <input
                           type="text"
                           id="apiKey"
-                          className="rounded-l-only"
+                          className="cursor-not-allowed rounded-l-only"
                           value={data?.apiKey}
                           readOnly
                         />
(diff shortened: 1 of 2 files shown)


src/components/Settings/SettingsMain.tsx
```diff
@@ -181,7 +181,7 @@ const SettingsMain: React.FC = () => {
                         <input
                           type="text"
                           id="apiKey"
-                          className="rounded-l-only"
+                          className="cursor-not-allowed rounded-l-only"
```
sct, line 184: This field can be interacted with, though. You can click and drag to select inside the box and copy manually if that's your jam. Having the "not allowed" cursor makes this more difficult.

TheCatLady: Ah.  Perhaps we shouldn't modify the API key field, in that case.  I still think we should change the cursor for the Plex server name though, since that one is kinda confusing and I don't see why you'd want to copy that?

sct: Yes. I only commented about the API key field after all :)

sct (approved):

PR #1211, opened 2021-03-17
feat(notif): include requested season numbers in notifications

### LINKED ISSUE #315: Include Season Information with Notification of Show Being Available
After requesting a season of a TV show, the email notification does not communicate which season(s) is available.

In the email, communicate the seasons that are now available to watch

Can't think of any

**Additional context**
Screen Shot 2020-12-15 at 8 43 56 AM

TheCatLady:
#### Description

- Adds requested season numbers to all notification types (only Discord notifications included this information previously)
- Sets `notifyUser` only for user-specific notification types (`MEDIA_APPROVED`, `MEDIA_DECLINED`, `MEDIA_AVAILABLE`), since otherwise notifications will go to all admins
- Fixes email agent using the PGP key of the media requester rather than that of the email recipient for admin notifications
- Includes media plot/overview in `MEDIA_FAILED` notifications, since we are already showing a failure message in the subject of those notifications

#### To-Dos

#### Issues Fixed or Closed

- Closes #315

### DIFF
--- a/server/entity/MediaRequest.ts
+++ b/server/entity/MediaRequest.ts
@@ -179,7 +179,7 @@ export class MediaRequest {
             subject: movie.title,
             message: movie.overview,
             image: `https://image.tmdb.org/t/p/w600_and_h900_bestv2${movie.poster_path}`,
-            notifyUser: this.requestedBy,
+            notifyUser: autoApproved ? undefined : this.requestedBy,
             media,
             request: this,
           }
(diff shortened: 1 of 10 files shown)


server/entity/MediaRequest.ts
```diff
@@ -444,15 +444,10 @@ export class MediaRequest {
...
-            const admin = await userRepository.findOneOrFail({
-              select: ['id', 'plexToken'],
-              order: { id: 'ASC' },
-            });
+
             notificationManager.sendNotification(Notification.MEDIA_FAILED, {
               subject: movie.title,
-              message: 'Movie failed to add to Radarr',
-              notifyUser: admin,
+              message: movie.overview,
```
sct, line 450: Do we really need the overview on the failed notification? What was wrong with how it was before?

TheCatLady: The notifications themselves already say that the the request failed, so I feel the original message here is unnecessary.  We don't have to include the plot/overview in the actual notification, but it doesn't hurt, and it would then match with the other media notifications.  I personally like having the plot, since it's in every other notification type, and OCD... 😅

TheCatLady: TBH though, setting the subject/message here at all is kind of unnecessary for media notifications, since we also pass the media object in the payload.

sct: Perhaps. I originally tried to abstract the need to pass along specific objects to display any notification flexibly. I ended up not going that way, so this is a legacy decision. We can change it in the future.

sct (approved):

PR #1240, opened 2021-03-20
feat(ui): display movie/series original title

TheCatLady:
#### Description

Displays the original title of a movie/series if its original language does not match the user's locale.

Also fixes an issue where there's an extra margin above the summary/overview when the tagline is missing.

- **Example where locale does not match media original language:**
image
- **Example where locale matches media original language:**
   image

#### To-Dos

#### Issues Fixed or Closed

N/A


src/components/TvDetails/index.tsx
```diff
@@ -615,6 +617,12 @@ const TvDetails: React.FC<TvDetailsProps> = ({ tv }) => {
                 )}
               </div>
             )}
+            {data.originalName && data.originalLanguage !== locale.slice(0, 2) && (
```
sct, line 620: are the local languages _always_ two-character codes?

sct:
```suggestion
            {data.originalName && data.originalLanguage.startsWith(locale.slice(0,2)) && (
```
maybe?

sct (approved):

src/components/TvDetails/index.tsx
TheCatLady: API documentation here: https://developers.themoviedb.org/3/configuration/get-languages

PR #1241, opened 2021-03-20
fix(lang): edit search placeholder to reflect that it also returns person results

TheCatLady:
#### Description

"Search Movies & TV" isn't 100% accurate, given that the search also returns person results.  Also, we don't use "TV" anywhere else in the UI, instead opting for the term "Series," so it would be better to use that same language here.

#### To-Dos

#### Issues Fixed or Closed

N/A


TheCatLady:
Closing since the suggested string wouldn't fit on mobile. 😿

PR #1243, opened 2021-03-20
fix: fetch localized person details from TMDb

TheCatLady:
#### Description

We aren't setting the `language` parameter when fetching person details, so the biography is currently always in English.

**Edit:** Also, add missing person details (DOB, hometown, alternate names).

#### To-Dos

#### Issues Fixed or Closed

N/A


src/components/PersonDetails/index.tsx
```diff
@@ -82,6 +85,43 @@ const PersonDetails: React.FC = () => {
...
+            month: 'long',
+            day: 'numeric',
+          }),
+          deathdate: data.deathday
+            ? intl.formatDate(data.deathday, {
+                year: 'numeric',
+                month: 'long',
+                day: 'numeric',
+              })
+            : null,
```
sct, line 105: You don't need to check if `data.deathday` is set here.

TheCatLady: Oh yup, that was leftover from before I changed the string. Will fix shortly!

sct (approved):

PR #1262, opened 2021-03-22
feat(ui): store sort order and page size of userlist in localstorage

danshilm:
#### Description

Use localStorage to remember the sort order and page size of the user list just like it's done for the requests list and logs.

N/A

#### To-Dos

#### Issues Fixed or Closed

- Fixes none

### DIFF
--- a/src/components/Settings/SettingsLogs/index.tsx
+++ b/src/components/Settings/SettingsLogs/index.tsx
@@ -77,7 +77,7 @@ const SettingsLogs: React.FC = () => {
   const { data: appData } = useSWR('/api/v1/status/appdata');
 
   useEffect(() => {
-    const filterString = window.localStorage.getItem('logs-filter-settings');
+    const filterString = window.localStorage.getItem('logs-display-settings');
 
     if (filterString) {
       const filterSettings = JSON.parse(filterString);
@@ -89,7 +89,7 @@ const SettingsLogs: React.FC = () => {
 
   useEffect(() => {
     window.localStorage.setItem(
-      'logs-filter-settings',
+      'logs-display-settings',
       JSON.stringify({
         currentFilter,
         currentPageSize,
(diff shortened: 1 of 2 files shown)


src/components/UserList/index.tsx
```diff
@@ -110,6 +110,29 @@ const UserList: React.FC = () => {
...
+
+  useEffect(() => {
+    window.localStorage.setItem(
+      'userlist-display-settings',
+      JSON.stringify({
+        currentSort,
+        currentPageSize,
+      })
+    );
+  }, [currentSort, currentPageSize]);
```
TheCatLady, line 134:
```suggestion
  useEffect(() => {
    const filterString = window.localStorage.getItem('ul-filter-settings');

    if (filterString) {
      const filterSettings = JSON.parse(displayString);

      setCurrentSort(filterSettings.currentSort);
      setCurrentPageSize(filterSettings.currentPageSize);
    }
  }, []);

  useEffect(() => {
    window.localStorage.setItem(
      'ul-filter-settings',
      JSON.stringify({
        currentSort,
        currentPageSize,
      })
    );
  }, [currentSort, currentPageSize]);
```
We should probably follow the same naming pattern used for the request list.

danshilm: I don't think this should be `x-filter-settings` since this does not only save the filter/sort, but the page size too.
I can rename it to `ul-display-settings` though.

TheCatLady: @danshilm Page size is actually also a filter, however.  Same as with the requests list?

src/components/Settings/SettingsLogs/index.tsx
```diff
@@ -77,19 +77,19 @@ const SettingsLogs: React.FC = () => {
   const { data: appData } = useSWR('/api/v1/status/appdata');
 
   useEffect(() => {
-    const displayString = window.localStorage.getItem('logs-display-settings');
+    const filterString = window.localStorage.getItem('logs-filter-settings');
```
sct, line 80: I don't mind changing `displayString` to `filterString` but since we already committed to `logs-display-settings` lets just stick to it. Otherwise, people will have that value sitting around in their local storage forever.

sct (approved):

PR #1280, opened 2021-03-24
refactor(ui): improved alerts for services settings & removed alert titles

TheCatLady:
#### Description

Removes alert titles, eliminating strings that weren't particularly meaningful/helpful and also giving the application a more uniform look throughout.

Also splits the alerts shown on the Services settings page into two types:
1. user does not have any default servers
2. user has a default server but not a default non-4K server.

image

#### To-Dos

#### Issues Fixed or Closed

N/A


src/components/Settings/Notifications/NotificationsEmail.tsx
```diff
@@ -160,38 +159,40 @@ const NotificationsEmail: React.FC = () => {
...
+                            <strong className="font-normal text-indigo-100">
+                              {msg}
+                            </strong>
+                          );
+                        },
+                      }
+                    )}
+                  </p>
+                </>
+              }
```
sct, line 193: I think this makes more sense as a child of Alert instead of all in the title? Title isn't required.

TheCatLady: I'll share a screenshot of it in a bit, but since all the alerts are bolded (I literally changed all of them), it looks better as the title so that it matches the other notification agent alerts.

TheCatLady: image
image

TheCatLady: Alternatively, we could make all of the alerts set the message/body instead of the title?  Let me know what you think is better.  I sort of like that the alerts are a heavier font weight, makes them easier to read on the brightly-colored background.

sct (approved):

PR #1282, opened 2021-03-24
fix(ui): restore saved states of quota override checkboxes

TheCatLady:
#### Description

Currently, the "Enable Override" checkboxes reset to unchecked when you return to the user settings page after  saving.  (This is just a visual bug, the values are saved properly to the DB.)

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/src/components/UserProfile/UserSettings/UserGeneralSettings/index.tsx
+++ b/src/components/UserProfile/UserSettings/UserGeneralSettings/index.tsx
@@ -1,7 +1,7 @@
 import axios from 'axios';
 import { Field, Form, Formik } from 'formik';
 import { useRouter } from 'next/router';
-import React, { useMemo, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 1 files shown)


src/components/UserProfile/UserSettings/UserGeneralSettings/index.tsx
```diff
@@ -57,6 +57,11 @@ const UserGeneralSettings: React.FC = () => {
     user ? `/api/v1/user/${user?.id}/settings/main` : null
   );
 
+  useEffect(() => {
+    setMovieQuotaEnabled(!!data?.movieQuotaLimit && !!data?.movieQuotaDays);
+    setTvQuotaEnabled(!!data?.tvQuotaLimit && !!data?.tvQuotaDays);
```
sct, line 62: You must check if the limits are null/undefined here or unlimited won't check the box.

sct (changes requested):

sct (approved):

PR #1314, opened 2021-03-29
fix(backend): fix getShowByTvdbId() error message

TheCatLady:
#### Description

Currently, we are logging messages such as the following:
```
2021-03-29T08:30:25.985Z [error][Sonarr Scan]: Failed to process Sonarr media {"errorMessage":"[TMDb] Failed to get TV show using the external TVDB ID: [TMDb] Failed to find a TV show with the provided TVDB ID: 309496","title":"Rock Records In Love"}
```

This is because we are throwing an error in `getShowByTvdbId()` which gets caught by a `try`/`catch` block which then re-throws and adds to the message.

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/server/api/themoviedb/index.ts
+++ b/server/api/themoviedb/index.ts
@@ -11,6 +11,7 @@ import {
   TmdbNetwork,
   TmdbPersonCombinedCredits,
   TmdbPersonDetail,
+  TmdbProductionCompany,
   TmdbRegion,
   TmdbSearchMovieResponse,
   TmdbSearchMultiResponse,
@@ -18,7 +19,6 @@ import {
   TmdbSeasonWithEpisodes,
   TmdbTvDetails,
   TmdbUpcomingMoviesResponse,
-  TmdbProductionCompany,
 } from './interfaces';
 
 interface SearchOptions {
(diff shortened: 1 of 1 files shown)


server/api/themoviedb/index.ts
```diff
@@ -613,15 +613,15 @@ class TheMovieDb extends ExternalAPI {
...
-      );
     } catch (e) {
       throw new Error(
         `[TMDb] Failed to get TV show using the external TVDB ID: ${e.message}`
       );
     }
+
+    throw new Error(
+      `[TMDb] Failed to find a TV show with the provided TVDB ID: ${tvdbId}`
+    );
```
sct, line 624: Not sure what the point of moving the error down here does? Instead why not change the thrown error message on lines 616 to say something like `No show returned from API for ID <id>`  or something along those lines? Making the full message read `[TMDb] Failed to get TV show using the external TVDB ID: No show returned from API for ID <id>`

TheCatLady: Yeah, sure!  I just wanted to remove the redundant part.

PR #1335, opened 2021-03-30
fix(ui): hide alert when email notifs are already configured

TheCatLady:
#### Description

- If email notifications are already enabled, do not show alert about the email notification requirement for automatically generating passwords
- If email notifications are not enabled, do not show the checkbox for automatically generating passwords
- Fix capitalization of the checkbox label

#### To-Dos

#### Issues Fixed or Closed

N/A


src/components/UserList/index.tsx
```diff
@@ -360,19 +368,21 @@ const UserList: React.FC = () => {
...
+                        <Field
+                          type="checkbox"
+                          id="genpassword"
+                          name="genpassword"
+                          onClick={() => setFieldValue('password', '')}
+                        />
+                      </div>
                     </div>
-                  </div>
+                  )}
```
sct, line 385: I think it's better here to just leave it disabled instead of hiding the option entirely.

TheCatLady: Sure, will change that!

src/components/UserList/index.tsx
```diff
@@ -87,6 +87,9 @@ const UserList: React.FC = () => {
       pageIndex * currentPageSize
     }&sort=${currentSort}`
   );
+  const { data: emailSettings } = useSWR(
+    '/api/v1/settings/notifications/email'
+  );
```
sct, line 92: Users need the manage settings permission to hit this endpoint but they only need `manage users` to create a user.

TheCatLady: Ah. How can we verify that email notifications are enabled, then? 🤔  This would also be an issue for my other PR.

TheCatLady: I'm exposing the "email enabled" state via the user notification settings endpoint in #1172, so marking this PR as a draft for now.

sct (approved):

PR #1392, opened 2021-04-09
fix(user.ts): Time value of a quota was being ignored

### LINKED ISSUE #1358: Request limit not resetting
#### Description

I have request limits set to 3 movies per 1 day. However, I requested 2 movies 2 days ago as one of my users, and the request limit has not reset back to having 3 available.

#### Version

1.22.0

#### Steps to Reproduce
1. set request limit to x movies per 1 day
2. request movie
3. wait at least 1 day (or 2 days in my case)

#### Expected Behavior

 request limit should reset back to X available movies but it doesn't 

#### Device

What device were you using when you encountered this issue? Please provide this information to help us reproduce and investigate the bug.

- **Device:** desktop
- **OS:** arch linux
- **Browser:** firefox

**Additional information**

using global request limit (no override)

jonocairns:
re #1358

#### Description
This would cause inaccurate quota rules applied to new requests, as it may or may not artificially
extend the actual cooldown period.

Also added .gitattributes for the pesky windows users that have different line endings than the repo

- Request limit hit at 3PM.
- Day one of the cooldown is the next day (3PM -> 3PM then cut off time time value so we're just at the next day)
- Day two of the cooldown period is the next after that (3PM -> 3PM again then cut of the time again)
- Day 3 at 12:01AM they are able to request again.

This means that 48 hours + 9 hours (3PM -> 11:59PM on the first day) have passed since requesting = 57 hours.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1358

### DIFF
--- a/server/entity/User.ts
+++ b/server/entity/User.ts
@@ -236,11 +236,9 @@ export class User {
     const movieDate = new Date();
     if (movieQuotaDays) {
       movieDate.setDate(movieDate.getDate() - movieQuotaDays);
-    } else {
-      movieDate.setDate(0);
     }
-    // YYYY-MM-DD format
-    const movieQuotaStartDate = movieDate.toJSON().split('T')[0];
+    const movieQuotaStartDate = movieDate.toJSON();
+
     const movieQuotaUsed = movieQuotaLimit
       ? await requestRepository.count({
           where: {
@@ -261,11 +259,8 @@ export class User {
     const tvDate = new Date();
     if (tvQuotaDays) {
       tvDate.setDate(tvDate.getDate() - tvQuotaDays);
-    } else {
-      tvDate.setDate(0);
     }
-    // YYYY-MM-DD format
-    const tvQuotaStartDate = tvDate.toJSON().split('T')[0];
+    const tvQuotaStartDate = tvDate.toJSON();
     const tvQuotaUsed = tvQuotaLimit
       ? (
           await requestRepository
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf


server/entity/User.ts
```diff
@@ -236,11 +236,9 @@ export class User {
     const movieDate = new Date();
     if (movieQuotaDays) {
       movieDate.setDate(movieDate.getDate() - movieQuotaDays);
-    } else {
-      movieDate.setDate(0);
```
TheCatLady, line 240: Why was this removed?

server/entity/User.ts
```diff
@@ -261,11 +259,8 @@ export class User {
     const tvDate = new Date();
     if (tvQuotaDays) {
       tvDate.setDate(tvDate.getDate() - tvQuotaDays);
-    } else {
-      tvDate.setDate(0);
```
TheCatLady, line 265: Same with this...

server/entity/User.ts
sct: I'm not sure it matters since you can't set the interval to 0 anyways?

TheCatLady: It's not the interval though? It's to return the all-time request count if there is no quota.

Though, agreed, it doesn't really matter since we don't actually use that info in any way.

sct: Yeah, but setting the date to 0 does nothing? It will return the same value as the default we set with `new Date()`?

TheCatLady: Oh true, you're right 🤔

sct (approved):
Looks good. I am not sure why we were stripping the time data in the first place but this explains some issues people were having.

Thanks for the fix!

sct:
@all-contributors please add @jonocairns for code

Issue #517, opened 2020-12-27
Blank/Empty Request Card


JohnnyGrey86:
**Describe the bug**
A blank, pulsing request card appears on the home screen, as well as the requests page. After some [Discord discussion](https://discord.com/channels/783137440809746482/789002037529804833/792889878340829214), we discovered this is from a deleted TMDB page. The movie in question now appears under a different number, resulting in this card endlessly attempting to "load."

**What version of Overseerr are you running?**
develop-d00e470b55327489b49d770144b7cfdb24045be6

You can find it under: Settings -> About -> Version

**To Reproduce**
Not applicable.

**Expected behavior**
As sct mentioned in Discord, a proper error message should show, along with the presence of a delete button.

**Screenshots**
![image](https://user-images.githubusercontent.com/14120301/103181408-4ebff880-4866-11eb-99ad-bed9b3935514.png)
![image](https://user-images.githubusercontent.com/14120301/103181416-5d0e1480-4866-11eb-9cd4-e6e9c586145b.png)

**Desktop (please complete the following information):**

- OS: Win10 Pro
- Browser Chrome
- Version 87.0.4280.88 (Official Build) (64-bit)

DirtyCajunRice:
For those that are experiencing the issue currently, if you would like to remove the entry, and that is the only item that is in requested status:
```
docker exec -it <name of container> sh
apk add sqlite
DB="/app/config/db/db.sqlite3"
BAD=$(sqlite3 ${DB} "SELECT id FROM media_request WHERE status=1 LIMIT 1; ")
sqlite3 ${DB} "DELETE FROM season_request WHERE requestId=${BAD};"
sqlite3 ${DB} "DELETE FROM media_request WHERE id=${BAD};"
```

sct:
We good stalebot.

Issue #377, opened 2020-12-17
Deleted titles are stuck on "Available"


ecelebi29:
**Is your feature request related to a problem? Please describe.**

Removing media from Plex does not update the "Available" status on scanned media, even when performing a full manual library scan.

**Describe the solution you'd like**

Full Library Sync should detect missing media and unmark it in Overseerr.

**Describe alternatives you've considered**

Using the Clear All Media Data option on the deleted media and performing a manual library scan does correctly remove it and keep it that way. However, this also clears the request history and therefore isn't a good alternative.

hirenshah:
@sct Should this not be a bug rather than enhancement?

sct:
> @sct Should this not be a bug rather than enhancement?

Not really. We don't have anything built to detect if titles are deleted so this would be a new feature.

douglasparker:
From what I understand, this isn't a stale issue and shouldn't be closed.

flammable:
I'd really appreciate this feature, too! I use Overseerr exclusively for stuff that other people request, and I have a script running on my NAS that deletes those files once their modification dates are past 30 days (it keeps the server clean and allows me to let requests auto-approve). Happy to share the script if it's useful to others.

It'd be ideal if users could go back and re-request things in Overseerr if they're automatically deleted. It's not frequent, but I have a couple of people that request things and forget to watch them before the 30 days have passed, necessitating another request. 😅

Thanks for Overseerr, btw. I had tried to use Ombi before and it just barely worked. Overseerr is clean, easy to use, and works great. It's allowed me to stop fielding requests from my Plex users, which has been a huge lift off of my shoulders. Thank you!

726a67:
My use case is almost identical to the one described by @flammable -- really hoping this issue remains open until it can be implemented.

kevindd992002:
I really hope this gets implemented soon as I think it's a major feature that's needed.

kevindd992002:
Is this new feature implemented already?

douglasparker:
> Is this new feature implemented already?

Not that I am aware of.

I absolutely despise stale-bot as it is currently being used. It's obnoxious.

kevindd992002:
> > Is this new feature implemented already?
> 
> Not that I am aware of.
> 
> I absolutely despise stale-bot as it is currently being used. It's obnoxious.

That's what I thought. I've been waiting for this festure for like forever. I delete media after I watch them which makes the tagging in overseerr inaccurate.

[...]

PR #1397, opened 2021-04-11
feat: Remove media items that no longer exist in Plex

jonocairns:
#### Description

On a full scan, if there are orphaned media items we should remove them any likely anything else relating to that item to ensure we're in sync with the source (plex). 

This is related to #377 but I'm likely missing quite a bit of context for the fix so would appreciate any direction as to how to do this properly. I'm not a huge fan of using the class with a property bag for processed items.

#### To-Dos

### DIFF
--- a/server/lib/scanners/plex/index.ts
+++ b/server/lib/scanners/plex/index.ts
@@ -3,6 +3,7 @@ import { getRepository } from 'typeorm';
 import animeList from '../../../api/animelist';
 import PlexAPI, { PlexLibraryItem, PlexMetadata } from '../../../api/plexapi';
 import { TmdbTvDetails } from '../../../api/themoviedb/interfaces';
+import Media from '../../../entity/Media';
 import { User } from '../../../entity/User';
 import { getSettings, Library } from '../../settings';
 import BaseScanner, {
@@ -35,10 +36,12 @@ class PlexScanner
   private libraries: Library[];
   private currentLibrary: Library;
   private isRecentOnly = false;
+  private processedMediaIds: number[];
 
   public constructor(isRecentOnly = false) {
     super('Plex Scan');
     this.isRecentOnly = isRecentOnly;
+    this.processedMediaIds = [];
   }
 
   public status(): SyncStatus {
@@ -54,6 +57,7 @@ class PlexScanner
   public async run(): Promise<void> {
     const settings = getSettings();
     const sessionId = this.startRun();
+    this.processedMediaIds = [];
     try {
       const userRepository = getRepository(User);
       const admin = await userRepository.findOne({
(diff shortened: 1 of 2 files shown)


samwiseg0:
I think this should be optional since this would effect quotas if something is removed that was requested? 

What if Plex thinks the media is offline/missing? Will this wipe everything in overseerr?

Also what happens if the TMDB ID changes?

jonocairns:
> I think this should be optional since this would effect quotas if something is removed that was requested?
Yeah I think you're right here - could potentially be a switch next to the full scan?

> What if Plex thinks the media is offline/missing? Will this wipe everything in overseerr?
Good point, I'd assume that `plexClient.getLibraryContents(library.id)` would throw an exception and kill any further execution but I haven't tried that locally yet.

> Also what happens if the TMDB ID changes in Plex?
I picked this ID as it was the only non nullable identifier (and I'm fairly new to the codebase!). I'd think that if the TMDB ID changes then this would potentially clean up duplicates, as another media item would be created for the same item. You raise a good point though, I wasn't entirely sure on using this as an ID - it'd be nice to have a sourceId of some sort on the media item, as then we could use the  `serviceId` + `sourceId` to properly sync between the systems.

sct:
> > What if Plex thinks the media is offline/missing? Will this wipe everything in overseerr?
> 
> Good point, I'd assume that `plexClient.getLibraryContents(library.id)` would throw an exception and kill any further execution but I haven't tried that locally yet.

No exceptions will be thrown if the media itself is "offline" (aka file is just missing). Volumes can go offline or be migrated and we probably shouldn't change any availability in that case.

server/lib/scanners/plex/index.ts
```diff
@@ -105,12 +109,33 @@ class PlexScanner
...
+        );
+
+        this.log(`Found ${deletedItems.length} items to delete`, 'info');
+
+        for (const itemToDelete of deletedItems) {
+          const item = existingMedia.find((em) => em.tmdbId === itemToDelete);
+          if (item) {
+            this.log(`Deleting ${item.tmdbId}`, 'info');
+            // is there anything else we should be deleting here? requests?
+            await mediaRepository.remove(item);
```
sct, line 136: Associated requests would be deleted here anyways through the cascade delete.

sct:
I was thinking it makes more sense to instead create a new job that looks through the Overseerr's media items and checks for their existence in Plex (or Sonarr/Radarr) instead of trying to handle all of this in the Plex scanner.

jonocairns:
> I was thinking it makes more sense to instead create a new job that looks through the Overseerr's media items and checks for their existence in Plex (or Sonarr/Radarr) instead of trying to handle all of this in the Plex scanner.

Ah yup - much better idea to separate it out like that. Will close this for now

jonocairns:
sidenote: I think having a separate job for this would potentially solve https://github.com/sct/overseerr/issues/517 as well @sct

TheCatLady:
> sidenote: I think having a separate job for this would potentially solve #517 as well @sct

Still necessary to fix #517... since even if we add a job to "clean up" media items, broken cards would still show until the job runs.  And if we're scanning all media, the default job interval would probably be similar to the full Plex/Radarr/Sonarr scans (24 hours).

PR #1401, opened 2021-04-11
fix(ui): allow canceling from request list & hide edit button for own requests

TheCatLady:
#### Description

- Hide `Edit` button in movie request modal, since users cannot edit their own requests (`MANAGE_REQUESTS` permission is required to edit requests, and users with that permission also have auto-approve perms)
- Show `Cancel Requests` button on request list to allow users to cancel their own requests
- Fix `Recent Requests` slider empty string to reflect that it no longer only displays pending/processing requests

#### To-Dos

#### Issues Fixed or Closed

N/A


sct:
image

🤔

sct:
This PR also removes the ability for admins to cancel a request from the popup edit interface for movies.

TheCatLady:
> image
> 
> 🤔
> 
> (This happens to your own requests that are not approved)

Huh, how are you able to submit requests that aren't approved but still see those buttons?  You shouldn't be allowed to Approve/Deny requests if you have `MANAGE_REQUESTS`, and if you have that perm, your requests should have been auto-approved?

TheCatLady:
> This PR also removes the ability for admins to cancel a request from the popup edit modal for movies.

We don't allow admins to cancel others' pending requests on the request list; only approve/deny actions are shown. Wouldn't it be better to be consistent here?  Again, I'm not sure how you're submitting requests as an admin that aren't immediately approved (and you can't edit already-approved requests).

TheCatLady:
Added a check to verify that the user does not have the `MANAGE_REQUESTS` perm, in case the user was granted `MANAGE_REQUESTS` after submitting a request that has not yet been approved.

sct:
> > This PR also removes the ability for admins to cancel a request from the popup edit modal for movies.
> 
> We don't allow admins to cancel others' pending requests on the request list; only approve/deny actions are shown. Wouldn't it be better to be consistent here? I imagine some admins who find themselves in the `View Request` modal might click the `Cancel Request` button when they really intend to decline a request.
> 
> Also, there is no such `Cancel Request` button for series requests since we show a `Request More` button instead of the `View Request` button.
> 
> And again, I'm not sure how you're submitting requests as an admin that aren't immediately approved (and you can't edit already-approved requests).

There actually should be a cancel request button for series it's just not correctly implemented at the moment. Users who make a series request should be able to edit the request by unchecking the seasons they requested. If all seasons are unchecked (just like the admin edit interface) the button would become "Cancel Request." I suppose in this case, we should if anything allow admins to approve/decline requests from the view request modal. (Not in this PR)

sct (approved):

PR #1404, opened 2021-04-12
feat(inputs): add support for toggling security on input fields

### LINKED ISSUE #1228: Mask sensitive fields in the UI
#### Description

Mask sensitive data such as API keys in the UI. 

#### Desired Behavior

Mask sensitive data such as API keys in the UI. This would help with support as users are sharing screenshots and have to mask this data for help. 

#### Additional Context

N/A

ankarhem:
#### Description
Adds a new `SensitiveInput` component that has a button to toggle visibility.

#### To-Dos

#### Issues Fixed or Closed

- Closes #1228


src/assets/eye-off.svg
```diff
@@ -0,0 +1 @@
+<svg class="w-6 h-6" fill="none" stroke="currentColor" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M13.875 18.825A10.05 10.05 0 0112 19c-4.478 0-8.268-2.943-9.543-7a9.97 9.97 0 011.563-3.029m5.858.908a3 3 0 114.243 4.243M9.878 9.878l4.242 4.242M9.88 9.88l-3.29-3.29m7.532 7.532l3.29 3.29M3 3l3.59 3.59m0 0A9.953 9.953 0 0112 5c4.478 0 8.268 2.943 9.543 7a10.025 10.025 0 01-4.132 5.411m0 0L21 21"></path></svg>
```
sct, line 1: Better to leave the class attribute out of the SVG.

src/assets/eye.svg
```diff
@@ -0,0 +1 @@
+<svg class="w-6 h-6" fill="none" stroke="currentColor" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M15 12a3 3 0 11-6 0 3 3 0 016 0z"></path><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M2.458 12C3.732 7.943 7.523 5 12 5c4.478 0 8.268 2.943 9.542 7-1.274 4.057-5.064 7-9.542 7-4.477 0-8.268-2.943-9.542-7z"></path></svg>
```
sct, line 1: Same here.

src/components/Common/SensitiveInput/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  isLastButton?: boolean;
+}
+
+type SensitiveInputProps = CustomInputProps | CustomFieldProps;
+
+const SensitiveInput = ({
+  isLastButton = true,
+  as = 'input',
+  ...props
+}: SensitiveInputProps): JSX.Element => {
```
sct, line 22: It's better to type this as a `React.FC` to get the correct typing for children and possible return types. (Even though we know the return type here, it's best to keep this component consistent with how we do all our other components)

sct, line 22:
```suggestion
const SensitiveInput: React.FC<SensitiveInputProps> = ({
  isLastButton = true,
  as = 'input',
  ...props
}) => {
```

TheCatLady:
Do we really want to mask the notification agent settings? 🤔   If we do, should we also mask the notification-related fields in User Settings?

src/components/Common/SensitiveInput/index.tsx
ankarhem: As we don't allow passing children here it would make more sense to not type it as if children were allowed?

sct: Children are always allowed, even if they don't do anything.

danshilm:
Can we also mask:
- `SMTP Password` in Notifications -> Email
- `New Password`/`Confirm Password` in Users -> \<user\> -> User settings -> Password

src/components/Common/SensitiveInput/index.tsx
```diff
@@ -0,0 +1,57 @@
+import { Field } from 'formik';
+import React, { useState } from 'react';
+import EyeOffIcon from '../../../assets/eye-off.svg';
+import EyeIcon from '../../../assets/eye.svg';
```
TheCatLady, line 4:
```suggestion
import { EyeIcon, EyeOffIcon } from '@heroicons/react/solid';
```
No need to copy in the SVGs anymore with 8fc71bea08ac5ce3b691ec1de3df1724d91839af 😸 And IMO we should use the solid variant of the icons here, to be consistent with other form buttons (copy and refresh).

src/assets/eye-off.svg
```diff
@@ -0,0 +1 @@
+<svg fill="none" stroke="currentColor" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M13.875 18.825A10.05 10.05 0 0112 19c-4.478 0-8.268-2.943-9.543-7a9.97 9.97 0 011.563-3.029m5.858.908a3 3 0 114.243 4.243M9.878 9.878l4.242 4.242M9.88 9.88l-3.29-3.29m7.532 7.532l3.29 3.29M3 3l3.59 3.59m0 0A9.953 9.953 0 0112 5c4.478 0 8.268 2.943 9.543 7a10.025 10.025 0 01-4.132 5.411m0 0L21 21"></path></svg>
```
TheCatLady, line 1:
```suggestion
```
No longer needed with `@heroicons/react`.

src/assets/eye.svg
```diff
@@ -0,0 +1 @@
+<svg fill="none" stroke="currentColor" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M15 12a3 3 0 11-6 0 3 3 0 016 0z"></path><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M2.458 12C3.732 7.943 7.523 5 12 5c4.478 0 8.268 2.943 9.542 7-1.274 4.057-5.064 7-9.542 7-4.477 0-8.268-2.943-9.542-7z"></path></svg>
```
TheCatLady, line 1:
```suggestion
```
No longer needed with `@heroicons/react`.

TheCatLady:
@sct fixed the `.gitattributes` file in 6a4ea1b0c5637f7b4a308941b177463fd2140c4a, so you should be able to revert the `jpg` changes 🙂

src/components/Common/SensitiveInput/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+        }
+        style={{ borderTopRightRadius: '0px', borderBottomRightRadius: '0px' }}
+      />
+      <button
+        onClick={(e) => {
+          e.preventDefault();
+          setHidden(!isHidden);
+        }}
+        className={`relative inline-flex items-center px-4 py-2 -ml-px text-sm font-medium leading-5 text-white transition duration-150 ease-in-out bg-indigo-600 border border-gray-500 hover:bg-indigo-500 focus:outline-none focus:ring-blue focus:border-blue-300 active:bg-gray-100 active:text-gray-700 ${
+          isLastButton ? 'rounded-r-md' : ''
```
TheCatLady, line 50: These styles are now being set in `globals.css` so this is no longer necessary.  (You can also remove the `isLastButton` property!)

src/components/Common/SensitiveInput/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+        }`}
+      >
+        {isHidden ? (
+          <EyeOffIcon className="w-5 h-5" />
+        ) : (
+          <EyeIcon className="w-5 h-5" />
```
TheCatLady, line 56:
```suggestion
          <EyeOffIcon />
        ) : (
          <EyeIcon />
```

src/components/Settings/Notifications/NotificationsPushover/index.tsx
```diff
@@ -189,7 +190,8 @@ const NotificationsPushover: React.FC = () => {
                 </label>
                 <div className="form-input">
                   <div className="form-input-field">
-                    <Field
+                    <SensitiveInput
+                      as="field"
```
TheCatLady, line 194:
```suggestion
                    <Field
```
As per our previous discussion, this field shouldn't be masked.

ankarhem: The API token should not be protected?

TheCatLady: @ankarhem It's tied to the user account, it's not really an API token in the same sense as the others.  It is useless without the user token.  With Pushover, it is really easy to mix up the application and user tokens since they are quite similar (both 30-char alphanum strings), so I'm against masking them.

TheCatLady:
Would be nice to also use the `SensitiveInput` component for the password fields in `src/components/Login/LocalLogin.tsx` and `src/components/ResetPassword/index.tsx`! 😃

src/components/Common/SensitiveInput/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <Component
+        {...componentProps}
+        type={
+          isHidden
+            ? 'password'
+            : props.type !== 'password'
+            ? props.type
+            : 'text'
+        }
+        style={{ borderTopRightRadius: '0px', borderBottomRightRadius: '0px' }}
```
TheCatLady, line 39: Can alternatively apply the `rounded-l-only` class here.

PR #1424, opened 2021-04-14
fix(ui): remove 'Upcoming Series' slider

TheCatLady:
#### Description

The results returned by TMDb for `Upcoming Series` slider honestly leave much to be desired, and the slider often surfaces new entries that people have added to TMDb but not necessarily tagged correctly.

The slider problematically frequently returns adult content high in the list of results, even with region and/or language filtering in place.  (TMDb does not support filtering out adult results for TV series the way it does for movies, but that is a whole other issue.)

This PR simply removes the slider from the `Discover` page. The API endpoint and `Upcoming Series` page (`/discover/tv/upcoming`) are left intact, since we will hopefully be able to reinstate this slider in the future once we have added support for filtering by age ratings.  Users will still be able to navigate to and view this page, but there will no longer be any links to it from the UI.

#### To-Dos

- [x] Translation keys `yarn i18n:extract` (**Note:** The removed string is also shared/used by another component, so there are no changes to `en.json`.)

#### Issues Fixed or Closed

N/A

### DIFF
--- a/src/components/Discover/index.tsx
+++ b/src/components/Discover/index.tsx
@@ -19,7 +19,6 @@ const messages = defineMessages({
   recentrequests: 'Recent Requests',
   popularmovies: 'Popular Movies',
   populartv: 'Popular Series',
-  upcomingtv: 'Upcoming Series',
   recentlyAdded: 'Recently Added',
   noRequests: 'No requests.',
   upcoming: 'Upcoming Movies',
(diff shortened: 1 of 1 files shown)


sct:
Frequently returns adult content? Is that true? I've never once seen adult content in the results besides the one today.

I actually find the results pretty useful. There are several shows in the results right now I wouldn't have know existed without that slider.

That's not to say there it's always great content in the list, but the first page or so of content is valuable.

TheCatLady:
> Frequently returns adult content? Is that true? I've never once seen adult content in the results besides the one today.

I've seen adult content in the slider itself once (without clicking through to the full page of results). I ended up enabling the `Discover Language` filtering a few days ago in an attempt to reduce the chances/frequency of this occurring.

I just think that the fact that it does surface adult results regularly (and today wasn't the first time we got a report of this happening) means that we need to do something to fix it, since it is NOT okay to display those results when there are potentially younger users being granted access to Overseerr.

PR #1427, opened 2021-04-14
fix(ui): dim password field when password generation option is selected

TheCatLady:
#### Description

Dims password field in `Create Local User` modal when `Automatically Generate Password` is checked, so that it is clearer that the password field is disabled.

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/src/components/UserList/index.tsx
+++ b/src/components/UserList/index.tsx
@@ -388,7 +388,9 @@ const UserList: React.FC = () => {
                       />
                     </div>
                   </div>
-                  <div className="form-row">
+                  <div
+                    className={`form-row ${values.genpassword && 'opacity-50'}`}
+                  >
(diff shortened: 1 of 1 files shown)


src/components/UserList/index.tsx
```diff
@@ -388,7 +388,9 @@ const UserList: React.FC = () => {
                       />
                     </div>
                   </div>
-                  <div className="form-row">
+                  <div
+                    className={`form-row ${values.genpassword && 'opacity-50'}`}
```
sct, line 392: this will end up printing out the class name `false` when `values.genpassword` is false. You shouldn't use the `&&` pattern like this in anything except for rendering React components.

TheCatLady: Ah okay, fixed!

sct (approved):

PR #1440, opened 2021-04-16
feat(jobs): wip allow modifying job schedules

### LINKED ISSUE #803: Allow adjustments to background job schedule
Not related to a problem - feature request

On the jobs page in settings, it would be great to have the ability to change frequency/schedule of automated jobs. By default some of them run quite frequently which is unnecessary for my low-traffic build. It also increases CPU usage for low power hardware like raspberry pis. 

Could not find any config files to edit this. 

**Additional context**
Capture

danshilm:
#### Description

This implements the ability to modify the schedule of a job using a cron string.

#### To-Dos

- [x] Implement backend
- [ ] Add modal/UI to modify job schedule

#### Issues Fixed or Closed

- Fixes #803


TheCatLady:
I feel like while it would allow for the greatest degree of customizability, a `cron` string wouldn't be the most user-friendly.

What do you think about instead simply allowing the user to customize the frequency of each of the jobs?  E.g., allow users to choose between several presets for `every X hours`?

danshilm:
> 
> 
> I feel like while it would allow for the greatest degree of customizability, a `cron` string wouldn't be the most user-friendly.
> 
> What do you think about instead simply allowing the user to customize the frequency of each of the jobs? E.g., allow users to choose between several presets for `every X hours`? (We can still save the setting as a `cron` string, but just generate the string for the user.)

Yeah that sounds better actually. I'll add that.

There's a good amount of customisation that can be applied to the job schedule; its possible to also set the `StartTime` and `EndTime`. Should that be implemented as well?

sct:
> > I feel like while it would allow for the greatest degree of customizability, a `cron` string wouldn't be the most user-friendly.
> > What do you think about instead simply allowing the user to customize the frequency of each of the jobs? E.g., allow users to choose between several presets for `every X hours`? (We can still save the setting as a `cron` string, but just generate the string for the user.)
> 
> Yeah that sounds better actually. I'll add that.
> 
> There's a good amount of customisation that can be applied to the job schedule; it's possible to also set the `StartTime` and `EndTime`. Should that be implemented as well?

Let's keep it simple for now.

server/lib/settings.ts
```diff
@@ -191,6 +196,7 @@ interface AllSettings {
   sonarr: SonarrSettings[];
   public: PublicSettings;
   notifications: NotificationSettings;
+  jobs: JobSettings[];
```
sct, line 199: What about doing something like `Record<JobName, JobSettings>` instead? It will be easier to pull the job data later this way instead of doing find lookups. They can be empty objects by default maybe?

danshilm: Yep, that would be better. I'll make the change.

src/components/Settings/SettingsJobsCache/index.tsx
```diff
@@ -126,6 +152,111 @@ const SettingsJobs: React.FC = () => {
...
+          {({ handleSubmit, isSubmitting, errors, touched }) => {
+            return (
+              <Modal
+                title={intl.formatMessage(messages.editJobSchedule)}
+                okText={
+                  isSubmitting
+                    ? intl.formatMessage(globalMessages.saving)
+                    : intl.formatMessage(globalMessages.save)
+                }
+                iconSvg={<SolidPencilIcon className="w-5 h-5" />}
```
TheCatLady, line 205:
```suggestion
                iconSvg={<SolidPencilIcon className="w-6 h-6" />}
```
(@sct this is why we need #1464 😉)

src/components/Settings/SettingsJobsCache/index.tsx
```diff
@@ -51,6 +62,13 @@ const messages: { [messageName: string]: MessageDescriptor } = defineMessages({
   'sonarr-scan': 'Sonarr Scan',
   'download-sync': 'Download Sync',
   'download-sync-reset': 'Download Sync Reset',
+  editJobSchedule: 'Modify Job Schedule',
+  editJobScheduleDescription:
+    'A job schedule can be represented by a CRON expression. Use a tool like <cronhub>crontab.cronhub.io</cronhub> to generate valid CRON expressions.',
+  jobScheduleEditSaved: 'Job schedule modified successfully!',
+  jobScheduleEditFailed: 'Something went wrong while saving the job schedule',
+  editJobSchedulePrompt: 'Enter CRON expression',
+  validationJobSchedule: 'You must provide a CRON expression',
```
TheCatLady, line 71: I thought we weren't going to allow users to enter the cron expression directly?

danshilm: Yes, this was just what I was already working on. I'll change it to the presets next.

server/lib/settings.ts
```diff
@@ -182,6 +182,12 @@ interface NotificationSettings {
   agents: NotificationAgents;
 }
 
+interface JobSettings {
+  schedule: string;
+}
+
+type JobName = string;
```
rcohen-unext, line 189: I don't really like having the job name as a string. It in some ways defeats the purpose of using a Record. We should have a type of all the possible job ids (which shouldn't be changed from the IDs they were before) and use that as the type on the Record.

```javascript
type JobIds = 'plex-recently-added' | 'some-other-job';

Record<JobIds, JobSettings>
```

Then later, typescript will know which jobs we can reference and we avoid possible undefined objects.

server/job/schedule.ts
```diff
@@ -17,17 +18,22 @@ interface ScheduledJob {
...
     name: 'Plex Recently Added Scan',
     type: 'process',
-    job: schedule.scheduleJob('0 */5 * * * *', () => {
-      logger.info('Starting scheduled job: Plex Recently Added Scan', {
-        label: 'Jobs',
-      });
-      plexRecentScanner.run();
-    }),
+    job: schedule.scheduleJob(
+      jobs['Plex Recently Added Scan']?.schedule ?? '0 */5 * * * *',
```
rcohen-unext, line 29: Here is an example where having a record with strict keys makes you not have to check if this is undefined.

We can set the defaults job values in settings as well and avoid using null coalescing here if schedule is not set?

server/lib/settings.ts
sct, line 189: I don't really like having the job name as a string. It in some ways defeats the purpose of using a Record. We should have a type of all the possible job ids (which shouldn't be changed from the IDs they were before) and use that as the type on the Record.

```javascript
type JobIds = 'plex-recently-added' | 'some-other-job';

Record<JobIds, JobSettings>
```

Then later, typescript will know which jobs we can reference and we avoid possible undefined objects.

server/job/schedule.ts
sct, line 29: Here is an example where having a record with strict keys makes you not have to check if this is undefined.

We can set the defaults job values in settings as well and avoid using null coalescing here if schedule is not set?

src/components/Settings/SettingsJobsCache/index.tsx
```diff
@@ -51,6 +59,13 @@ const messages: { [messageName: string]: MessageDescriptor } = defineMessages({
   'sonarr-scan': 'Sonarr Scan',
   'download-sync': 'Download Sync',
   'download-sync-reset': 'Download Sync Reset',
+  editJobSchedule: 'Modify Job Schedule',
+  jobScheduleEditSaved: 'Job schedule modified',
+  jobScheduleEditFailed:
+    'Something went wrong while modifying the job schedule',
+  editJobSchedulePrompt: 'Select job schedule',
+  editJobScheduleSelector:
+    'Every {jobScheduleMinutes} minute(s) or {jobScheduleHours} hour(s)',
```
TheCatLady, line 68: This is confusing and not very user-friendly.


How about something like:
```
Every [ <number>  v] [ minutes  v]
                     [ hour(s)   ]
```

(We should not allow users to run anything more often than once every 5 minutes, by the way.)

TheCatLady: In fact, we should probably enforce some "sane" limits here so that people don't try to run full Plex scans very few minutes.

Perhaps we should have some jobs for which users can choose a "every X hours" job frequency, and others for which they can select a "every X minutes" job frequency.

server/job/schedule.ts
danshilm: Alright, gotcha. Will make the changes.

src/components/Settings/SettingsJobsCache/index.tsx
```diff
@@ -51,6 +59,13 @@ const messages: { [messageName: string]: MessageDescriptor } = defineMessages({
   'sonarr-scan': 'Sonarr Scan',
   'download-sync': 'Download Sync',
   'download-sync-reset': 'Download Sync Reset',
+  editJobSchedule: 'Modify Job Schedule',
+  jobScheduleEditSaved: 'Job schedule modified',
+  jobScheduleEditFailed:
+    'Something went wrong while modifying the job schedule',
+  editJobSchedulePrompt: 'Select job schedule',
```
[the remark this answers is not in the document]
danshilm: Thanks!

src/components/Settings/SettingsJobsCache/index.tsx
```diff
@@ -1,6 +1,12 @@
-import { PlayIcon, StopIcon, TrashIcon } from '@heroicons/react/outline';
+import {
+  PencilIcon,
+  PlayIcon,
+  StopIcon,
+  TrashIcon,
+} from '@heroicons/react/outline';
+import { PencilIcon as SolidPencilIcon } from '@heroicons/react/solid';
```
TheCatLady, line 7:
```suggestion
import {
  PlayIcon,
  StopIcon,
  TrashIcon,
} from '@heroicons/react/outline';
import { PencilIcon } from '@heroicons/react/solid';
```
We should be consistent in using the solid pencil icon IMO, like in other places in the UI.

src/components/Settings/SettingsJobsCache/index.tsx
```diff
@@ -126,6 +191,77 @@ const SettingsJobs: React.FC = () => {
...
+        show={jobEditModal.isOpen}
+      >
+        <Modal
+          title={intl.formatMessage(messages.editJobSchedule)}
+          okText={
+            isSaving
+              ? intl.formatMessage(globalMessages.saving)
+              : intl.formatMessage(globalMessages.save)
+          }
+          iconSvg={<SolidPencilIcon className="w-6 h-6" />}
```
TheCatLady, line 210:
```suggestion
          iconSvg={<PencilIcon />}
```

src/components/Settings/SettingsJobsCache/index.tsx
```diff
@@ -179,6 +315,14 @@ const SettingsJobs: React.FC = () => {
                   </div>
                 </Table.TD>
                 <Table.TD alignText="right">
+                  <Button
+                    className="mr-2"
+                    buttonType="warning"
+                    onClick={() => setJobEditModal({ isOpen: true, job: job })}
+                  >
+                    <PencilIcon className="w-5 h-5 mr-1" />
```
TheCatLady, line 323:
```suggestion
                    <PencilIcon />
```

TheCatLady (commented):
#1464 has now been merged, so you no longer need to specify the `className` attribute for modal/button icons 😸

src/components/Settings/SettingsJobsCache/index.tsx
danshilm: Do we want to restrict the frequency for jobs that can only be ran "every X hours" to something else than 1 to 100? I was thinking 72 would be a reasonable maximum.

src/components/Settings/SettingsJobsCache/index.tsx
```diff
@@ -51,12 +56,15 @@ const messages: { [messageName: string]: MessageDescriptor } = defineMessages({
   'sonarr-scan': 'Sonarr Scan',
   'download-sync': 'Download Sync',
   'download-sync-reset': 'Download Sync Reset',
+  editJobSchedule: 'Modify Job',
+  jobScheduleEditSaved: 'Job edited successfully!',
+  jobScheduleEditFailed: 'Something went wrong while saving the job.',
+  editJobSchedulePrompt: 'Frequency',
+  editJobScheduleSelectorMinutes: 'Every {jobScheduleMinutes} minutes',
+  editJobScheduleSelectorHours: 'Every {jobScheduleHours} hour(s)',
```
TheCatLady, line 64:
```suggestion
  editJobScheduleSelectorMinutes: 'Every {jobScheduleMinutes, plural, one {minute} other {{jobScheduleMinutes} minutes}}',
  editJobScheduleSelectorHours: 'Every {jobScheduleHours, plural, one {hour} other {{jobScheduleHours} hours}}',
```

src/components/Settings/SettingsJobsCache/index.tsx
danshilm: Yeah that sounds reasonable for "long" jobs.

About `Download Sync`, I'm not really sure what would be right way to deal with this. Making it uneditable is a good enough solution imo since that job barely requires any computing power unlike a `Plex Recently Added Scan` for example.

TheCatLady: Yeah that's also my thought.  There really shouldn't be any need to modify that one.  So I guess we'll have three job types, "fixed" or "static" (unsure the best name for this), "short," and "long"? 🤔

src/components/Settings/SettingsJobsCache/index.tsx
```diff
@@ -118,6 +138,41 @@ const SettingsJobs: React.FC = () => {
...
+    const jobScheduleCron = ['0', '0', '*', '*', '*', '*'];
+
+    try {
+      if (jobEditModal.job?.intervalDuration === 'short') {
+        jobScheduleCron[1] = `*/${jobScheduleMinutes}`;
+      }
+
+      if (jobEditModal.job?.intervalDuration === 'long') {
+        jobScheduleCron[2] = `*/${jobScheduleHours}`;
+      }
```
TheCatLady, line 151:
```suggestion
      if (jobEditModal.job?.intervalDuration === 'short') {
        jobScheduleCron[1] = `*/${jobScheduleMinutes}`;
      } else if (jobEditModal.job?.intervalDuration === 'long') {
        jobScheduleCron[2] = `*/${jobScheduleHours}`;
      } else {
        return;
      }
```

danshilm: Gotcha

TheCatLady:
Don't forget to run `yarn i18n:extract` again :)

overseerr-api.yml
```diff
@@ -2249,6 +2249,54 @@ paths:
...
+                    enum: [short, long, fixed]
+                  name:
+                    type: string
+                    example: A Job Name
+                  nextExecutionTime:
+                    type: string
+                    example: '2020-09-02T05:02:23.000Z'
+                  running:
+                    type: boolean
+                    example: false
```
TheCatLady, line 2299: Probably best to add a new schema at the top, and reference it in all of the job-related endpoints.

danshilm: Gotcha

PR #1464, opened 2021-04-18
refactor(css): add styles targeting SVG button/modal icons

TheCatLady:
#### Description

~~Adds and uses global CSS classes `sm-button-icon`, `button-icon`, `button-with-text-icon`, `modal-icon`, and `slider-link-icon`.~~

Applies consistent styles to button and modal icons.

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/src/components/Common/ButtonWithDropdown/index.tsx
+++ b/src/components/Common/ButtonWithDropdown/index.tsx
@@ -19,16 +19,16 @@ const DropdownItem: React.FC<DropdownItemProps> = ({
   buttonType = 'primary',
   ...props
 }) => {
-  let styleClass = '';
+  let styleClass = 'button-md text-white';
 
   switch (buttonType) {
     case 'ghost':
-      styleClass =
-        'text-white bg-gray-700 hover:bg-gray-600 hover:text-white focus:border-gray-500 focus:text-white';
+      styleClass +=
+        ' bg-gray-700 hover:bg-gray-600 focus:border-gray-500 focus:text-white';
       break;
     default:
-      styleClass =
-        'text-white bg-indigo-600 hover:bg-indigo-500 hover:text-white focus:border-indigo-700 focus:text-white';
+      styleClass +=
+        ' bg-indigo-600 hover:bg-indigo-500 focus:border-indigo-700 focus:text-white';
   }
   return (
     <a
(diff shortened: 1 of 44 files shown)


sct:
I think this is more confusing (and actually more to type) than just using the classes directly. Global classes let us change styling site-wide by editing one place, but should we really be doing that in most cases for these SVGs? The context of the SVG's size directly belongs to where it's being used.

TheCatLady:
> I think this is more confusing (and actually more to type) than just using the classes directly. Global classes let us change styling site-wide by editing one place, but should we really be doing that in most cases for these SVGs? The context of the SVG's size directly belongs to where it's being used.

I was thinking that it'd be easier to enforce consistent sizing for certain places where we have icons, which is why these classes don't apply to all SVGs. For example, button icons should all be sized consistently, and the modal icons should always be the same size as well.

sct:
It still means typing out a class for every button icon. A better way of doing this would be creating new classes for buttons that directly target child SVGs and apply the necessary styles.

TheCatLady:
> It still means typing out a class for every button icon. A better way of doing this would be creating new classes for buttons that directly target child SVGs and apply the necessary styles.

Ah, that's very true!  Definitely a better solution.

src/components/Common/ButtonWithDropdown/index.tsx
```diff
@@ -60,24 +60,24 @@ const ButtonWithDropdown: React.FC<ButtonWithDropdownProps> = ({
...
+    dropdownClasses: 'button-md ',
   };
 
   switch (buttonType) {
     case 'ghost':
       styleClasses.mainButtonClasses +=
         ' bg-transparent border-gray-600 hover:border-gray-200 focus:border-gray-100 active:border-gray-100';
       styleClasses.dropdownSideButtonClasses = styleClasses.mainButtonClasses;
-      styleClasses.dropdownClasses = 'bg-gray-700';
+      styleClasses.dropdownClasses += 'bg-gray-700';
```
sct, line 73: why not prepend the space here as we do for the other classes?

src/styles/globals.css
```diff
@@ -281,6 +285,23 @@ select.short {
   @apply w-min;
 }
 
+button svg,
+.button-md svg {
+  @apply w-5 h-5 mr-2 last:mr-0;
+}
+
+.button-sm svg {
+  @apply w-4 h-4 mr-1.5 last:w-5 last:h-5 last:mr-0;
```
sct, line 294: Why is the icon larger only if it's the last item?

TheCatLady: We use 5x5 icons along with the smaller button size when there is no accompanying text (e.g., logs page). It's not so much that it's the last item, but the only item.

sct (approved):

PR #1474, opened 2021-04-20
feat(perms): add separate REQUEST_MOVIE and REQUEST_TV permissions

### LINKED ISSUE #930: Hide ability to request Shows OR Movies
It would be nice if you could set a user's ability to see and request movies or tv show **only**. 

Hide shows  **OR** moveis from the UI.

ombi 

**Additional context**
Sometimes you don't want people to request stuff you don't want to get. It's easier just to hide the possibility than to try to explain why.

TheCatLady:
#### Description

Adds new permissions for `REQUEST_MOVIE` and `REQUEST_TV` to allow admins to only grant users the ability to submit requests of one type.

#### To-Dos

#### Issues Fixed or Closed

- Closes #930


src/components/MovieDetails/index.tsx
```diff
@@ -122,6 +122,9 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
 
   if (
     data.mediaInfo?.plexUrl4k &&
+    hasPermission([Permission.REQUEST, Permission.REQUEST_MOVIE], {
+      type: 'or',
+    }) &&
     hasPermission([Permission.REQUEST_4K, Permission.REQUEST_4K_MOVIE], {
       type: 'or',
     })
```
sct, line 130: Should we require base request permissions to be able to request 4k?

TheCatLady: @sct We actually already do. This is simply fixing the behavior in places where we don't currently check for the base request perms.

TheCatLady: (I would not be opposed to modifying this behavior, however.)

sct: Yeah I mean now that we have the ability to select movies/tv for both types, one use case might be ONLY allowing 4k movies/tv, but no standard requests?

sct (approved):

PR #1495, opened 2021-04-24
feat(notif): add LunaSea agent

TheCatLady:
#### Description

Adds official support for [LunaSea push notifications](https://docs.lunasea.app/lunasea/notifications/overseerr).

Also optimized & resized notification agent logos so that they have a consistent, square aspect ratio. 🐱

image

#### To-Dos

#### Issues Fixed or Closed

N/A


sct:
Maybe we can come up with something more friendly than `Authorization Header` now that its a direct integration? @JagandeepBrar ideas?

JagandeepBrar:
> Maybe we can come up with something more friendly than `Authorization Header` now that its a direct integration? @JagandeepBrar ideas?

Currently I'm using just the username part of a standard basic authentication header to pull the user's profile name from the request. If you could make the field just "Profile Name" or something, and then generate a basic auth header, that would make it super easy for the users :)

Just to clarify so there's no confusion, if my profile name was `LunaSea`, I would base64 encode `LunaSea:` (no password, but still need the colon) which encodes to `THVuYVNlYTo=`. And so the authorization header would be `Basic THVuYVNlYTo=`.

If the header is completely empty it assumes the default LunaSea profile (`default`), so it can remain an optional field.

TheCatLady:
> Currently I'm using just the username part of a standard basic authentication header to pull the user's profile name from the request. If you could make the field just "Profile Name" or something, and then generate a basic auth header, that would make it super easy for the users :)
> 
> Just to clarify so there's no confusion, if my profile name was `LunaSea`, I would base64 encode `LunaSea:` (no password, but still need the colon) which encodes to `THVuYVNlYTo=`. And so the authorization header would be `Basic THVuYVNlYTo=`.
> 
> If the header is completely empty it assumes the default LunaSea profile (`default`), so it can remain an optional field.

Done! From my testing, however, it appears that notifications are delivered regardless of whether or not the decoded profile name actually matches an existing profile though?

JagandeepBrar:
> Done! From my testing, however, it appears that notifications are delivered regardless of whether or not the decoded profile name actually matches an existing profile though?

That's correct, there's no way for me to guarantee that the profile exists (since any data stored in the cloud is encrypted in LunaSea). If a user tries to open a notification linked to a profile that doesn't exist, they'll get an in-app notification regarding it.

sct (approved):

PR #1496, opened 2021-04-24
feat: add option to only allow Plex sign-in from existing users

TheCatLady:
#### Description

Currently, any Plex user with access to the connected Plex server can sign in to Overseerr, even if they have not been imported.

This PR adds a setting option so that this behavior is configurable, as some server admins may not want to allow Overseerr access to their Plex users automatically.

image

#### To-Dos

#### Issues Fixed or Closed

N/A


server/lib/settings.ts
```diff
@@ -356,6 +359,7 @@ class Settings {
       applicationTitle: this.data.main.applicationTitle,
       hideAvailable: this.data.main.hideAvailable,
       localLogin: this.data.main.localLogin,
+      newPlexLogin: this.data.main.newPlexLogin,
```
sct, line 362: Why is this being added to public settings?

TheCatLady: Oh yup, good call, it doesn't need to be.

sct (approved):

PR #1509, opened 2021-04-26
feat(pwa): add shortcuts to PWA

danshilm:
#### Description
Adds shortcuts to the Requests, Users and Settings pages.

Example on Windows:
image

#### To-Dos

#### Issues Fixed or Closed

- Fixes none


sct:
My issue with this is non-admins will also see the users/settings links (And they just won't work.)

Maybe instead of users/settings we can add Profile/Settings (but the users settings)?

TheCatLady:
Pretty sure you can use SVG icons in the manifest :)

danshilm:
> 
> 
> My issue with this is non-admins will also see the users/settings links (And they just won't work.)
> 
> Maybe instead of users/settings we can add Profile/Settings (but the users settings)?

Ah yes, I didn't think about that.
Sounds good, I'll make the change.

danshilm:
> 
> 
> ~Pretty sure you can use SVG icons in the manifest :)~ Scratch that, Chrome doesn't support SVG here apparently. 😿 https://web.dev/app-shortcuts/#icons-(optional)
> 
> That said:
> 
> > If you want pixel-perfect icons, provide them in increments of 48dp (i.e. 36x36, 48x48, 72x72, 96x96, 144x144, 192x192 pixel icons). Otherwise, it is recommended that you use a single 192x192 pixel icon.

192x192 px it is 😄

sct (approved):
LGTM 💯 Thanks!

PR #1510, opened 2021-04-27
fix(css): don't target button globally

TheCatLady:
#### Description

Fixes inadvertent changes from 3e5e9c0ad181926b48f73e33edd1abb0f2885fc5.

Also removes now-unnecessary Tailwind variant config. 🥳

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/src/components/RequestBlock/index.tsx
+++ b/src/components/RequestBlock/index.tsx
@@ -97,7 +97,6 @@ const RequestBlock: React.FC<RequestBlockProps> = ({ request, onUpdate }) => {
               <>
                 <Button
                   buttonType="success"
-                  buttonSize="sm"
                   className="mr-1"
                   onClick={() => updateRequest('approve')}
                   disabled={isUpdating}
@@ -106,7 +105,6 @@ const RequestBlock: React.FC<RequestBlockProps> = ({ request, onUpdate }) => {
                 </Button>
                 <Button
                   buttonType="danger"
-                  buttonSize="sm"
                   className="mr-1"
                   onClick={() => updateRequest('decline')}
                   disabled={isUpdating}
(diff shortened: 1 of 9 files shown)


src/components/Toast/index.tsx
```diff
@@ -48,7 +48,7 @@ const Toast: React.FC<ToastProps> = ({
                 <div className="flex flex-shrink-0 ml-4">
                   <button
                     onClick={() => onDismiss()}
-                    className="inline-flex text-gray-400 transition duration-150 ease-in-out focus:outline-none focus:text-gray-500"
+                    className="inline-flex text-gray-400 transition duration-150 ease-in-out button-md focus:outline-none focus:text-gray-500"
```
sct, line 51: Why are we changing this?

src/components/RequestBlock/index.tsx
```diff
@@ -97,7 +97,6 @@ const RequestBlock: React.FC<RequestBlockProps> = ({ request, onUpdate }) => {
               <>
                 <Button
                   buttonType="success"
-                  buttonSize="sm"
```
sct, line 100: Why are we changing the button size here?

TheCatLady: It actually was `md` before, I think I changed it because the icons for these buttons don't match those of the `md` buttons in the rest of the UI. So this actually reverts the change you mentioned on Discord.

src/components/Toast/index.tsx
TheCatLady: Ah we shouldn't, reverting.

src/components/RequestBlock/index.tsx
sct: Huh, i guess at some point I changed them back to `md`. I think they may have originally been smaller 🤷 Okay!

sct (approved):

src/components/RequestBlock/index.tsx
TheCatLady: `sm` buttons with 4x4 icons are pretty small and probably not super great on mobile, so you might have changed it for that reason? 🤷🏻‍♀️

PR #1512, opened 2021-04-27
feat: allow users to select notification types

TheCatLady:
#### Description

- Adds `NotificationTypeSelector` to user notification settings pages, to allow users to select the notification types they would like to receive via each agent
    - Users can only view and select the notification types they are able to receive, based on their permissions
    - `Enable Notifications`/`Enable Mentions` checkbox is removed since it serves no purpose with the `NotificationTypeSelector`

- Removes `NotificationTypeSelector` from app/system notification settings pages for agents which _only_ support delivering notifications to individual users (email and web push)

- Adds validation to ensure that at least one notification type is selected when an agent is enabled

- Unchecks all notification types when an agent is disabled

- Removes email alert explaining notification recipients, as it is no longer needed (users can see exactly which notification types they are entitled to in their user notification settings)

- Since it is necessary to verify user notification type selections either in addition to or in conjunction with system notification type selections, moved all `hasNotificationType()` logic out of `shouldSend()` and into `send()`
    - Fixes a bug that prevents test notification emails from being sent if a user has their email notifications disabled (`hasNotificationType()` should always return true for test notifications)
    - Telegram agent also now sends management notifications to admin users, matching the behavior of other agents

- Adds additional logging to the web push agent

- User notification settings (user does not have auto-approve perms):
    image

- User email notification settings (user is admin):
    image

- LunaSea notification settings (agent enabled but no notification types selected):
  image

#### To-Dos

#### Issues Fixed or Closed

N/A


overseerr-api.yml
```diff
@@ -1597,8 +1595,9 @@ components:
         pgpKey:
           type: string
           nullable: true
-        discordEnabled:
-          type: boolean
+        discordEnabledTypes:
+          type: number
+          nullable: true
```
sct, line 1600: This is a breaking API change.

overseerr-api.yml
```diff
@@ -1213,8 +1213,6 @@ components:
               type: string
             userToken:
               type: string
-            priority:
-              type: number
```
sct, line 1217: So is this, technically.

TheCatLady: This was never used anywhere actually.  Should we not remove it?

sct (approved):
Well, I went through this and wasn't able to break anything yet. It's a huge PR, and so it's likely something slipped by. Let's get this merged finally and let users find those edge-cases!

PR #1525, opened 2021-04-28
fix(pwa): add Discover shortcut and fix/optimize icons

TheCatLady:
#### Description

Adds `Discover` shortcut, and also fixes the icons to match those used in the Overseerr UI (outline clock, solid user, and solid cog).

Also updates all URLs in the manifest to be relative, so that no changes will be required when we eventually support base URLs.

image

#### To-Dos

#### Issues Fixed or Closed

N/A


danshilm:
The `Discover` shortcut is somewhat redundant here since cold opening the PWA will open the Discover page itself.
It's also something I've never seen app shortcuts do; my Camera app has shortcuts to Portrait mode, Pro mode, Video mode, Selfie mode, and not the usual Photo mode that I would land on when cold opening the app.

sct (approved):

TheCatLady:
> The `Discover` shortcut is somewhat redundant here since cold opening the PWA will open the Discover page itself.
> It's also something I've never seen app shortcuts do; my Camera app has shortcuts to Portrait mode, Pro mode, Video mode, Selfie mode, and not the usual Photo mode that I would land on when cold opening the app.

This is true. However, I think there is value in allowing users to explicitly choose to open the app to "discover" items.

I can definitely remove the `Discover` entry if you all are against having it, but I think it's nice to have (albeit a bit unnecessary).

sct:
I think its fine. It's redundant but it also doesn't hurt to have it there 🤷

PR #1537, opened 2021-04-29
fix: correctly fall back to English name in LanguageSelector

TheCatLady:
#### Description

When multiple languages are selected for the `Discover Language` server setting, the fallback logic does not use the individual language code to determine the TMDb English name, but the entire `|`-delimited string (e.g., `en|zh|ko`).

In addition to fixing that bug, this PR also refactors sorting and name fallback logic for both the `LanguageSelector` and `RegionSelector`, to use `lodash` `sortBy` and remove duplicated code for `intl.formatDisplayName` & TMDb `english_name` fallbacks.

- Before:
  Screenshot

- After:
  Screenshot

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/src/components/LanguageSelector/index.tsx
+++ b/src/components/LanguageSelector/index.tsx
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
+import { sortBy } from 'lodash';
 import dynamic from 'next/dynamic';
 import React, { useMemo } from 'react';
 import { defineMessages, useIntl } from 'react-intl';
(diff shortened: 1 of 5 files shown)


src/components/LanguageSelector/index.tsx
```diff
@@ -29,34 +31,43 @@ const selectStyles = {
...
+
+    return sortBy(languages, 'name');
+  }, [intl, languages]);
+
+  function languageName(langCode: string): string {
+    return (
+      sortedLanguages?.find((language) => language.iso_639_1 === langCode)
+        ?.name ?? langCode
+    );
+  }
```
sct, line 66: Any specific reason we didn't use an arrow function here?

src/components/RegionSelector/index.tsx
```diff
@@ -39,32 +40,24 @@ const RegionSelector: React.FC<RegionSelectorProps> = ({
...
 
-  const defaultRegionNameFallback =
-    regions?.find((region) => region.iso_3166_1 === currentSettings.region)
-      ?.english_name ?? currentSettings.region;
+  function regionName(regionCode: string): string {
+    return (
+      sortedRegions?.find((region) => region.iso_3166_1 === regionCode)?.name ??
+      regionCode
+    );
+  }
```
sct, line 60: Same here.

sct (approved):

PR #1552, opened 2021-05-01
fix(webpush): only prompt user to allow notifications if enabled in user settings

TheCatLady:
#### Description

Users who opt out of receiving web push notifications should not be prompted to allow notifications in their browser.

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/server/routes/index.ts
+++ b/server/routes/index.ts
@@ -81,10 +81,16 @@ router.get('/status/appdata', (_req, res) => {
 });
 
 router.use('/user', isAuthenticated(), user);
-router.get('/settings/public', async (_req, res) => {
+router.get('/settings/public', async (req, res) => {
   const settings = getSettings();
 
-  return res.status(200).json(settings.fullPublicSettings);
+  if (!req.user?.settings?.notificationTypes.webpush) {
+    return res
+      .status(200)
+      .json({ ...settings.fullPublicSettings, enablePushRegistration: false });
+  } else {
+    return res.status(200).json(settings.fullPublicSettings);
+  }
 });
 router.use(
   '/settings',
--- a/src/components/Settings/Notifications/NotificationsWebPush/index.tsx
+++ b/src/components/Settings/Notifications/NotificationsWebPush/index.tsx
@@ -3,7 +3,7 @@ import { Field, Form, Formik } from 'formik';
 import React, { useState } from 'react';
 import { defineMessages, useIntl } from 'react-intl';
 import { useToasts } from 'react-toast-notifications';
-import useSWR from 'swr';
+import useSWR, { mutate } from 'swr';
 import globalMessages from '../../../../i18n/globalMessages';
 import Button from '../../../Common/Button';
 import LoadingSpinner from '../../../Common/LoadingSpinner';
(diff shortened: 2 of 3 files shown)


sct:
The thing is they will be prompted anyways when web push is enabled and they visit Overseerr for the first time. They can't make it to their settings page to disable it without the popup coming up.

It also creates a strange UX where the user may have allowed it on one machine but then disabled web push. I feel some people might expect the machine they originally hit allow on to still receive notifications.

That's why I decided to not do this. Allowing notifications in the browser is different from allowing web push notifications to be sent at all.

TheCatLady:
> The thing is they will be prompted anyways when web push is enabled and they visit Overseerr for the first time. They can't make it to their settings page to disable it without the popup coming up.

Yes, but they can turn it off to avoid getting prompted on other browsers in the future.

> It also creates a strange UX where the user may have allowed it on one machine but then disabled web push. I feel some people might expect the machine they originally hit allow on to still receive notifications.

If they disabled web push, then they will no longer receive notifications regardless?

sct:
> > The thing is they will be prompted anyways when web push is enabled and they visit Overseerr for the first time. They can't make it to their settings page to disable it without the popup coming up.
> 
> 
> 
> Yes, but they can turn it off to avoid getting prompted on other browsers in the future.

To you this makes sense but to the average user I'm not sure it will.

> > It also creates a strange UX where the user may have allowed it on one machine but then disabled web push. I feel some people might expect the machine they originally hit allow on to still receive notifications.
> 
> 
> 
> If they disabled web push, then they will no longer receive notifications regardless?
> 
> 

Yea but the browser also asked them. They make mistake disabling web push for disabling the browser and then be surprised they don't receive them anymore. I'd rather keep the two things separate.

TheCatLady:
> > > The thing is they will be prompted anyways when web push is enabled and they visit Overseerr for the first time. They can't make it to their settings page to disable it without the popup coming up.
> > 
> > Yes, but they can turn it off to avoid getting prompted on other browsers in the future.
> 
> To you this makes sense but to the average user I'm not sure it will.

If they purposefully disable the notification agent though, wouldn't they expect the feature to be fully disabled?  Otherwise, we're prompting enable notifications and then they may be confused when they never receive any?

> > > It also creates a strange UX where the user may have allowed it on one machine but then disabled web push. I feel some people might expect the machine they originally hit allow on to still receive notifications.
> > 
> > If they disabled web push, then they will no longer receive notifications regardless?
> 
> Yea but the browser also asked them. They make mistake disabling web push for disabling the browser and then be surprised they don't receive them anymore. I'd rather keep the two things separate.

I just think it's best that we are consistent?  It can be confusing for users who have disabled web push to continue to see these prompts?

TheCatLady:
Also, admins may choose to disable the web push agent on behalf of certain users. So if we don't respect the user's web push setting, they will keep getting browser prompts for something that they were never supposed to see.

I disabled web push for my parents because it's too confusing and I already set up email notifications for them, only to find that they still get prompted to allow/block notifications in their browser.

sct (approved):

PR #1563, opened 2021-05-03
fix(ui): improve form usability

TheCatLady:
#### Description

Improves form usability, especially on mobile/touchscreen devices.  Some of these changes were pulled from #1512 since they are not dependent on the other changes in that PR.

- Sets the `inputmode` property for text fields as appropriate, so that the appropriate keyboard type is opened on mobile devices
- Sets `autocomplete` property to `one-time-code` for sensitive fields that are not Overseerr passwords, so that browsers do not autofill or offer to save those inputs
  - Ideally we would simply set `autocomplete` to `off`, but browsers such as Google Chrome do not respect that property and will still autofill fields 😠😩
- Removes input placeholders for text fields, in favor of reasonable default values and always-visible tips/hints in labels (see https://www.nngroup.com/articles/form-design-placeholders/ for reasons why placeholders are usually not ideal)
- Removes the alert boxes from notification agent settings pages (with the exception of the email agent, for which we need user-configurable notification types before we can remove the alert)

Note: This PR is not as big as it seems; most of the lines changed were the result of removing a layer of indentation that is no longer needed with the alerts removed. I could add the outer `<>` and `</>` tags back to minimize the changes in this PR, but I'm personally in favor of removing them.

**Example of removed alert box and added tips/hints to labels:**
image

#### To-Dos

#### Issues Fixed or Closed

N/A


sct:
> Note: This PR is not as big as it seems; most of the lines changed were the result of removing a layer of indentation that is no longer needed with the alerts removed. I could add the outer `<>` and `</>` tags back to minimize the changes in this PR, but I'm personally in favor of removing them.

I straight up can't tell what was changed in half this PR so maybe it makes sense to do it in two steps so I can at least see the changes :)

TheCatLady:
> I straight up can't tell what was changed in half this PR so maybe it makes sense to do it in two steps so I can at least see the changes :)

Oh true, I should have thought of that! I'll revert and split it into two commits 😅

sct (approved):

PR #1574, opened 2021-05-04
fix(locale): default user locale should be the server setting

TheCatLady:
#### Description

The default setting for `Display Language` in user settings should be the server default value.

N/A

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/server/routes/user/usersettings.ts
+++ b/server/routes/user/usersettings.ts
@@ -69,68 +69,68 @@ userSettingsRoutes.get<{ id: string }, UserSettingsGeneralResponse>(
   }
 );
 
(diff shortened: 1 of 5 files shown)


server/routes/user/usersettings.ts
```diff
@@ -69,68 +69,68 @@ userSettingsRoutes.get<{ id: string }, UserSettingsGeneralResponse>(
   }
 );
 
-userSettingsRoutes.post<
-  { id: string },
-  UserSettingsGeneralResponse,
-  UserSettingsGeneralResponse
```
sct, line 75: why did we remove this typing?

sct (approved):

PR #1581, opened 2021-05-05
feat(plex): add support for custom Plex Web App URLs

### LINKED ISSUE #1575: Feature Request: Custom Domains in "Play on Plex"
Description

Currently, the "Play on Plex" button directs users to plex.tv. I would love the ability to set this as a custom domain, much in the same way that Tautulli does this. I direct my users to my own domain always so using the play on plex can confuse some.

 Desired Behavior

Having the ability to set and/or use a custom domain for the "play on plex" button within available content.

 Additional Context

As mentioned, Tautulli does this exactly as I would request. Just as an example of what I mean.

image

TheCatLady:
#### Description

Adds support for custom [Plex Web App URLs](https://support.plex.tv/articles/200288666-opening-plex-web-app/).

Also removes the `Server Name` field from the Plex Settings page, as it serves no real purpose and is confusing. It is not user-editable, and results in weird UX if the user selects a server with a different name from the preset dropdown menu.

Additionally, renames the `Enable SSL` fields for Plex and the *arrs to `Use SSL`, which is more accurate; we aren't enabling SSL for those services, but using SSL to connect to them.

image

#### To-Dos

#### Issues Fixed or Closed

- Closes #1575


src/components/Settings/SettingsPlex.tsx
```diff
@@ -57,6 +55,10 @@ const messages = defineMessages({
   cancelscan: 'Cancel Scan',
   validationHostnameRequired: 'You must provide a valid hostname or IP address',
   validationPortRequired: 'You must provide a valid port number',
+  webAppUrl: 'Web App URL',
+  webAppUrlTip:
+    'Only required if not using the "hosted" <WebAppLink>web app</WebAppLink>',
```
sct, line 60: Not sure how I feel about this language. It's not required even if you don't use the hosted app (everything still works). And why is "hosted" in quotes?

src/components/Settings/SettingsPlex.tsx
```diff
@@ -108,14 +110,26 @@ const SettingsPlex: React.FC<SettingsPlexProps> = ({ onComplete }) => {
...
     hostname: Yup.string()
+      .nullable()
       .required(intl.formatMessage(messages.validationHostnameRequired))
       .matches(
         /^(([a-z]|\d|_|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])([a-z]|\d|-|\.|_|~|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])*)?([a-z]|\d|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])$/i,
         intl.formatMessage(messages.validationHostnameRequired)
       ),
     port: Yup.number()
-      .typeError(intl.formatMessage(messages.validationPortRequired))
+      .nullable()
```
sct, line 120: Why is port nullable? 🤔

src/components/Settings/SettingsPlex.tsx
```diff
@@ -108,14 +110,26 @@ const SettingsPlex: React.FC<SettingsPlexProps> = ({ onComplete }) => {
...
+    webAppUrl: Yup.string()
+      .nullable()
+      .url(intl.formatMessage(messages.validationWebAppUrl))
+      .test(
+        'valid-web-app-url',
+        intl.formatMessage(messages.validationWebAppUrl),
+        (value) =>
+          !value ||
+          value === 'https://app.plex.tv/desktop' ||
+          value.endsWith('/web')
```
sct, line 131: I think I've seen people who have done some fancy Nginx rewrites to get rid of the `/web` or change it to another path.

src/components/Settings/SettingsPlex.tsx
TheCatLady: Ah, that's true about it not being required, I'll reword it. And I put "hosted" in quotes since Plex does in some places as well, but it probably doesn't need to be.

src/components/Settings/SettingsPlex.tsx
TheCatLady: It's not that it's nullable, it's just so that null values are not considered type errors. It's still a required field.  See https://github.com/jquense/yup#mixedrequiredmessage-string--function-schema

src/components/Settings/SettingsPlex.tsx
TheCatLady: Oh hmm.  My concern is just that without a check, users who haven't done that might not enter a valid Plex Web URL, and that would result in none of the play buttons working. 🤔

TheCatLady: Also, there's nothing stopping those users who are doing fancy Nginx rewrites from modifying the value of this setting in their `settings.json` file.  If they're using special not-officially-supported configs to proxy their Plex install, it shouldn't be a problem for them to modify their settings file manually? 🤔

sct: This is sort of an advanced option anyways, no? Nothing is stopping someone from putting in a plex URL that doesn't work anyways. I don't think we should be so restrictive for a field thats entire purpose is to allow custom domains for their plex setup. And we should never design anything with the intent for people to have to manually edit settings.json to get it to work for them?

We can protect users but only to an extent. If they enable and use things they don't understand, we can't help them beyond that.

TheCatLady: That's true. I'll remove the validation and put the `Advanced` badge on this setting then?

sct: I think that's better. Maybe just a check that there is no trailing slash?

TheCatLady: I actually just verified that it still works with a trailing slash! (I tested with `https://plex.mydomain.com/web/`)

sct: Huh, well I guess a double slash in the URL wouldn't be the worst thing but it's technically incorrect.

TheCatLady: @sct It actually doesn't result in a double slash:
```ts
      this.plexUrl = `${
        webAppUrl ? webAppUrl : 'https://app.plex.tv/desktop'
      }#!/server/${machineId}/details?key=%2Flibrary%2Fmetadata%2F${
        this.ratingKey
      }`;
```

sct: Oh, right, they use a hash. Nevermind.

sct (approved):

PR #1584, opened 2021-05-06
fix(locale): set locale based on user settings upon login

TheCatLady:
#### Description

The sign-in page uses the server default locale.  However, the locale should be set to the user's preference once they log in.

Currently, a page refresh is required for the user-set locale to take effect after signing in.

N/A

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/src/components/Layout/index.tsx
+++ b/src/components/Layout/index.tsx
@@ -3,6 +3,9 @@ import { ArrowLeftIcon, InformationCircleIcon } from '@heroicons/react/solid';
 import { useRouter } from 'next/router';
 import React, { useEffect, useState } from 'react';
 import { defineMessages, useIntl } from 'react-intl';
+import { AvailableLocales } from '../../context/LanguageContext';
+import useLocale from '../../hooks/useLocale';
+import useSettings from '../../hooks/useSettings';
 import { Permission, useUser } from '../../hooks/useUser';
 import SearchInput from './SearchInput';
 import Sidebar from './Sidebar';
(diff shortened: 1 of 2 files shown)


src/components/Layout/index.tsx
```diff
@@ -16,9 +19,21 @@ const messages = defineMessages({
...
+
+  useEffect(() => {
+    if (setLocale) {
+      setLocale(
+        (user?.settings?.locale
+          ? user.settings.locale
+          : currentSettings.locale) as AvailableLocales
+      );
+    }
+  }, [setLocale, currentSettings.locale, user]);
```
sct, line 36: This also means though that after the setup is completed for the first time the language will reset back to English.

TheCatLady: Yes, but if the user refreshes the page after completing setup, it will also reset back to English. 🤔

Perhaps at the end of setup, we should save the selected locale to settings?

src/components/Setup/index.tsx
```diff
@@ -32,15 +34,19 @@ const Setup: React.FC = () => {
...
   const finishSetup = async () => {
-    setIsUpdating(false);
+    setIsUpdating(true);
     const response = await axios.post<{ initialized: boolean }>(
       '/api/v1/settings/initialize'
     );
 
     setIsUpdating(false);
     if (response.data.initialized) {
+      await axios.post('/api/v1/settings/main', { locale });
```
sct, line 47: Should we really be changing the global display language here or just the users?

TheCatLady: I think that the server owner's family/friends would be likely to have similar language preferences, so I think that the locale selected during setup is a pretty safe default setting.

TheCatLady: I think that always setting the server default display language to English would be less ideal than using the user-selected setup language (in that it probably isn't what the user would want, and would force them to find and change the setting manually).

sct (approved):

PR #1598, opened 2021-05-07
fix: switch PGP regex to span multiple lines

Jabster28:
#### Description

copying PGP keys from any PGP client almost never results in a single-lined key, and as such I don't think the input regex should require it all to be in one line, as the openPGP NPM module allows for whitespace in their keys.

N/A
#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


TheCatLady:
The current regex doesn't require the key to be on a single line, however.  It just requires that it begins with `-----BEGIN PGP PUBLIC KEY BLOCK-----` and ends with `-----END PGP PUBLIC KEY BLOCK-----`.

TheCatLady:
You also need to update the regex in `src/components/Settings/Notifications/NotificationsEmail.tsx`. 🙂

sct (approved):
LGTM :) Thanks for this!

PR #1599, opened 2021-05-07
feat(webpush): add warning to web push settings re: HTTPS requirement

TheCatLady:
#### Description

Web push notifications only work when Overseerr is being served over HTTPS, so we should warn users about the requirement.

This PR adds a warning to the web push settings page that only appears when accessed over HTTP.

**Accessed via HTTP:**
image

**Accessed via HTTPS:**
image

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/src/components/Settings/Notifications/NotificationsWebPush/index.tsx
+++ b/src/components/Settings/Notifications/NotificationsWebPush/index.tsx
@@ -5,6 +5,7 @@ import { defineMessages, useIntl } from 'react-intl';
 import { useToasts } from 'react-toast-notifications';
 import useSWR, { mutate } from 'swr';
 import globalMessages from '../../../../i18n/globalMessages';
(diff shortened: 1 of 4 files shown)


src/components/Settings/Notifications/NotificationsWebPush/index.tsx
```diff
@@ -32,6 +35,12 @@ const NotificationsWebPush: React.FC = () => {
 
   return (
     <>
+      {!window.location.protocol.startsWith('https') && (
+        <Alert
+          title={intl.formatMessage(messages.httpsRequirement)}
+          type="warning"
+        />
+      )}
```
sct, line 43: This will crash on server side render because window will not exist. We need to check in a useEffect which only runs in the client and set state based on the result.

PR #1646, opened 2021-05-16
fix(locale): properly restore display language upon page refresh

### LINKED ISSUE #1645: Language fallsback to App's Settings when refreshing webpage
#### Description

App's General Settings> Display Language , overrides users' display language choices once page is refreshed. Tried it with Greek, English, Spanish, Dutch and German.

#### Version

Develop. 22415642e8602809e3507e5b13dc2f8de3000003

#### Steps to Reproduce and their results

So as an Admin. App is set to English as default language. 

1) I log in into a default user and change the language to something else, such as Greek. Result=whole UI changes to Greek.
2) I then refresh or click on overseerr logo (to refresh) which then reverts back to App's general language setting which is English.

So it override's user's choice to display selected language. In order to fix it I need to log out and log in again and Greek is shown again (if not refreshed). Same happens with other languages as well.

#### Screenshots
**App's Settings. Display language set on English**
Screenshot_2021-05-16 General - Settings - Overseerr

**User's account when logged in. Still not refreshed.**
USER'S ACCOUNT WHEN LOGGED IN - NO REFRESH

**User's account after refreshing. It revert's back to app's general settings. Overriding user's choice.**
user's account after refreshing

#### Device
Tried it with Edge and Firefox. Both in Private and not.
Server is ubuntu server 18.04
Running overseerr in docker

TheCatLady:
#### Description

Currently, the user's locale setting is being lost upon page refresh.

N/A

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1645

### DIFF
--- a/server/middleware/auth.ts
+++ b/server/middleware/auth.ts
@@ -11,7 +11,7 @@ export const checkUser: Middleware = async (req, _res, next) => {
 
     let userId = 1; // Work on original administrator account
 
-    // If a User ID is provided, we will act on that users behalf
+    // If a User ID is provided, we will act on that user's behalf
     if (req.header('X-API-User')) {
       userId = Number(req.header('X-API-User'));
     }
@@ -20,6 +20,10 @@ export const checkUser: Middleware = async (req, _res, next) => {
     if (user) {
       req.user = user;
     }
+
+    req.locale = user?.settings?.locale
+      ? user.settings.locale
+      : settings.main.locale;
   } else if (req.session?.userId) {
     const userRepository = getRepository(User);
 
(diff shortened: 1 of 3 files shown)


server/middleware/auth.ts
```diff
@@ -29,10 +33,13 @@ export const checkUser: Middleware = async (req, _res, next) => {
...
       req.user = user;
-      req.locale = user.settings?.locale
-        ? user.settings?.locale
-        : settings.main.locale;
     }
+
+    req.locale = user?.settings?.locale
+      ? user.settings.locale
+      : settings.main.locale;
+  } else {
```
sct, line 41: Why not just do all of this once outside of this block before the `next()`?

TheCatLady: Hm yeah we could do that!

sct (approved):

PR #1666, opened 2021-05-18
fix(tv): don't show duplicate air date

danshilm:
#### Description

Only show the Next Air Date if it is not the same as the First Air Date

Fixes:
Screenshot_201

#### To-Dos

#### Issues Fixed or Closed

- Fixes none


TheCatLady:
I think you mean `Only show the Next Air Date if it is not the same as the First Air Date` 😉  But LGTM!

sct (approved):

sct:
Thanks :D

PR #1699, opened 2021-05-26
fix(plex): do not fail to import Plex users when Plex Home has managed users

### LINKED ISSUE #1698: Import all users from Plex
#### Description

I have a few managed users, which don't have email addresses. I saw in the code there was a .toLowerCase() causing query to fail and I assume it's because emails are null. 

Where I assume the problem is, [link](https://github.com/sct/overseerr/blob/928b8a71cf361b7bc2b8957c621f5b66c4657b1e/server/routes/user/index.ts#L88)

#### Version

e88042ba28f0ec02b67f55273835a29dffd95851, latest develop.

#### Steps to Reproduce

Please tell us how we can reproduce the undesired behavior.

1. Have managed users on plex
2. Import users from plex
3. Error displayed in console for 

#### Expected Behavior

Import successfully

#### Screenshots

image

TheCatLady:
#### Description

Managed users do not have email addresses, which causes the DB query to check for an existing user to fail when we attempt to lowercase the e-mail address.

Also:
- ~~Do not set username when it is the same as the user's email address~~
- Fix weird UX when a user doesn't have a username set
- Make the display name sort on the user list case-insensitive
- Set appropriate placeholder for display name field (so it reflects what the user's display name would be if they do not set a username)
- Hide user email addresses from users without the `Manage Users` permission (unless the user does not have a display name set, in which case their email address is also their display name)

N/A

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1698

### DIFF
--- a/server/entity/User.ts
+++ b/server/entity/User.ts
@@ -48,7 +48,13 @@ export class User {
   @PrimaryGeneratedColumn()
   public id: number;
 
-  @Column({ unique: true })
+  @Column({
+    unique: true,
+    transformer: {
+      from: (value: string): string => value.toLowerCase(),
+      to: (value: string): string => value.toLowerCase(),
+    },
+  })
   public email: string;
 
   @Column({ nullable: true })
--- a/server/entity/User.ts
+++ b/server/entity/User.ts
@@ -52,7 +52,7 @@ export class User {
   public email: string;
 
   @Column({ nullable: true })
-  public plexUsername: string;
+  public plexUsername?: string;
 
   @Column({ nullable: true })
   public username?: string;
@@ -220,7 +220,7 @@ export class User {
 
   @AfterLoad()
   public setDisplayName(): void {
-    this.displayName = this.username || this.plexUsername;
+    this.displayName = this.username || this.plexUsername || this.email;
   }
 
   public async getQuota(): Promise<QuotaResponse> {
--- a/server/routes/auth.ts
+++ b/server/routes/auth.ts
@@ -57,7 +57,10 @@ authRoutes.post('/plex', async (req, res, next) => {
       // Update the user's avatar with their Plex thumbnail, in case it changed
       user.avatar = account.thumb;
       user.email = account.email;
-      user.plexUsername = account.username;
+      user.plexUsername =
+        account.username.toLowerCase() === account.email.toLowerCase()
+          ? ''
+          : account.username;
 
       // In case the user was previously a local account
       if (user.userType === UserType.LOCAL) {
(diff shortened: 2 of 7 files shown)


src/components/UserProfile/ProfileHeader/index.tsx
```diff
@@ -65,11 +65,13 @@ const ProfileHeader: React.FC<ProfileHeaderProps> = ({
...
-                ({user.email})
-              </span>
-            )}
+            {user.displayName.toLowerCase() !== user.email.toLowerCase() &&
+              (user.id === loggedInUser?.id ||
+                hasPermission(Permission.MANAGE_USERS)) && (
+                <span className="text-sm text-gray-400 sm:text-lg sm:ml-2">
+                  ({user.email})
+                </span>
+              )}
```
sct, line 74: We already filter email address from the API responses if the user doesn't have the `MANAGE_USERS` permission I the API.

server/entity/User.ts
```diff
@@ -52,7 +52,7 @@ export class User {
   public email: string;
 
   @Column({ nullable: true })
-  public plexUsername: string;
+  public plexUsername?: string;
```
sct, line 55: This should no longer be optional.

src/components/UserProfile/ProfileHeader/index.tsx
```diff
@@ -65,9 +65,8 @@ const ProfileHeader: React.FC<ProfileHeaderProps> = ({
                 {user.displayName}
               </a>
             </Link>
-            {user.displayName.toLowerCase() !== user.email.toLowerCase() &&
-              (user.id === loggedInUser?.id ||
-                hasPermission(Permission.MANAGE_USERS)) && (
+            {user.email &&
+              user.displayName.toLowerCase() !== user.email.toLowerCase() && (
```
sct, line 69: Why not just force the response from the api endpoint to lowercase everything so we don't need to do this all over the app?

TheCatLady: @sct Ahh, good call!

server/routes/auth.ts
```diff
@@ -43,7 +43,7 @@ authRoutes.post('/plex', async (req, res, next) => {
     let user = await userRepository
       .createQueryBuilder('user')
       .where('user.plexId = :id', { id: account.id })
-      .orWhere('LOWER(user.email) = :email', {
+      .orWhere('user.email = :email', {
```
sct, line 46: I actually wonder if this will work here. Since this is a direct query on the database I don't know if the transformer will be applied.

TheCatLady: @sct It does, just verified!

sct (approved):

PR #1724, opened 2021-06-01
fix: check that application URL and email agent are configured for password reset/generation

TheCatLady:
#### Description

Password reset and generation require both an application URL to be set and the email agent to be enabled.

Currently, we do not enforce this... so password reset/generation emails are sent with broken links when the application URL is not configured.

- Do not show the password reset button if requirements are not met
- Enforce and make note of the application URL requirement in the `Create Local User` modal, and only show the alert to users with the `Manage Settings` permission
- Remove weird gray line above local login / password reset fields

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/src/components/Login/LocalLogin.tsx
+++ b/src/components/Login/LocalLogin.tsx
@@ -5,6 +5,7 @@ import Link from 'next/link';
 import React, { useState } from 'react';
 import { defineMessages, useIntl } from 'react-intl';
 import * as Yup from 'yup';
+import useSettings from '../../hooks/useSettings';
 import Button from '../Common/Button';
 import SensitiveInput from '../Common/SensitiveInput';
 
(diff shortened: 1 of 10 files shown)


src/components/Login/LocalLogin.tsx
```diff
@@ -101,17 +107,23 @@ const LocalLogin: React.FC<LocalLoginProps> = ({ revalidate }) => {
...
-                        <span>
-                          {intl.formatMessage(messages.forgotpassword)}
-                        </span>
-                      </Button>
-                    </Link>
-                  </span>
+                <div
+                  className={`flex ${
+                    passwordResetEnabled ? 'justify-between' : 'justify-end'
+                  }`}
```
sct, line 113: If we reverse the flex direction here, you can leave it as justify-between and don't need a conditional.

TheCatLady: Done!

sct (approved):

PR #1741, opened 2021-06-06
docs: adding npm section reverse-proxy

HubDuck:
Adds npm configuration example to the docs for the reverse-proxy page under extending-overseerr.

Adds example image to public/images/docs

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


docs/extending-overseerr/reverse-proxy.md
```diff
@@ -39,6 +39,42 @@ server {
...
+
+9 Enable “HTTP/2 Support”
+
+If you requested a new SSL Certificate in step 7 continue, otherwise click Save
+
+10 Enter your Email address
+
+11 Enable “I Agree to the Let’s Encrypt Terms of Service” if you agree
+
+Click Save
```
TheCatLady, line 76:
```suggestion
Add a new proxy host with the following settings:

### Details

- **Domain Names:** Your desired external Overseerr hostname; e.g., `overseerr.example.com`
- **Scheme:** `http`
- **Forward Hostname / IP:** Internal Overseerr hostname or IP
- **Forward Port:** `5055`
- **Cache Assets:** yes
- **Block Common Exploits:** yes

### SSL

- **SSL Certificate:** Select one of the options; if you are not sure, pick “Request a new SSL Certificate”
- **Force SSL:** yes
- **HTTP/2 Support:** yes
```

While nice, I don't think the image is necessary here.  We would also need to update it anytime NPM makes changes to their UI, and I don't think we want to be responsible for maintaining that.

samwiseg0:
I think this would be better under the Nginx section. Can you add a third tab called `Proxy Manager`?

TheCatLady:
> I think this would be better under the Nginx section. Can you add this under a third tab called `Proxy Manager` with the changes @TheCatLady suggested?

I agree that this would be better as a tab in that section.  Good call, @samwiseg0!

docs/extending-overseerr/reverse-proxy.md
HubDuck: Since swag also uses nginx, should we put all nginx solutions into one tab?

In my opinion npm is more than just nginx, exactly like swag is.

HubDuck:
> I think this would be better under the Nginx section. Can you add this under a third tab called `Proxy Manager` with the changes @TheCatLady suggested?

Since swag also uses nginx, should we put all nginx solutions into one tab?

In my opinion npm is more than just nginx, exactly like swag is.

TheCatLady:
> > I think this would be better under the Nginx section. Can you add this under a third tab called `Proxy Manager` with the changes @TheCatLady suggested?
> 
> Since swag also uses nginx, should we put all nginx solutions into one tab?
> 
> In my opinion npm is more than just nginx, exactly like swag is.

Hmmm.  I do see what you mean.  Since there are some instructions specific to the "regular" Nginx setup below the two tabs, I think it'd be okay to have NPM as its own section... but it probably should be at the bottom of the page, after the Nginx section.

@samwiseg0, what do you think? 🤔  I like the idea of putting all the Nginx configs together, but if we want to go that route, the refactor should probably be a separate PR.

TheCatLady (approved):

PR #1761, opened 2021-06-08
feat(ui): link processing/requested status badges to service URL

TheCatLady:
#### Description

Similar to how available / partially available status badges link to Plex, make processing / requested status badges clickable links to the *arrs (for users with the `Manage Requests` permission only).

Also refactors the `StatusBadge` component to dedupe logic and fix missing `inProgress` spinner display for 4K available / partially available media items.

#### To-Dos

#### Issues Fixed or Closed

N/A


src/components/StatusBadge/index.tsx
```diff
@@ -14,94 +15,18 @@ interface StatusBadgeProps {
   is4k?: boolean;
   inProgress?: boolean;
   plexUrl?: string;
-  plexUrl4k?: string;
+  serviceUrl?: string;
```
sct, line 18: What about adding a `url` prop instead that is just used whenever it's present? That makes the API usage clearer I feel.

TheCatLady: But we display the Plex vs. service URL conditionally based on the status. If we only have a single URL prop, we'd end up repeating that conditional logic everywhere we show a status badge?

TheCatLady: It would probably make sense to add a `url` prop to the `Badge` component though... 🤔

sct: Ah, yeah, maybe on the Badge component then.

TheCatLady: Already done in aa6029ac2a887e43663391fdf4c4fab2ad87feef 😉

sct:
I think this should maybe only show for Admins? Just because I give a user permission to manage requests it doesn't mean they necessarily have access to radarr/sonarr.

TheCatLady:
> I think this should maybe only show for Admins? Just because I give a user permission to manage requests it doesn't mean they necessarily have access to radarr/sonarr.

Oh hmm, good point!  I'll make that change.

PR #1770, opened 2021-06-10
feat(ui): add 'show more/less...' for studios on movie details page

### LINKED ISSUE #1489: Small enhancement for the studio list on detail page
#### Description
The studio's list on the details page is in some cases really long. It will be great if there are some limitations in place to prevent a long list of studios.

#### Desired Behavior
Maybe a text something like **See more** or **Show all**

#### Additional Context
image

dphildebrandt:
#### Description

Display a max of 3 studios before prompting to expand the list

studios

#### To-Dos

#### Issues Fixed or Closed

- Fixes #1489

### DIFF
--- a/src/components/MovieDetails/index.tsx
+++ b/src/components/MovieDetails/index.tsx
@@ -91,7 +91,7 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
   const { locale } = useLocale();
   const [showManager, setShowManager] = useState(false);
   const minStudios = 3;
-  const [studiosToShow, setStudiosToShow] = useState(minStudios);
+  const [showMoreStudios, setShowMoreStudios] = useState(false);
 
   const { data, error, revalidate } = useSWR<MovieDetailsType>(
     `/api/v1/movie/${router.query.movieId}`,
@@ -118,7 +118,6 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
   }
 
   const showAllStudios = data.productionCompanies.length <= minStudios + 1;
-
   const mediaLinks: PlayButtonLink[] = [];
 
   if (
@@ -638,9 +637,9 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
                   {data.productionCompanies
                     .slice(
                       0,
-                      showAllStudios
+                      showAllStudios || showMoreStudios
                         ? data.productionCompanies.length
-                        : studiosToShow
+                        : minStudios
                     )
                     .map((s) => {
                       return (
@@ -656,20 +655,16 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
                     <button
                       type="button"
                       onClick={() => {
-                        setStudiosToShow(
-                          studiosToShow === minStudios
-                            ? data.productionCompanies.length
-                            : minStudios
-                        );
+                        setShowMoreStudios(!showMoreStudios);
                       }}
                     >
                       <span className="flex items-center">
                         {intl.formatMessage(
-                          studiosToShow === minStudios
+                          !showMoreStudios
                             ? messages.showmore
                             : messages.showless
                         )}
-                        {studiosToShow === minStudios ? (
+                        {!showMoreStudios ? (
                           <ChevronDoubleDownIcon className="w-4 h-4 ml-1" />
                         ) : (
                           <ChevronDoubleUpIcon className="w-4 h-4 ml-1" />
--- a/src/components/MovieDetails/index.tsx
+++ b/src/components/MovieDetails/index.tsx
@@ -6,6 +6,8 @@ import {
 } from '@heroicons/react/outline';
 import {
   CheckCircleIcon,
+  ChevronDoubleDownIcon,
+  ChevronDoubleUpIcon,
   DocumentRemoveIcon,
   ExternalLinkIcon,
 } from '@heroicons/react/solid';
@@ -73,6 +75,8 @@ const messages = defineMessages({
   play4konplex: 'Play in 4K on Plex',
   markavailable: 'Mark as Available',
   mark4kavailable: 'Mark as Available in 4K',
+  showmore: 'Show More',
+  showless: 'Show Less',
 });
 
 interface MovieDetailsProps {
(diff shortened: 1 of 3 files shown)


TheCatLady:
We should never use `...` instead of the ellipsis character, but I don't really think this looks great as text.  Perhaps an icon would look better?  Solid variants of `chevron-double-down`/`chevron-double-up` maybe? 🤔

sct:
> We should never use `...` instead of the ellipsis character, but I don't really think this looks great as text. Perhaps an icon would look better? Solid variants of `chevron-double-down`/`chevron-double-up` maybe? Or the `dots-horizontal` icon? 🤔

Chevron would be better for this I feel. We should also change the color to indicate the user can interact with this and its just not another line of text?

TheCatLady:
> Chevron would be better for this I feel. We should also change the color to indicate the user can interact with this and its just not another studio?

Well, FWIW users can interact with the studios in the list as well 😛

sct:
> > Chevron would be better for this I feel. We should also change the color to indicate the user can interact with this and its just not another studio?
> 
> Well, FWIW users can interact with the studios in the list as well 😛

True, the icon may be enough to differentiate it.

dphildebrandt:
Updated screenshot in PR body.

src/components/MovieDetails/index.tsx
```diff
@@ -111,6 +117,9 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
     return <Error statusCode={404} />;
   }
 
+  const studios = data.productionCompanies;
+  const showAllStudios = studios.length - minStudios <= minStudios;
```
TheCatLady, line 121:
```suggestion
  const showAllStudios = studios.length - minStudios < 2;
```
I think it's only necessary to force the display of all studios when the `Show More`/`Show Less` buttons wouldn't actually save any space (i.e., when the button just replaces a single entry). 🤔

We can simply increase the minimum value (5 seems reasonable?) if we want this behavior to trigger less often.

But I'm sure @sct has an option on this, so maybe see that he says about it.

dphildebrandt: Took your suggestion, but tweaked the logic.

src/components/MovieDetails/index.tsx
```diff
@@ -245,8 +254,8 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
         onClose={() => setShowManager(false)}
         subText={data.title}
       >
-        {((data?.mediaInfo?.downloadStatus ?? []).length > 0 ||
-          (data?.mediaInfo?.downloadStatus4k ?? []).length > 0) && (
+        {(!!(data?.mediaInfo?.downloadStatus ?? []).length ||
```
sct, line 257: Why exactly are we refactoring all of these conditionals? 🤔

dphildebrandt: Ah, I was adopting the new style TheCatLady was using in her PR #1778. I can definitely leave out this change if preferred. 😊

dphildebrandt: Oops, removed from this PR

TheCatLady:
@dphildebrandt You don't need to squash your commits down btw, we squash before merging anyway 😄 Makes it hard to tell what changed when you do that.

src/components/MovieDetails/index.tsx
```diff
@@ -111,6 +117,9 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
     return <Error statusCode={404} />;
   }
 
+  const studios = data.productionCompanies;
```
TheCatLady, line 120: I don't think this is necessary 😅

Can also revert the changes on lines 631, 635, and 639.

dphildebrandt:
> @dphildebrandt You don't need to squash your commits down btw, we squash before merging anyway 😄 Makes it hard to tell what changed when you do that.

Ah, I wasn't sure. I was trying to keep the commits atomic, and didn't think some of these changes warranted a full-blown husky-compatible commit message.

src/components/MovieDetails/index.tsx
dphildebrandt: 👍

src/components/MovieDetails/index.tsx
```diff
@@ -619,24 +627,56 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
                 </span>
               </div>
             )}
-            {data.productionCompanies.length > 0 && (
+            {!!data.productionCompanies.length && (
```
TheCatLady, line 630: As @sct mentioned already, no need to change this. 😄

TheCatLady:
> Ah, I wasn't sure. I was trying to keep the commits atomic, and didn't think some of these changes warranted a full-blown husky-compatible commit message.

I think that it makes it harder to difficult to review and see what changed since the last time we viewed the PR.  Best to just add separate commits, especially once you've opened a PR and people have begun reviewing it.

src/components/MovieDetails/index.tsx
```diff
@@ -86,6 +90,8 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
   const intl = useIntl();
   const { locale } = useLocale();
   const [showManager, setShowManager] = useState(false);
+  const minStudios = 3;
+  const [studiosToShow, setStudiosToShow] = useState(minStudios);
```
TheCatLady, line 94: Perhaps this should be a boolean?  Since it is only ever one of two values (`minStudios` or the number of elements in the array).  Would then be able to eliminate the `studiosToShow === minStudios` comparisons below.

src/components/MovieDetails/index.tsx
```diff
@@ -111,6 +117,7 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
     return <Error statusCode={404} />;
   }
 
+  const showAllStudios = data.productionCompanies.length <= minStudios + 1;
```
TheCatLady, line 120: Perhaps move this up to where the other new `const`s are?

TheCatLady: PR looks good to me though :)

dphildebrandt: I don't think I can because `data` isn't guaranteed to be defined yet.

TheCatLady: Ah you're right, whoops! :P

sct (approved):
Sorry for the long wait! Things have been super busy!

Looks great! Thanks for the contribution again!

Magikarplvl4:
Thanks guys and girls for this! <3

PR #1779, opened 2021-06-11
feat(discord): add 'Enable Mentions' setting

TheCatLady:
#### Description

Some admins may be sending Discord notifications to a channel that their users do not have access to.

This PR adds a setting to the Discord agent, to allow admins to enable/disable mentions (and the notification type selector in user settings).

By default, this setting is enabled, so the behavior is unchanged for users already utilizing the Discord agent.

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/overseerr-api.yml
+++ b/overseerr-api.yml
@@ -1138,6 +1138,8 @@ components:
               type: string
             webhookUrl:
               type: string
+            enableMentions:
+              type: boolean
     SlackSettings:
       type: object
       properties:
--- a/server/lib/notifications/agents/discord.ts
+++ b/server/lib/notifications/agents/discord.ts
@@ -258,35 +258,37 @@ class DiscordAgent
     const userMentions: string[] = [];
 
     try {
-      if (payload.notifyUser) {
-        if (
-          payload.notifyUser.settings?.hasNotificationType(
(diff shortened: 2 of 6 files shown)


sct:
If you mention a user in a channel with no access does it actually have any effect? They won't get pinged or anything, I believe.

TheCatLady:
> If you mention a user in a channel with no access does it actually have any effect? They won't get pinged or anything, I believe.

True, but users will see the option to enable mentions in Overseerr, and may get confused as to why they aren't getting notified on Discord.

sct:
> > If you mention a user in a channel with no access does it actually have any effect? They won't get pinged or anything, I believe.
> 
> True, but users will see the option to enable mentions in Overseerr, and may get confused as to why they aren't getting notified on Discord.

I'm not sure how this PR stops them from seeing the ability to enable mentions? 🤔  Am I missing something?

TheCatLady:
> I'm not sure how this PR stops them from seeing the ability to enable mentions? 🤔 Am I missing something?

It removes the notification type selector from the Discord user notification settings, so that users cannot select notifications to receive.

sct (approved):

PR #1784, opened 2021-06-12
refactor(ui): break out streaming providers into its own component

dphildebrandt:
#### Description

Removes duplicate code.


dphildebrandt:
@TheCatLady This is for you :)

PR #1807, opened 2021-06-17
fix(ui): center logo on password reset pages

TheCatLady:
#### Description

The new logo is not being centered on the password reset pages.

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/src/components/Login/index.tsx
+++ b/src/components/Login/index.tsx
@@ -78,11 +78,7 @@ const Login: React.FC = () => {
         <LanguagePicker />
       </div>
       <div className="relative z-40 flex flex-col items-center px-4 mt-10 sm:mx-auto sm:w-full sm:max-w-md">
-        <img
-          src="/logo_stacked.svg"
-          className="max-w-full mx-auto mb-10"
-          alt="Logo"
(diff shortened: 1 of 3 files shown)


src/components/Login/index.tsx
```diff
@@ -78,7 +78,11 @@ const Login: React.FC = () => {
         <LanguagePicker />
       </div>
       <div className="relative z-40 flex flex-col items-center px-4 mt-10 sm:mx-auto sm:w-full sm:max-w-md">
-        <img src="/logo_stacked.svg" className="max-w-full mb-10" alt="Logo" />
+        <img
+          src="/logo_stacked.svg"
+          className="max-w-full mx-auto mb-10"
```
sct, line 83: We are using flex to center here. Let's do the same in the other places? We shouldn't need mx-auto then.

sct (changes requested):

src/components/Login/index.tsx
TheCatLady: Gotcha!

sct (approved):

Issue #1858, opened 2021-06-30
request limit not working


arjohnson15:
#### Description

Request limit on users isn't preventing them from requesting more. They are just showing negative numbers as they past the limit. I am unsure if its a bug in my install or the version. 

#### Version

What version of Overseerr are you running? 1.25.0

#### Steps to Reproduce

Please tell us how we can reproduce the undesired behavior.

1. set a request limit of one to a user
2. have that user request 2 movies
3. check the users profile 

#### Expected Behavior

I expext the limit to stop them from requesting when they have 0 remaining requests like it did in 1.24.0

#### Screenshots

![image](https://user-images.githubusercontent.com/49290460/123969216-9e9a8c00-d97d-11eb-9803-0018343a0f5d.png)

#### Device

What device were you using when you encountered this issue? Please provide this information to help us reproduce and investigate the bug.

- Overseerr running in a docker
-Unsure what device they are using to request from

danshilm:
I'm unable to reproduce this since I can't request content if the limit has been set. 

The only thing I can reproduce is the negative number as the amount of requests left for the 7 day period by already having x amount of content requested then setting the limit for that user to less than the value of x.

arjohnson15:
I will attempt to install another instance and see if my install had an issue. @danshilm that is how it worked before I upgraded to 1.25.0 so it is possible an issue arose during my install. 
In 1.24.0 I had the limit set on a few individual users then in 1.25.0 I set it globally and they were able to request past the limit. I did start fresh on 1.25.0.

bison529:
> I will attempt to install another instance and see if my install had an issue. @danshilm that is how it worked before I upgraded to 1.25.0 so it is possible an issue arose during my install.
> In 1.24.0 I had the limit set on a few individual users then in 1.25.0 I set it globally and they were able to request past the limit. I did start fresh on 1.25.0.

I have same issue, local users can they were able to request past the limit.

PR #1859, opened 2021-06-30
fix(ui): do not display negative remaining quota

TheCatLady:
#### Description

If a user has exceeded their quota, we should display the remaining limit as 0 instead of a negative value.

#### To-Dos

#### Issues Fixed or Closed

- Fixes UI issue reported in #1858

### DIFF
--- a/src/components/UserProfile/index.tsx
+++ b/src/components/UserProfile/index.tsx
@@ -150,7 +150,10 @@ const UserProfile: React.FC = () => {
                           limit: (
                             <span className="text-3xl font-semibold">
                               {intl.formatMessage(messages.limit, {
-                                remaining: quota.movie.remaining,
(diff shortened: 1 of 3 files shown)


src/components/UserProfile/index.tsx
```diff
@@ -150,7 +150,10 @@ const UserProfile: React.FC = () => {
                           limit: (
                             <span className="text-3xl font-semibold">
                               {intl.formatMessage(messages.limit, {
-                                remaining: quota.movie.remaining,
+                                remaining: Math.max(
+                                  0,
+                                  quota.movie.remaining ?? 0
```
sct, line 155: We should probably correct this on the API side

sct (approved):

PR #1904, opened 2021-07-15
feat: adding embedded trailers to desktop and deep linking to mobile 

### LINKED ISSUE #962: Embedded  / Popup Trailers instead of linking
No, but don't like having a new tab/window opened. :)

Instead of an external link to a trailer, load it inside overseerr. I'm sure there's some term for it I don't know but like a popup/embedded option. 

**Additional context**
An example would be TMDB 
image

### LINKED ISSUE #1576: "Watch Trailer" (on mobile browser) opens unnecessary blank tab in browser
#### Description

On a mobile browser, when you press "Watch Trailer," a blank, new page tab is opened, then the user is redirected to the YouTube app. Then when you switch back to the browser, you need to delete the newly opened blank tab and go back to Overseerr

#### Version

1.23.2

#### Steps to Reproduce

(ON MOBILE DEVICE BROWSER)

1. View Media (Movie or Series)
2. Tap "Watch Trailer"

#### Expected Behavior

When the user clicks Watch Trailer, blank tab should not be created. Should redirect to YouTube / open YouTube app without blank tab. 

#### Device

What device were you using when you encountered this issue? Please provide this information to help us reproduce and investigate the bug.

- **Platform:** smartphone
- **Device:** iPhone 11
- **OS:** iOS 8.1
- **Browser:** Chrome & Firefox tested

lukepolo:
feat: adding deep link for youtube trailers

#### Description
Adding Embedded Trailers for Desktop / Adding Deep Linking Capabilities for mobile

image

#### To-Dos

#### Issues Fixed or Closed

- Fixes #962
- Fixed #1576

Im sure you will want to move a couple of things around , also this assumes that only `youtube` is used (which is bad IMO).

Let me know if you would like me to move forward with this.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -28,6 +28,7 @@ import { getSettings } from './lib/settings';
 import logger from './logger';
 import routes from './routes';
 import { getAppVersion } from './utils/appVersion';
+import { URL } from 'url';
 
 const API_SPEC_PATH = path.join(__dirname, '../overseerr-api.yml');
 
@@ -166,6 +167,28 @@ app
       next();
     });
     server.use('/api/v1', routes);
+    server.get('/deep-link', (req, res) => {
+      const url = req.query.url as string;
+      if (!url) {
+        res.status(400);
+        res.send('URL was not supplied.');
+        return;
+      }
+
+      const parsedUrl = new URL(url);
+      const deepLinkURL = parsedUrl.searchParams.get('v');
+
+      res.send(`
+        <script type="text/javascript">
+            window.onload = function() {
+                window.location = 'youtube://${deepLinkURL}';
+                window.setTimeout(function() {
+                    window.location = "${url}";
+                }, 25);
+            };
+        </script>
+      `);
+    });
     server.get('*', (req, res) => handle(req, res));
     server.use(
       (
--- a/server/index.ts
+++ b/server/index.ts
@@ -168,7 +168,7 @@ app
     });
     server.use('/api/v1', routes);
     server.get('/deep-link', (req, res) => {
-      const url = req.query.url as string;
+      const url = encodeURI(req.query.url as string);
       if (!url) {
         res.status(400);
         res.send('URL was not supplied.');
--- a/src/components/Common/PlayButton/index.tsx
+++ b/src/components/Common/PlayButton/index.tsx
@@ -8,8 +8,8 @@ interface PlayButtonProps {
 }
 
 export enum PlayButtonLinkTypes {
-  Web = 'web',
-  Trailer = 'trailer',
+  WEB = 'web',
+  TRAILER = 'trailer',
 }
 
 export interface PlayButtonLink {
--- a/src/components/Common/PlayButton/index.tsx
+++ b/src/components/Common/PlayButton/index.tsx
@@ -1,50 +1,83 @@
-import React, { ReactNode } from 'react';
+import React, { ReactNode, useState } from 'react';
+
 import ButtonWithDropdown from '../ButtonWithDropdown';
+import Trailer from '../../Trailer';
 
 interface PlayButtonProps {
   links: PlayButtonLink[];
 }
 
+export enum PlayButtonLinkTypes {
+  Web = 'web',
+  Trailer = 'trailer',
+}
+
 export interface PlayButtonLink {
+  type?: string;
   text: string;
   url: string;
   svg: ReactNode;
 }
 
 const PlayButton: React.FC<PlayButtonProps> = ({ links }) => {
+  const [viewingUrl, setViewingUrl] = useState<string | undefined>();
+
+  function openLink(link: PlayButtonLink): void {
+    switch (link.type) {
+      case PlayButtonLinkTypes.Trailer:
+        if (/Android|iPhone|iPad|iPod/i.test(navigator.userAgent)) {
+          window.open(`/deep-link?url=${link.url}`, '_blank');
+          return;
+        }
+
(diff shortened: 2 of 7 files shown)


src/components/Common/PlayButton/index.tsx
```diff
@@ -1,51 +1,104 @@
...
+  type?: string;
   text: string;
   url: string;
   svg: ReactNode;
 }
 
-const PlayButton: React.FC<PlayButtonProps> = ({ links }) => {
-  if (!links || !links.length) {
-    return null;
+const PlayButton = class PlaybuttonClass extends React.Component<
```
sct, line 25: Why did you convert this to a class component?

server/index.ts
```diff
@@ -144,6 +145,28 @@ app
...
+        <script type="text/javascript">
+            window.onload = function() {
+                window.location = 'youtube://${deepLinkURL}';
+                window.setTimeout(function() {
+                    window.location = "${url}";
+                }, 25);
+            };
+        </script>
+      `);
+    });
```
sct, line 169: I'm not sure what the purpose of this endpoint is?

.nvmrc
```diff
@@ -0,0 +1 @@
+14.17
```
sct, line 1: Lets not commit this file please.

server/api/themoviedb/interfaces.ts
```diff
@@ -1,3 +1,5 @@
+import { VideoType } from '../../models/Movie';
```
sct, line 1: It doesn't make sense to import a type from models into the typing for TMDB API. If anything it should be the other way around, or just have two mirrored, compatible types.

server/index.ts
lukepolo: this allows for deeplinking, without it it keeps a tab open on mobile : fixes #1576

src/components/Common/PlayButton/index.tsx
lukepolo: as my PR states, i have basic knowledge of react at best , but I believe it's because i've made it non function as im changing state within it. 

If this is possible via functional component still let me know and ill go read more docs

server/api/themoviedb/interfaces.ts
lukepolo: agreed

lukepolo: I've mirrored them for now , probably requires more clean up, as the Movie ones make sure its a youtube but the TV does not.

server/index.ts
sct: Okay sure I can understand why you want to do this but why not just handle this on the JS side in the frontend instead of doing it in the API?

src/components/Common/PlayButton/index.tsx
sct: It's very possible! You use `useState` to accomplish it. Take a look at our other components and look into React hooks. Class components are super bug-prone so we don't use them at all in Overseerr.

server/index.ts
sct: Also what happens if the user doesn't have the youtube app installed on their phone?

lukepolo: It has to be a new window for this to work , if it YouTube isn’t on their phone it will go to the YouTube site

src/components/Common/PlayButton/index.tsx
lukepolo: Sounds good! I’ll go do that

server/api/themoviedb/interfaces.ts
lukepolo: I reverted this, no need for it.

sct:
Sorry for the wait @lukepolo I have been dumb busy. I will try to get to this soon! 🙇

sct:
Hi @lukepolo let's get this rebased and I will take another look.

I think the design will need some work but the logic seems sound otherwise. I can handle changing the design once we get it merged :)

lukepolo:
Rebased, confirmed it is still working on web/iOS chrome.

src/components/Common/PlayButton/index.tsx
```diff
@@ -1,50 +1,83 @@
...
 import ButtonWithDropdown from '../ButtonWithDropdown';
+import Trailer from '../../Trailer';
 
 interface PlayButtonProps {
   links: PlayButtonLink[];
 }
 
+export enum PlayButtonLinkTypes {
+  Web = 'web',
+  Trailer = 'trailer',
```
TheCatLady, line 12:
```suggestion
  WEB = 'web',
  TRAILER = 'trailer',
```

server/index.ts
TheCatLady: I also don't understand why this isn't done in the frontend. 🤔

lukepolo: I've tried this but it seems it has to be a new window for this to work. You could probably render this from a view but I ran into problems trying that as well (its been a few months so i can't recall all the particulars)

TheCatLady: I don't understand what the need for a new window has to do with this being handled in the frontend, though?  This doesn't belong in the API.

lukepolo: we can move this into a deeplink.html and serve that instead. But you don't want to load the entire app if you don't have to.

lukepolo: i found a way todo it via UI only , it may not work for all phones, i only tested on my iOS phone.

lukepolo:
forgot I was working on this, cant remember but I thought something wasn't working with the implementation. may look at it here  soon

PR #1995, opened 2021-08-15
feat(servarr): auto fill base url when testing service if missing

danshilm:
#### Description

This will fill the base URL of the *arr service only if it's missing and the base URL hasn't been
provided beforehand

N/A

#### To-Dos

#### Issues Fixed or Closed

- Fixes none


server/routes/settings/radarr.ts
```diff
@@ -57,6 +58,10 @@ radarrRoutes.post<
         path: folder.path,
       })),
       tags,
+      urlBase:
+        req.body.baseUrl && req.body.baseUrl !== ''
```
TheCatLady, line 62:
```suggestion
        req.body.baseUrl
```

server/routes/settings/sonarr.ts
```diff
@@ -55,6 +56,10 @@ sonarrRoutes.post('/test', async (req, res, next) => {
       })),
       languageProfiles,
       tags,
+      urlBase:
+        req.body.baseUrl && req.body.baseUrl !== ''
```
TheCatLady, line 60:
```suggestion
        req.body.baseUrl
```

src/components/Settings/RadarrModal/index.tsx
```diff
@@ -317,6 +318,9 @@ const RadarrModal: React.FC<RadarrModalProps> = ({
                     port: values.port,
                     useSsl: values.ssl,
                   });
+                  if (!values.baseUrl || values.baseUrl === '') {
```
TheCatLady, line 321:
```suggestion
                  if (!values.baseUrl) {
```

src/components/Settings/SonarrModal/index.tsx
```diff
@@ -348,6 +349,9 @@ const SonarrModal: React.FC<SonarrModalProps> = ({
                     port: values.port,
                     useSsl: values.ssl,
                   });
+                  if (!values.baseUrl || values.baseUrl === '') {
```
TheCatLady, line 352:
```suggestion
                  if (!values.baseUrl) {
```

TheCatLady (commented):
Empty string is a falsy value 😉

server/routes/settings/radarr.ts
danshilm: Thanks for pointing that out, I forgot about it.

server/api/servarr/base.ts
```diff
@@ -81,6 +110,22 @@ class ServarrBase<QueueItemAppendT> extends ExternalAPI {
     this.apiName = apiName;
   }
 
+  public getSystemStatus = async (): Promise<SystemStatus> => {
+    try {
+      const data = await this.getRolling<SystemStatus>(
```
sct, line 115: I think it's best we don't use the rolling cache here. Instead, just do a direct request every time so we always have the latest status information.

We use rolling in other places primarily because we don't want to slow down showing profiles/root folders every time we open the request modal.

danshilm: Noted 👍

PR #1996, opened 2021-08-16
fix(api): remove duplicate recommendation results

maxmonciardini:
#### Description
I was scrolling through Overseerr today when I clicked on the Recommendations tab and noticed that there were duplicate movie and TV show results (screenshots below). After investigating the issue, I discovered that the bug is not in fact caused by Overseerr, but instead a problem of the TMDB API itself. 

When requesting paginated results from the TMDB API, the last result of page 1 is repeated as the first result of page 2. I put some screenshots below so you can see what I'm talking about. 

To fix this issue, I removed the first result from every page after page 1. 

Duplicate movie
Duplicate movie in recommendations

Duplicate TV show
Duplicate tv show in recommendations

Page 1 of TMDB Recommendation API Request
Page 1 TMDB

Page 2 of TMDB Recommendation API Request
Page 2 TMDB

All fixed!
Fixed movie recommendations

Fixed tv recommendations

### DIFF
--- a/server/routes/movie.ts
+++ b/server/routes/movie.ts
@@ -39,6 +39,12 @@ movieRoutes.get('/:id/recommendations', async (req, res) => {
     language: req.locale ?? (req.query.language as string),
   });
 
+  // The TMDB API returns the last result of page 1 as the first result of page 2 and so on.
+  // We need to remove the first result of the results array to prevent duplicates.
+  if (results.page > 1) {
+    results.results.shift();
+  }
+
   const media = await Media.getRelatedMedia(
     results.results.map((result) => result.id)
   );
--- a/server/routes/movie.ts
+++ b/server/routes/movie.ts
@@ -39,12 +39,6 @@ movieRoutes.get('/:id/recommendations', async (req, res) => {
     language: req.locale ?? (req.query.language as string),
   });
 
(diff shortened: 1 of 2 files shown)


TheCatLady:
Since it isn't an Overseerr issue, I don't think this is a good idea.  TMDb should fix the issue on their end.

sct:
It doesn't _always_ duplicate in this manner. Its a cache mismatch on TMDB's side. This PR would actually remove non-duplciates in most cases.

maxmonciardini:
> It doesn't _always_ duplicate in this manner. Its a cache mismatch on TMDB's side. This PR would actually remove non-duplciates in most cases.
Good to know! Maybe instead we could go about this by filtering the results on the client side and removing any movies/shows that share the same "id" key.

sct:
> > It doesn't _always_ duplicate in this manner. Its a cache mismatch on TMDB's side. This PR would actually remove non-duplciates in most cases.
> 
> Good to know! Maybe instead we could go about this by filtering the results on the client side and removing any movies/shows that share the same "id" key.

That would mean the client would needs to store each item it's already displayed and then iterate through that list to check each title for duplicates. I fear this will mean a performance hit for large scrolling lists. This is something we just aren't going to bother fixing, and leave it up to TMDb to fix sometime.

PR #2051, opened 2021-09-03
Add healthcheck to dockerfile

modem7:
#### Description

Add healthcheck to dockerfile

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,9 @@ RUN apk add --no-cache tzdata tini
 # copy from build image
 COPY --from=BUILD_IMAGE /app ./
 
+HEALTHCHECK --interval=30s --timeout=10s --retries=3 --start-period=10s \
+    CMD wget http://localhost:5055/api/v1/status -qO /dev/null || exit 1
+
 ENTRYPOINT [ "/sbin/tini", "--" ]
 CMD [ "yarn", "start" ]
 


TheCatLady:
I'm not sure how I feel about this..  Healthchecks can result in a lot of syslog spam, and so I kinda think that this should be opt-in (define the check yourself in your `docker run` command or Compose file) rather than opt-out (defining the check in the `Dockerfile` and forcing users to disable the checks).

modem7:
> I'm not sure how I feel about this..  Healthchecks can result in a lot of syslog spam, and so I kinda think that this should be opt-in (define the check yourself in your `docker run` command or Compose file) rather than opt-out (defining the check in the `Dockerfile` and forcing users to disable the checks).

In this particular case, wget is set to quiet and sends any logs to dev/null. There shouldn't be any spam.

TheCatLady:
Even if we disable logging from the healthcheck, there's also moby/moby#42200.

TheCatLady:
Sorry @modem7, discussed this with @sct and we aren't going to be merging this.  Anyone who wants the healthcheck should define it themselves.

Krojack:
Curl is the most commonly used command to use for healchecks via compose file and it's not even included with the image.

Salvora:
image has wget in it so, this works for a good healthcheck in compose

```yaml
    healthcheck:
      test:
        ["CMD", "wget", "--spider", "-q", "http://localhost:5055/api/v1/status"]
      start_period: 10s
      interval: 30s
      timeout: 20s
      retries: 3
```

modem7:
> image has wget in it so, this works for a good healthcheck in compose
> 
> ```yaml
>     healthcheck:
>       test:
>         ["CMD", "wget", "--spider", "-q", "http://localhost:5055/api/v1/status"]
>       start_period: 10s
>       interval: 30s
>       timeout: 20s
>       retries: 3
> ```

That's similar to what I've put in the PR:

```dockerfile
HEALTHCHECK --interval=30s --timeout=10s --retries=3 --start-period=10s \
    CMD wget http://localhost:5055/api/v1/status -qO /dev/null || exit 1
```

And for my compose:

```yaml
    healthcheck:
      test: wget http://localhost:5055/api/v1/status -qO /dev/null || exit 1
      interval: 30s
      retries: 3
```

PR #2181, opened 2021-10-09
fix(frontend): better request/media cards for items without valid TMDb IDs

### LINKED ISSUE #2323: Provide more information when 'title' is no longer available
### Description

Without more information, it's hard to search for a new title that matches the original request.

### Version

1.27

### Steps to Reproduce

Not sure.

### Screenshots

image

### Logs

_No response_

### Platform

desktop

### Device

N/A

### Operating System

N/A

### Browser

Chrome

### Additional Context

_No response_

### Code of Conduct

### LINKED ISSUE #2898: Empty entry in Recently Added and this is not dissapearing after a restart of the software
### Description

There is an empty Entry in there and I suspect something more is not working correctly.

### Version

Docker V1.29.1

### Steps to Reproduce

No idea just use the software and suddenly this appeared.

### Screenshots

chrome_2022-08-11_08-54-58

### Logs

```shell
I could provide logs but only directly to the developer, this also applies to Database files.
I don't want any information to be publicly accesible
```

### Platform

desktop

### Device

computer, samsung galaxy s21 ultra, etc 

### Operating System

all does not matter

### Browser

Chrome, Firefox

### Additional Context

_No response_

### Code of Conduct

TheCatLady:
#### Description

Currently, if a request's TMDb is not valid, we display a vague message that doesn't give any clues as to what the media item was.

This PR adds additional details, including the request type (movie or series), the TMDb/TVDB IDs, the requester, and the request date.  This should help admins identify the media that is associated with "broken" requests (even if these IDs are no longer valid, they can be useful in matching broken requests to items added to Radarr/Sonarr).

Additionally, we currently display a placeholder title card for available media without valid TMDb IDs. This PR also adds an error card that allows for deletion of these items.

#### To-Dos

- [x] Cypress tests

#### Issues Fixed or Closed

- Closes #2323
- Fixes #2898


samwiseg0:
Can we add the title to this please? The ID is not helpful since it could have been deleted and there is no go way to know what it was at that point.

samwiseg0:
> > Can we add the title to this please? The ID is not helpful since it could have been deleted and there is no good way to know what it was at that point.
> 
> As discussed previously, there's no way to do that currently. The IDs can help locate the items in the *arrs if the requests have already been approved.

Why cant we cache the title when it is requested? We can then only use that title if the ID becomes invalid. I think knowing what the user intended to request is important. We did discuss it but you were more concerned about the translations. Which I think having a way to know what has been orphaned is more important? 

Like you stated only if it was approved would the other ID be useful. I have had many instances where that is not the case since I had not approved the request.

TheCatLady:
> Why can't we cache the title when it is requested? We can then only use that title if the ID becomes invalid. I think knowing what the user intended to request is important. We did discuss it but you were more concerned about the translations. Which I think having a way to know what has been orphaned is more important?
> 
> Like you stated only if it was approved would the other ID be useful. I have had many instances where that is not the case since I had not approved the request.

Either way, that's beyond the scope of this PR, and maybe something to consider for the future.  I'm not making DB changes here, just displaying the info we already have.

src/components/RequestCard/index.tsx
```diff
@@ -47,37 +50,123 @@ const RequestCardPlaceholder = () => {
...
+const RequestCardError = ({ requestData }: RequestCardErrorProps) => {
   const { hasPermission } = useUser();
   const intl = useIntl();
 
   const deleteRequest = async () => {
-    await axios.delete(`/api/v1/media/${mediaId}`);
-    mutate('/api/v1/request?filter=all&take=10&sort=modified&skip=0');
+    await axios.delete(`/api/v1/media/${requestData?.media.id}`);
+    mutate(recentlyAddedApi);
+    mutate(recentRequestsApi);
```
sct, line 63: I dont really like having TItleCard import from Discover. Its a strange cross dependency. Better to just write out the known strings here imo.

TheCatLady: Hm, okay! I'll revert that change :)

sct (approved):

PR #2188, opened 2021-10-12
feat(plex): selective user import

TheCatLady:
#### Description

Allows selection of users to import from Plex, instead of always importing all users with library access.  This is useful for admins who have disabled the `Enable New Plex Sign-In` setting.

image

#### To-Dos

#### Issues Fixed or Closed

N/A


server/routes/settings/index.ts
```diff
@@ -225,6 +225,50 @@ settingsRoutes.post('/plex/sync', (req, res) => {
...
+
+    await Promise.all(
+      plexUsers.map(async (user) => {
+        if (
+          !(await qb
+            .where('user.plexId = :id', { id: user.id })
+            .orWhere('user.email = :email', {
+              email: user.email.toLowerCase(),
+            })
+            .getOne()) &&
```
sct, line 260: I wonder if we can check all the users at once with an `IN` instead of running a ton of queries like this?

sct (approved):
LGTM!

PR #2190, opened 2021-10-12
fix(ui): hide 'Recently Added' & 'Recent Requests' sliders when empty

TheCatLady:
#### Description

If the `Recently Added` and/or `Recent Requests` sliders are empty, hide them instead of taking up valuable screen real estate to display nothing. 😉

Also fixes a bug where users cannot click the `Recent Requests` slider heading in their profiles to view their own requests unless they have manage/view request perms.

N/A

#### To-Dos

#### Issues Fixed or Closed

N/A


sct:
After a very small amount of use, these sliders are very rarely hidden so I am not sure what this really offers. I'd like users to know the sliders exist even when they are empty. It lets them know that the home page will have those items as soon as they start interacting with the app.

TheCatLady:
> Yeah not sure I agree. I think this is still a core aspect of Overseerr and the cases in where these will remain empty is exceedingly rare. Also, a user may not have the requests permission but they may have had it at some point. So there is no saying whether or not they have requests at the moment.

Yeah, this is definitely is not a super common case.  But the way I'm thinking of it is, Overseerr's Discover page is a feed of sorts... and it's natural for things to appear in feeds only when they are applicable/relevant.

As for the case of the user without the request perms... yes, that is true.  However, we would still display the slider if they had requests to view, and only hide it if the slider isn't relevant.

I also personally have a couple of new users who like to browse but haven't requested anything (not because they don't have request perms, but because they are overly polite/courteous), and the empty requests slider occupies a nontrivial chunk of screen real estate near the top of the page unnecessarily.  They already know they haven't submitted any requests yet; there's no need to remind them every time they open Overseerr 😅

src/components/Discover/index.tsx
```diff
@@ -58,52 +58,56 @@ const Discover = () => {
   return (
     <>
       <PageTitle title={intl.formatMessage(messages.discover)} />
-      {hasPermission([Permission.MANAGE_REQUESTS, Permission.RECENT_VIEW], {
-        type: 'or',
-      }) && (
+      {(!media || !!media.results.length) &&
+        !mediaError &&
```
sct, line 62: I haven't tested this locally yet, but I imagine we will get pop-in for these sliders since we are not rendering them at all until the media comes back from the request. The alternative is having them... flash away. I don't think either solution will look very good 🤔

TheCatLady: Actually, it's the same logic as for the watchlist slider -- it flashes away . But I think that's more okay, since it is less likely for these sliders to actually be empty?

sct: That's true that the watchlist flashes away as well. I don't like how that one behaves either, haha. Not sure if there is a great compromise here 🤔

src/components/Slider/index.tsx
```diff
@@ -11,7 +11,7 @@ interface SliderProps {
   items?: JSX.Element[];
   isLoading: boolean;
   isEmpty?: boolean;
-  emptyMessage?: string;
+  emptyMessage?: string | React.ReactNode;
```
sct, line 14: ReactNode also takes a string, so can just make this accept only React.ReactNode
```suggestion
  emptyMessage?: React.ReactNode;
```

sct (approved):

PR #2242, opened 2021-10-28
feat(notif): issue notifications

TheCatLady:
#### Description

Implements issue notifications.

Also makes the following changes:
- The "author" field of Discord notifications is now set to a more useful event description string, rather than the application title; and the bot name now defaults to the application title unless explicitly set to something else
- Similarly uses the "context" field of Slack notifications to display the event description string
- Unused `media_imdb` field now removed from LunaSea & webhook payloads
- Reformatted Pushover, Pushbullet, and Telegram notifications (which do not support fields the way Discord/Slack do) to use less vertical space
- Refactored code to reduce duplicated logic

#### To-Dos

#### Issues Fixed or Closed

N/A


server/entity/MediaRequest.ts
```diff
@@ -151,14 +152,15 @@ export class MediaRequest {
             omission: '…',
           }),
           image: `https://image.tmdb.org/t/p/w600_and_h900_bestv2${movie.poster_path}`,
-          media,
```
sct, line 154: why are we no longer attaching the media object here?

server/lib/notifications/agents/discord.ts
```diff
@@ -143,104 +181,57 @@ class DiscordAgent
...
+        case Notification.ISSUE_REOPENED:
+          color = EmbedColors.RED;
+          break;
+        case Notification.ISSUE_COMMENT:
+          color = EmbedColors.ORANGE;
+          break;
+        case Notification.ISSUE_RESOLVED:
+          color = EmbedColors.GREEN;
+          break;
       }
```
sct, line 195: Why two different switch cases for this? You can combine this with the other switch case that sets colors?

TheCatLady: I am handling request notifications and issue notifications separately.

server/entity/MediaRequest.ts
TheCatLady: Since we're passing the request object, it isn't really necessary.

TheCatLady: Actually, nevermind, that doesn't work, adding the media objects back lol.

sct (approved):
Didn't test that every notification type is... perfect here? But it looks good! Don't see any major issues. Let's get this merged and see what pops up in dev!

PR #2288, opened 2021-11-20
Add 20th Century Studios to Studio Slider

TheMeanCanEHdian:
#### Description
Adds 20th Century Studios to the Studios Slider. 

With Disney's acquisition of Fox, new movies are being published under 20th Century Studios. I wasn't sure if 20th Century Fox should be outright replaced since it has a huge catalog of movies under it, so I opted to leave it.

I have also updated the icon on TMDB as the existing one didn't have transparency and looked bad when used in Overseerr. TMDB API is now returning this image for the studio as well.

No new text has been added to the UI.

Slider
image

Studio Discover
image

#### To-Dos


TheCatLady:
I think that if we want to add `20th Century Studios`, then it should replace `20th Century Fox`.  Looks odd to have both there in the slider, and if there is not going to be any new content from the latter then we should remove it.

@sct, thoughts?

sct:
> I think that if we want to add `20th Century Studios`, then it should replace `20th Century Fox`. Looks odd to have both there in the slider, and if there is not going to be any new content from the latter then we should remove it.
> 
> @sct, thoughts?

Not sure what to do in this situation. Unless TMDb migrated all of the titles to the new studio, there is still a ton of value in the old one. Maybe we can move it to the end of the list?

TheCatLady:
> Not sure what to do in this situation. Unless TMDb migrated all of the titles to the new studio, there is still a ton of value in the old one. Maybe we can move it to the end of the list?

Yeah, but there wouldn't be any new content to discover, so I don't think it needs to be in the slider?  I doubt TMDb would merge the two, unfortunately.

sct:
> > Not sure what to do in this situation. Unless TMDb migrated all of the titles to the new studio, there is still a ton of value in the old one. Maybe we can move it to the end of the list?
> 
> 
> 
> Yeah, but there wouldn't be any new content to discover, so I don't think it needs to be in the slider?  I doubt TMDb would merge the two, unfortunately.

There may not be any new content to discover but there may be content people haven't discovered in there yet. Lots of fantastic movies from that studio. Not really sure what to do here. Maybe just leaving both is best.

JonnyWong16:
Keep both but move `20th Century Fox` to the end of the list since it won't have any _new_ content.

TheMeanCanEHdian:
I see this was marked as stale. Is there anything else that I can do to have it included?

TheCatLady:
> I see this was marked as stale. Is there anything else that I can do to have it included?

I'm still of the opinion that 20th Century Fox should be replaced with 20th Century Studios. 🤔  The 20th Century Fox list is now "stale" and won't be updated further, meaning that over time there will be less and less content to discover from that list.  And, since the acquisition was 3 years ago, it's likely that many server owners already have most media they want from the studio.

In general, I don't think any of the sliders should display results that will no longer change / be updated.

TheMeanCanEHdian:
@sct Are you thinking you still want the 20th Century Fox removed or left?

TheMeanCanEHdian:
Any additional thoughts here? Is the consensus to remove 20th Century Fox or leave it?

sct:
Sorry for making you wait forever @TheMeanCanEHdian

Let's remove it and just keep the new one in its place!

TheMeanCanEHdian:
Removed 20th Century Fox studio slider.

sct (approved):
Thank you! Let's get this merged!

sct:
@all-contributors please add @TheMeanCanEHdian for code

PR #2294, opened 2021-11-21
fix(servarr): handle baseurl error when testing connection

danshilm:
#### Description

When testing the connection to a Radarr/Sonarr server using a base URL, if the base URL cannot be fetched, the test fails even when the base URL input by the user is valid in an attempt to auto fill the base URL for the user.

This PR makes it so that the base URL is only fetched from Radarr/Sonarr if the user has not input one. Errors when fetching the base URL from Radarr/Sonarr are also handled.

N/A

#### To-Dos

#### Issues Fixed or Closed

- Fixes none

### DIFF
--- a/server/routes/settings/radarr.ts
+++ b/server/routes/settings/radarr.ts
@@ -46,13 +46,10 @@ radarrRoutes.post<
       url: RadarrAPI.buildUrl(req.body, '/api/v3'),
     });
 
-    const urlBase =
-      req.body.baseUrl && req.body.baseUrl !== '/'
-        ? req.body.baseUrl
-        : await radarr
-            .getSystemStatus()
-            .then((value) => value.urlBase)
-            .catch(() => undefined);
+    const urlBase = await radarr
+      .getSystemStatus()
+      .then((value) => value.urlBase)
+      .catch(() => req.body.baseUrl);
     const profiles = await radarr.getProfiles();
     const folders = await radarr.getRootFolders();
     const tags = await radarr.getTags();
--- a/server/routes/settings/radarr.ts
+++ b/server/routes/settings/radarr.ts
@@ -46,7 +46,13 @@ radarrRoutes.post<
       url: RadarrAPI.buildUrl(req.body, '/api/v3'),
     });
 
-    const { urlBase } = await radarr.getSystemStatus();
+    const urlBase =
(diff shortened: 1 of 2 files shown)


server/routes/settings/radarr.ts
```diff
@@ -46,7 +46,13 @@ radarrRoutes.post<
       url: RadarrAPI.buildUrl(req.body, '/api/v3'),
     });
 
-    const { urlBase } = await radarr.getSystemStatus();
+    const urlBase =
+      req.body.baseUrl && req.body.baseUrl !== '/'
+        ? req.body.baseUrl
+        : await radarr
+            .getSystemStatus()
```
TheCatLady, line 53: Why not always fetch the base URL, and just fall back to the user-inputted value if an exception is thrown?

Also, any reason why we don't want to apply this same logic to Sonarr?

server/routes/settings/radarr.ts
```diff
@@ -46,7 +46,13 @@ radarrRoutes.post<
...
+            .then((value) => value.urlBase)
+            .catch(() => undefined);
```
TheCatLady, line 55: Why not always fetch the base URL, and just fall back to the user-inputted value if an exception is thrown?

Also, any reason why we don't want to apply this same logic to Sonarr?

danshilm: I fetch the base URL only if there's no base URL that the user provided already to be slightly faster.

As for Sonarr, I just missed it when I was making this fix lol 😅 I'll add that

TheCatLady: What if the user inputted an incorrect base URL?  Would be nice to try to fetch the correct one.  I don't imagine users are editing their *arr servers all the time, so I don't see the harm in always attempting the fetch.

TheCatLady: We could do something like this?
```suggestion
    const { urlBase } = await radarr
      .getSystemStatus()
      .then((value) => value.urlBase)
      .catch(() => req.body.baseUrl);
```

danshilm: That's good 😄 
I just don't personally like automatically fixing data that a user has already input. But here it should be no problem

sct (approved):
LGTM!

PR #2303, opened 2021-11-25
fix(logs): handle unexpected log messages

danshilm:
#### Description

This mostly changes the regexp that's used to parse logs messages for the log viewer.
It's now stricter and more specific. The log file is split right at the newline followed by the timestamp instead of just the newline since unexpected log messages like errors from Typeorm span more than just one line like:
```
2021-11-21T17:47:24.460Z [error]: Could not find any entity of type "MediaRequest" matching: {
    "where": {
        "id": 1391
    },
    "relations": [
        "requestedBy",
        "modifiedBy"
    ]
}
```
The better regexp also doesn't break when encountering titles like below which would incorrectly match the regexp for the JSON structured data in the log message:
```
2021-11-21T17:57:08.979Z [debug][Plex Scan]: Updating existing title: Lord El-Melloi II's Case Files {Rail Zeppelin} Grace note
```

N/A

#### To-Dos
N/A

#### Issues Fixed or Closed

- Fixes none

### DIFF
--- a/server/interfaces/api/settingsInterfaces.ts
+++ b/server/interfaces/api/settingsInterfaces.ts
@@ -3,7 +3,7 @@ import type { PaginatedResponse } from './common';
 export type LogMessage = {
   timestamp: string;
   level: string;
-  label: string;
+  label?: string;
   message: string;
   data?: Record<string, unknown>;
 };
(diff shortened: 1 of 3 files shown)


sct (approved):
🙏 Thanks

PR #2345, opened 2021-12-13
feat(frontend): add Discovery+ to network slider

dsgnr:
#### Description

Adds Discovery+ to the Network slider. I took a best guess on where it should go, so if there's a particular preference, please let me know.

Screenshot 2021-12-13 at 10 10 15 am
Screenshot 2021-12-13 at 10 10 21 am


src/components/Discover/NetworkSlider/index.tsx
```diff
@@ -50,6 +50,12 @@ const networks: Network[] = [
       'https://image.tmdb.org/t/p/w780_filter(duotone,ffffff,bababa)/tuomPhY2UtuPTqqFnKMVHvSb724.png',
     url: '/discover/tv/network/49',
   },
+  {
+    name: 'Discovery+',
+    image:
+      'https://www.themoviedb.org/t/p/h100_filter(negate,000,666)/1D1bS3Dyw4ScYnFWTlBOvJXC3nb.png',
```
danshilm, line 56: 2 things here:
- I'm not sure what the difference between using `https://www.themoviedb.org` instead of the usual `https://image.tmdb.org` domain is (I suspect it might have something to do with their CDN), but since the latter is what's used in the docs, we should use that instead.
- The filter you use makes the logo look a bit grey-ish compared to the others. Let's make it consistent with the other logos using the same filter as the others.
```suggestion
      'https://image.tmdb.org/t/p/h100_filter(duotone,ffffff,bababa)/1D1bS3Dyw4ScYnFWTlBOvJXC3nb.png',
```

sct (approved):
Great. Thanks!

PR #2367, opened 2021-12-25
feat(ui): add trakt external link

danshilm:
#### Description
This adds a logo that links to the media's Trakt page next to the current TMDb logo & link. 
[Link](https://trakt.docs.apiary.io/reference/search/id-lookup) to Trakt's documentation.

Screenshot_101

#### To-Dos

#### Issues Fixed or Closed

- Fixes none


sct:
I think we are already near the limit here in terms of icons.

image

What happens when this overflows?

TheCatLady:
Probably need to make some adjustments to the size/spacing of the icons to ensure they fit.  Also, I don't know about the ordering.  I feel like Trakt probably belongs at the end, rather than towards the beginning?

danshilm:
> I think we are already near the limit here in terms of icons.
> 
> image
> 

> What happens when this overflows?

The icons get smaller and keep the same horizontal spacing. Like so:
Screen Shot 2021-12-26 at 11 25 00 AM

If we do add a 7th item, the icons will get too small to be legible though.


> Probably need to make some adjustments to the size/spacing of the icons to ensure they fit. Also, I don't know about the ordering. I feel like Trakt probably belongs at the end, rather than towards the beginning?

How about between IMDB and the Rotten Tomatoes icons?

louisgrasset:
What about making a simple flex wrap ?

`<div class="flex flew-wrap items-center justify-center w-full space-x-5 space-y-5">...</div>`

image

danshilm:
> What about making a simple flex wrap ?

I hope we don't ever have to expand beyond one line of icons because I don't think there's a way we can make it look as good as it looks right now. For now, 6 icons on one line look okay; the text barely smaller than with 5 icons, but a 7th might make things look small on the smaller phone screens.

sct (approved):

PR #2401, opened 2022-01-08
feat: show alert/prompt when settings changes require restart

TheCatLady:
#### Description

Also removes the `MANAGE_SETTINGS` permission, since there aren't really any settings that should be available to non-admin users.  (Also, some settings require a server restart to apply changes.)

image
image

#### To-Dos

#### Issues Fixed or Closed

N/A


src/components/StatusChecker/index.tsx
```diff
@@ -0,0 +1,78 @@
+import { RefreshIcon, SparklesIcon } from '@heroicons/react/outline';
```
sct, line 1: Can perform a git mv operation on this to keep the history.

sct:
I don't think we should completely block the app in this case. If someone makes a breaking change by mistake then realizes right after saving, they can't fix it. Maybe a popup with a warning that they can close instead? Something that may appear on every page navigation. That way, if it's a mistake, they can revert the change?

TheCatLady:
> I don't think we should completely block the app in this case. If someone makes a breaking change by mistake then realizes right after saving, they can't fix it. Maybe a popup with a warning that they can close instead? Something that may appear on every page navigation. That way, if it's a mistake, they can revert the change?

Good call, will add a close button to the pop-up!

sct (approved):

PR #2413, opened 2022-01-12
fix(notif): show event in pop up notification for slack

### LINKED ISSUE #2408: Slack Popup Notifications 'Content Cannot be Displayed' - missing `text` property
### Description

When Slack messages are sent, users will get a popup notification. For Overseerr, these notifications arrive with the message 'Content Cannot be Displayed'

Plextrobe

The reason for this is usually when the Slack payload is created using 'Blocks' rather than a plain text message. When using Blocks, you still need to start with a text attribute. This is the attribute that Slack will use for the popup message - but doesn't get used in the main message.
Annoyingly, Slack's Blocks Builder utility won't show you this attribute (as it's not technically part of the blocks, but part of the broader payload)

eg...

```
                payload = {
                            "blocks": [
                           ...
                            ]
                        }
```

Will send a message to slack, but will generate the same 'Content Cannot be Displayed' message in the popup.

Adding the `text` attribute before the `blocks` will display the message in the popup

```
                payload = {
                            "text": "An Important Message from Overseerr!",
                            "blocks": [
                              ...
                            ]
                        }
```
Plextrobe

### Version

1.28.0

### Steps to Reproduce

1) Create a Slack notification to post to Slack when requested content is available

2) Request something

3) Make it available

4) Check the pop-up message (desktop or mobile) for Slack

### Screenshots

_No response_

### Logs

_No response_

### Platform

desktop

### Device

MacBook Pro

### Operating System

Montery

### Browser

Vivaldi

### Additional Context

_No response_

### Code of Conduct

danshilm:
#### Description

The `text` attribute being absent resulted in the pop up notification saying "That content can't be displayed".
This sets the text to be the name of event.

Example:
Screen Shot 2022-01-12 at 6 25 01 PM

#### To-Dos

#### Issues Fixed or Closed

- Fixes #2408


server/lib/notifications/agents/slack.ts
```diff
@@ -39,6 +39,7 @@ interface EmbedBlock {
 }
 
 interface SlackBlockEmbed {
+  text?: string;
```
TheCatLady, line 42:
```suggestion
  text: string;
```
Since we always want to include `text` to avoid this issue in the future.

danshilm: Gotcha

TheCatLady (approved):
👍🏻

sct (approved):

PR #2446, opened 2022-01-20
fix(tautulli): fetch additional user history as necessary to return 20 unique media

TheCatLady:
#### Description

Currently, the user "recently watched" results are the result of deduping a single Tautulli `get_history` API call.  This PR changes the behavior to fetch additional results until there are 20 unique media, in case a user watches a lot of TV shows (especially long-running series with many episodes).

N/A

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/server/api/tautulli.ts
+++ b/server/api/tautulli.ts
@@ -251,6 +251,7 @@ class TautulliAPI {
               order_column: 'date',
               order_dir: 'desc',
               user_id: user.plexId,
+              media_type: 'movie,episode',
               length: take,
               start,
             },
--- a/server/api/tautulli.ts
+++ b/server/api/tautulli.ts
@@ -232,7 +232,7 @@ class TautulliAPI {
   public async getUserWatchHistory(
     user: User
   ): Promise<TautulliHistoryRecord[]> {
-    let records: TautulliHistoryRecord[] = [];
+    let results: TautulliHistoryRecord[] = [];
 
     try {
       if (!user.plexId) {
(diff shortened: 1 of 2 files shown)


server/api/tautulli.ts
```diff
@@ -192,23 +193,47 @@ class TautulliAPI {
...
-            order_dir: 'desc',
-            user_id: user.plexId,
-            length: 100,
-          },
-        })
-      ).data.response.data.data;
+      const take = 100;
+      let start = 0;
+
+      while (records.length < 20) {
```
sct, line 206: What happens here when there are not at least 20 records in the first place?

TheCatLady: It will stop after the first API call that doesn't return any records (see the `if` clause below).

sct: Oh, I see. Missed that. 👍

server/routes/user/index.ts
```diff
@@ -534,26 +533,17 @@ router.get<{ id: string }, UserWatchDataResponse>(
...
+          watchHistory.map(
+            async (record) =>
+              await mediaRepository.findOne({
+                where: {
+                  ratingKey:
+                    record.media_type === 'movie'
+                      ? record.rating_key
+                      : record.grandparent_rating_key,
+                },
+              })
```
sct, line 545: Just realized how this is being done. We should refactor this as well to use an IN so we can query all the media at once?

TheCatLady: Oh yup, we can do that now!  Will fix.

danshilm (approved):

PR #2458, opened 2022-01-24
feat: verify Plex server access during auth for existing users with Plex IDs

TheCatLady:
#### Description

Currently, existing users are always able to authenticate and sign in to Overseerr, even if their Plex server access has been revoked.

This PR modifies the behavior so that users no longer able to access the Plex server can no longer sign in.

Also, redacts additional irrelevant Plex user account information from logs (subscription, entitlements) and improves/adds logging.

N/A

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/docs/using-overseerr/settings/README.md
+++ b/docs/using-overseerr/settings/README.md
@@ -66,9 +66,14 @@ This setting is **enabled** by default.
 
 ### Enable Local Sign-In
 
-When enabled, users who have configured passwords will be allowed to sign in using their email address.
+When enabled:
 
-When disabled, Plex OAuth becomes the only sign-in option, and any "local users" you have created will not be able to sign in to Overseerr.
+- Users who have configured passwords will be allowed to sign in using their email address.
+
+When disabled:
+
+- Plex OAuth becomes the only sign-in option, and any "local users" you have created will no longer be able to sign in to Overseerr.
+- User access to the Plex server is verified upon each sign-in attempt.
 
 This setting is **enabled** by default.
 
--- a/docs/using-overseerr/settings/README.md
+++ b/docs/using-overseerr/settings/README.md
@@ -66,14 +66,9 @@ This setting is **enabled** by default.
 
 ### Enable Local Sign-In
 
-When enabled:
+When enabled, users who have configured passwords will be allowed to sign in using their email address.
 
-- Users who have configured passwords will be allowed to sign in using their email address.
-
-When disabled:
-
-- Plex OAuth becomes the only sign-in option, and any "local users" you have created will no longer be able to sign in to Overseerr.
-- User access to the Plex server is verified upon each sign-in attempt.
+When disabled, Plex OAuth becomes the only sign-in option, and any "local users" you have created will not be able to sign in to Overseerr.
 
 This setting is **enabled** by default.
 
--- a/server/lib/settings.ts
+++ b/server/lib/settings.ts
@@ -286,7 +286,7 @@ class Settings {
           tv: {},
         },
         hideAvailable: false,
-        localLogin: false,
+        localLogin: true,
         newPlexLogin: true,
         region: '',
         originalLanguage: '',
--- a/server/lib/settings.ts
+++ b/server/lib/settings.ts
@@ -286,7 +286,7 @@ class Settings {
           tv: {},
         },
         hideAvailable: false,
-        localLogin: true,
+        localLogin: false,
         newPlexLogin: true,
         region: '',
         originalLanguage: '',
--- a/server/routes/auth.ts
+++ b/server/routes/auth.ts
@@ -48,98 +48,119 @@ authRoutes.post('/plex', async (req, res, next) => {
       })
       .getOne();
 
-    if (user) {
-      // Let's check if their Plex token is up-to-date
-      if (user.plexToken !== body.authToken) {
-        user.plexToken = body.authToken;
-      }
-
-      // Update the user's avatar with their Plex thumbnail, in case it changed
-      user.avatar = account.thumb;
-      user.email = account.email;
-      user.plexUsername = account.username;
-
-      // In case the user was previously a local account
-      if (user.userType === UserType.LOCAL) {
-        user.userType = UserType.PLEX;
-        user.plexId = account.id;
-      }
+    if (!user && !(await userRepository.count())) {
+      user = new User({
+        email: account.email,
+        plexUsername: account.username,
+        plexId: account.id,
+        plexToken: account.authToken,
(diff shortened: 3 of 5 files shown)


samwiseg0:
I believe the behavior most people would expect is that once a user is removed from Plex they also will not have access to Overseerr. Unfortunately, this change would not help most people as local login is enabled by default. 

Could we add additional checks to see if a user has set a password that would allow them to log in locally? Meaning if the plex user/admin _has not_ set a password for the user also disable login from that account.

samwiseg0 (approved):

danshilm (approved):

danshilm:
Also need to set `localLogin` to false in `getInitialProps` in the `_app.tsx` file.

JonnyWong16:
Is there a _technical limitation_ that is preventing you from rejecting local logins with they are unshared from the Plex server? Can you just compare the Plex user ID from the local login against the server?

JonnyWong16:
I strongly disagree with that approach. The type of users should not be grouped by the login method: Plex OAuth user vs local login user. From a **security perspective** this does not make sense. Users should be grouped based on their level of access: has a Plex user ID vs does not have a Plex user ID and the login method is irrelevant. Any user with a Plex user ID should have their access checked against the Plex server. A rejected user should not be able to _bypass security_ by using a different login method.

> Plex IDs linked during login/import, so a local user could end up rescinding their own access to Overseerr

This also doesn't make sense. The Plex ID link should be done when they submit the login, but _before_ they are allowed through. The user might have to wait an extra second while the server works on linking, but it makes a lot more sense this way. There should not be any scenario where the user is allowed through _first_, _and then_ rescinding themselves.

For local accounts without a Plex ID (different email than the Plex account), that's on the admin for providing an account without permission checks. They provided a login that can bypass their own Plex server security.

TheCatLady:
@JonnyWong16 There is no concept of two account types.  All users are the same type, and they either have a password or not, or they have a Plex ID or not.

JonnyWong16:
That agrees with everything I said.

> and they either have a password or not

Don't care about this, irrelevant.

> or they have a Plex ID or not

Yes, care about this, and check access based on this.

TheCatLady:
> > or they have a Plex ID or not
> 
> Yes, care about this, and check access based on this.

Isn't this the issue?  If the user hasn't signed in with a Plex account previously (and wasn't imported from Plex), they might not have a Plex ID, and thus will be granted access when authing with a password.  If the Plex ID gets added later (automatically, without any action from them), they will suddenly lose access to Overseerr.

JonnyWong16:
There are only 4 scenarios.

---
### Scenario 1

1. User enters email/password and clicks login.
2. Server sees user without Plex ID.
3. Server retrieves list of Plex users.
    i. User email is on the list of Plex users.
    ii. Server links user to Plex ID.
4. User is allowed to login.

The account is now linked so subsequent logins will be scenario 3 or 4.

---
### Scenario 2

1. User enters email/password and clicks login.
2. Server sees user without Plex ID.
3. Server retrieves list of Plex users.
    i. User email is _not_ on the list of Plex users.
4. User is allowed to login.

This is the scenario where the admin gave the user an account with access.
* If this email address has a Plex account which is later shared on the Plex server, then the next time they login it is scenario 1.
* If this email address has a Plex account which is later shared on the Plex server, but removed again before they login to Overseerr, then they next time they login it's just scenario 2 again.

---
### Scenario 3

1. User enters email/password and clicks login.
2. Server sees user with a Plex ID.
3. Server retrieves list of Plex users.
    i. Plex ID is on the list of Plex users.
4. User is allowed to login.

---
### Scenario 4

1. User enters email/password and clicks login.
2. Server sees user with a Plex ID.
3. Server retrieves list of Plex users.
    i. Plex ID is _not_ on the list of Plex users.
4. User is _not_ allowed to login.

TheCatLady:
@JonnyWong16 Hmm, I am inclined to agree (and just to be clear, I do agree with your general sentiment/view, I'm just still noodling on all the potential implications and/or support requests)...

But with Scenario 2, it is also possible the Plex server is shared, Plex ID is added outside of the login process (when importing Plex users, we try to add Plex IDs to _all_ existing users), and then Plex server access is revoked... in which case it would end up being Scenario 4 on the next login attempt.  This doesn't seem unreasonable to me, but I do very much dislike the indeterminate behavior of Scenario 2.  Perhaps it's not too much cause for concern, though, since it is certainly an edge case.

Also, out of scope for this PR (which only touches the login routes), but we should note that these changes would only affect users attempting to authenticate; it would not impact users who are already logged in until their sessions expire.  The best way to restrict access for these users is to remove their permissions and/or delete their Overseerr accounts.

JonnyWong16:
>  it is also possible the Plex server is shared, Plex ID is added outside of the login process (when importing Plex users, we try to add Plex IDs to all existing users), and then Plex server access is revoked... in which case it would end up being Scenario 4 on the next login attempt.

No, the only way the Plex ID get's linked when you import is if the user _currently has access to the Plex server_ so it would be scenario 3 after import.

Unless you import to link the Plex IDs, then unshare your Plex server before they login again, then it would be scenario 4. But in that case you unshared with them so they shouldn't have access anyways.

TheCatLady:
> Edit: Re-reading what you wrote, I think that's what you said, but worded differently.

Yeah, just talking about the case where the server was shared, then Plex ID added, and then unshared.  I think we're on the same page.  Will discuss with @sct about how he wants to handle this.

sct:
I have to agree with @JonnyWong16 as well. We can use the fact a Plex ID exists as all as the condition as to whether or not we check if they still have access.

The only scenario I can think of things breaking for people is if they decide to migrate to local accounts from Plex. So they remove their plex access but the plex ids will still be set on the account. This would probably only affect people migrating away from using Plex. I don't think that is a use-case we should build around though.

So with that said, let's adjust the PR to check for access if there is a Plex ID set.

JonnyWong16:
Does the Plex admin user bypass the server permission check (i.e. just check the Plex user ID == Overseerr admin)? There is the possibility the admin could lock themselves out if their Plex Media Server connection changes and Overseerr can't check against the server.

TheCatLady:
> Does the Plex admin user bypass the server permission check (i.e. just check the Plex user ID == Overseerr admin)? There is the possibility the admin could lock themselves out if their Plex Media Server connection (`machineIdentifier`) changes and Overseerr can't check against the server.

https://github.com/sct/overseerr/blob/f463b219e22ee4e570d87bd8f62381aa25722aae/server/routes/auth.ts#L73

The server owner isn't returned in their own friends list anyway 😉

PR #2461, opened 2022-01-24
feat(jobs): show current job frequency

### LINKED ISSUE #2711: broken UI on modify job schedule
### Description

The UI on job schedule is broken, as canceling out doesn't revert changes when re-entering the edit screen changes persist even upon cancel.

Also the field on "Plex Recently Added Scan" will quickly flip to display to a different setting for a split second after clicking cancel (changes don't persist though).

### Version

1.29.1

### Steps to Reproduce

Go to 'Settings > Jobs'
Click on 'Edit'
Change drop down schedule selection
Click 'Cancel'
Expect the drop down to not have changed.
Click "Edit" again and see the change appears to have been saved.

### Screenshots

_No response_

### Logs

_No response_

### Platform

desktop

### Device

-

### Operating System

-

### Browser

Chrome

### Additional Context

_No response_

### Code of Conduct

danshilm:
#### Description
This adds the current job frequency to the modal and also resets the dropdowns to their initial state when the modal is closed.

image

#### To-Dos

#### Issues Fixed or Closed

- Fixes #2711


TheCatLady:
I don't think this is great UX, forcing users to interpret a cron expression.  The existing dropdown should just display the current selection.

danshilm:
And what about the jobs whose schedules can't be shown in the dropdown (like the Radarr Scan whose default is at 4am everyday)? I think we can set jobs like those to a default that can be represented in the dropdown.

TheCatLady:
> And what about the jobs whose schedules can't be shown in the dropdown (like the Radarr Scan whose default is at 4am everyday)? I think we can set jobs like those to a default that can be represented in the dropdown.

In that case, it should reset to the original default setting.

danshilm:
Thinking about this, what do you think of:
- Showing a link to something like [crontab.guru](https://crontab.guru) with the cron schedule set already or
- Showing the human readable description of the cron schedule (maybe with the cron string itself alongside)

The reason I'm asking is because of the cron schedules that can't be represented using the dropdown of either hours or minutes; it's not a consistent experience and might confuse the user even more.

TheCatLady:
I think that users who edit their `settings.json` outside of the Overseerr web UI should understand that their changes may not be reflected accurately for something like this?

danshilm:
The defaults for the Plex Full Scan, Radarr Scan, Sonarr Scan and the Download Sync Reset jobs can't be represented in the dropdowns, let alone a cron schedule that a user would set in the `settings.json` file themselves.

TheCatLady:
> The defaults for the Plex Full Scan, Radarr Scan, Sonarr Scan and the Download Sync Reset jobs can't be represented in the dropdowns, let alone a cron schedule that a user would set in the `settings.json` file themselves.

Uh wait, they can't?

danshilm:
The full Plex scan and *arr scans are 30 mins after each other, starting from 3am daily. And the download sync is at 1am daily.

TheCatLady:
> The full Plex scan and *arr scans are 30 mins after each other, starting from 3am daily. And the download sync is at 1am daily.
> And using the drop-down, the only option for a daily schedule would be to set the job to happen at midnight.

Can't we build the offset into the setting configured here? Assign an offset to each job (doesn't need to be configurable) and generate the cron string from the dropdown selection and offset?

danshilm:
I think we should be able to. But I think that displaying the cron in a human readable string would be better than that.

PR #2481, opened 2022-01-27
wip: use forward proxy

### LINKED ISSUE #1907: Ability to configure an outbound web proxy server
#### Description

Ability to configure an outbound web proxy server

#### Desired Behavior

Be able to use the outbound proxy server to request image or metadata

#### Additional Context

image

danshilm:
#### Description

This adds the ability to use a forward proxy when making external API requests.
I am not familiar with forward proxies and their usage, so feedback would be appreciated.

This does _not_ add the ability to proxy images.

image

#### To-Dos

- [ ] Implement proxy for frontend (images mostly)

#### Issues Fixed or Closed

- Fixes #1907


src/components/Settings/SettingsMain.tsx
```diff
@@ -58,6 +59,19 @@ const messages = defineMessages({
...
   partialRequestsEnabled: 'Allow Partial Series Requests',
   locale: 'Display Language',
+  proxySettings: 'Proxy Settings',
+  proxySettingsDescription: 'Configure Overseerr to use a forward proxy',
+  proxyEnabled: 'Enable Proxy',
+  proxyUseSSL: 'Use SSL',
+  proxyHostname: 'Hostname or IP Address',
+  proxyPort: 'Port',
+  proxyAuthUser: 'Authenticating User',
+  proxyAuthPass: 'Authenticating Password',
```
TheCatLady, line 69:
```suggestion
  proxyAuthUser: 'Username',
  proxyAuthPass: 'Password',
```

src/components/Settings/SettingsMain.tsx
```diff
@@ -58,6 +59,19 @@ const messages = defineMessages({
...
+  proxyBypassLocalAddresses: 'Bypass Local Addresses',
+  proxyIgnoredAddresses: 'Ignored Addresses',
+  proxyIgnoredAddressesHint: 'Enter an address and press Enter or Tab…',
```
TheCatLady, line 72: I don't think this is necessary.

src/components/Settings/SettingsMain.tsx
```diff
@@ -85,6 +99,18 @@ const SettingsMain: React.FC = () => {
...
+      then: Yup.string().required(
+        intl.formatMessage(messages.proxyValidationHostname)
+      ),
+    }),
+    proxyPort: Yup.string().when('proxyEnabled', {
+      is: true,
+      then: Yup.string().required(
+        intl.formatMessage(messages.proxyValidationPort)
+      ),
+    }),
```
TheCatLady, line 113: Should be validated as number.  See https://github.com/sct/overseerr/blob/ca184728e912c94e40484a55edfab01cae696a1b/src/components/Settings/SettingsPlex.tsx#L162-L171 for an example.

src/components/Settings/SettingsMain.tsx
```diff
@@ -116,76 +142,96 @@ const SettingsMain: React.FC = () => {
...
+          partialRequestsEnabled: data?.partialRequestsEnabled,
+          trustProxy: data?.trustProxy,
+          proxyEnabled: data?.proxy.enabled,
+          proxyHostname: data?.proxy.options.hostname,
+          proxyPort: data?.proxy.options.port,
+          proxyUseSSL: data?.proxy.options.useSSL,
+          proxyAuthUser: data?.proxy.options.authUser,
+          proxyAuthPass: data?.proxy.options.authPass,
+          proxyBypassLocalAddresses: data?.proxy.options.bypassLocalAddresses,
+          proxyIgnoredAddresses: data?.proxy.options.ignoredAddresses ?? [],
```
TheCatLady, line 163: Similar to the Tautulli settings not being a part of Plex settings, I think this can have its own settings endpoint especially since you have it in its own section.  There's too much here.

server/api/externalapi.ts
```diff
@@ -97,6 +90,54 @@ class ExternalAPI {
...
   }
+
+  private createAxiosInstance(
+    baseUrl: string,
+    params: Record<string, unknown>,
+    options: ExternalAPIOptions
+  ) {
+    const proxySettings = getSettings().main.proxy;
+    const privateIPAddressRegexp = new RegExp(
+      /(^192\.168\.([0-9]|[0-9][0-9]|[0-2][0-5][0-5])\.([0-9]|[0-9][0-9]|[0-2][0-5][0-5])$)|(^172\.([1][6-9]|[2][0-9]|[3][0-1])\.([0-9]|[0-9][0-9]|[0-2][0-5][0-5])\.([0-9]|[0-9][0-9]|[0-2][0-5][0-5])$)|(^10\.([0-9]|[0-9][0-9]|[0-2][0-5][0-5])\.([0-9]|[0-9][0-9]|[0-2][0-5][0-5])\.([0-9]|[0-9][0-9]|[0-2][0-5][0-5])$)/
```
TheCatLady, line 101: This regex doesn't work.  For example, `192.168.249.1` doesn't match.  Also, what about IPv6?

TheCatLady:
> This adds the ability to use a forward proxy when making external API requests. I am not familiar with forward proxies and their usage, so feedback would be appreciated.
> 
> This does _not_ add the ability to proxy images.

I think it's clearer to say, this only proxies server-side requests.  This will not fix issues relating to client-side TMDb access. As such, I don't know how useful this would be on its own.  The client device would still need to utilize a separate proxy solution, in which case the user could just install that on both the server and client?

darki73:
Hey, just want to add something to this PR.
1. The server side implementation (as-is) will at least prevent this type of issues (which are literally rendering Overseerr as not usable)
image

2. With regards to actually adding support for proxy on the client + server side, it would be nice to have a proxy type selector (as presented in Sonarr / Radarr)
3. I would really much like to make a PR of my own, however, given the stuff happening in the world, and living in Russia, I am kinda afraid to pull this repo and use ```yarn install``` due to the fact, that there might be a dependency on the infamous package which replaces content of all files on the system with the ❤️ emoji.
4. While it is not possible to proxy images in the current implementation, what can be done is a Media Proxy service (like the one in the Sonarr) which caches images locally and returns them to the user (this will also help to speed up initial loads, as we usually load the same images once opening the main page). 

    a. Request title from the API
    b. Retrieve list of images
    c. Cache them for a day or two
    d. Serve cached images instead of remote ones (so there will be no need to query TMDB for each and every image)

So my main point is, even if we can't see the image of the item we are searching for, any implementation of proxy is still better than nothing.

But anyways, thanks for the project, and for this PR!

PR #2505, opened 2022-02-03
feat(logs): add search filter

### LINKED ISSUE #2290: Additional Filters on Logs
### Description

It'd be cool to filter the logs for specific things - like showing only when users log in.

### Desired Behavior

Additional log filters for key activities.

### Additional Context

_No response_

### Code of Conduct

danshilm:
#### Description

This adds a search box on the Logs page of the Settings that filters log messages. Log messages will be filtered by their `label`, the `message` or any of the extra information that's in the `data` attribute of a log message.
Why also the extra information that's in the `data` attribute? Because I can search for "10764" and the log message
```
2022-02-03T12:01:14.564Z [debug][Plex Scan]: GUIDs are cached. Skipping metadata request. {"mediaIds":{"imdbId":"tt0830515","tmdbId":10764,"tvdbId":400},"title":"Quantum of Solace"}
```
will be matched.

I'm not sure about my use of the `useDebouncedState` hook, but I do think I missed the mark and am not using it correctly here 👀

Desktop:
image
Mobile:
Screen Shot 2022-02-03 at 19 31 22

#### To-Dos

#### Issues Fixed or Closed

- Fixes #2290

### DIFF
--- a/server/routes/settings/index.ts
+++ b/server/routes/settings/index.ts
@@ -375,6 +375,25 @@ settingsRoutes.get(
       'data',
     ];
 
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const deepValues = (obj: Record<string, any>): any[] => {
+      const values = [];
+
+      for (const val of Object.values(obj)) {
+        if (typeof val === 'string' || typeof val === 'number') {
+          values.push(val);
+        } else if (typeof val === 'object') {
+          values.push(...deepValues(val));
+        } else {
+          values.push(val);
+        }
+      }
+
+      return values;
+    };
+
+    const searchRegexp = new RegExp(req.query.search as string, 'i');
+
     try {
       fs.readFileSync(logFile, 'utf-8')
         .split('\n')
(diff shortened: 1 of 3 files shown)


server/routes/settings/index.ts
```diff
@@ -372,6 +374,23 @@ settingsRoutes.get(
       'data',
     ];
 
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const deepValues = (obj: Record<string, any>): any[] => {
```
sct, line 378:
```suggestion
    const deepValues = (obj: Record<string, unknown>): unknown[] => {
```

Maybe `unknown` type here is better?

sct: Also, instead of looping through the whole object this way, I wonder if it would be faster to just JSON.stringify the entire log and just test against that string instead?

danshilm: > Also, instead of looping through the whole object this way, I wonder if it would be faster to just JSON.stringify the entire log and just test against that string instead?

Trying this out with a log file of 10k lines, it is slightly faster (an average of 5.2ms vs 5.9ms over a dozen tries) to search through all of it, but it also is less accurate since the log message keys (`timestamp`, `label`, etc) themselves are also being searched through.

> Maybe unknown type here is better?

If I set the type as unknown, I have to then assert the type when calling the function, so I opted for having it set as `any` instead in the function definition itself (still not ideal 😕).

sct (approved):

PR #2561, opened 2022-02-18
Added Albanian Language File

deniscerri:
Also Translated first 134 lines to Albanian. The rest i will do in the following days. :)


TheCatLady:
Translations are done via Weblate.

PR #2593, opened 2022-03-05
perf(discovery): improve discovery first page performance above the fold

Aghassi:
#### Description

This change aims to improve discovery page performance by
only hydrating above the fold content. The general use case
for machines that load this page tend not to show anything below
"trending". This change leans on intersection observer to
lazy hydrate the page. This means that as the user scrolls through
the discovery page, the content below the fold (the unrendered part
of the app) will fire the SWR requests and fetch data.

This change is aimed to free up network congestion that stalls the page
render. A secondary improvement, not in this diff, is to play with
the Recently Requested component as that may also be slowing down page
load.

Coming... (need to take before and after)

#### To-Dos

#### Issues Fixed or Closed
 N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "react-dom": "17.0.2",
     "react-intersection-observer": "^8.33.1",
     "react-intl": "5.24.6",
+    "react-lazy-hydration": "0.1.0",
     "react-markdown": "^8.0.0",
     "react-select": "^5.2.2",
     "react-spring": "^9.4.2",
--- a/src/components/Discover/index.tsx
+++ b/src/components/Discover/index.tsx
@@ -2,6 +2,7 @@ import { ArrowCircleRightIcon } from '@heroicons/react/outline';
 import Link from 'next/link';
 import React from 'react';
 import { defineMessages, useIntl } from 'react-intl';
+import LazyHydrate from 'react-lazy-hydration';
 import useSWR from 'swr';
 import type { MediaResultsResponse } from '../../../server/interfaces/api/mediaInterfaces';
 import type { RequestResultsResponse } from '../../../server/interfaces/api/requestInterfaces';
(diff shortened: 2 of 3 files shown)


Aghassi:
I should have opened this in draft, whoops 🤷

Aghassi:
@sct Maybe I should have started an issue around this but I'm seeing terrible page load on my instance and it looks to be related to how the app is hydrating my content. But I'm not sure if it's my machine or it's my setup. With a cold cache it takes me almost 10s to load the landing page. My setup is such that my server is running Plex + Radarr + Sonarr + Overseerr and I have Overseerr reverse proxied by Cloudflare. I've tried having overseerr interact with Radarr and sonarr via direct IP and via Cloudflare domain, but it's still a heavy load and I'd like to not have to wait for such long network calls.
image

sct:
A few things.

I have seen other users report a similar issue, but we have never reproduced it. They usually ended up fixing it by adjusting their docker network interface. I think some had success with running Overseerr as host?

As for the actual improvement here, it will be very marginal. Nothing on this page is blocked by any external API servers like Sonarr, Radarr, or Plex. Technically, this will slightly improve performance, but I don't think you will see any massive improvement in your case. I would look into fixing your docker network first off. The most impactful performance sliders are actually recent requests and recently added. For each item, there is an individual request. They will always be above the fold as well.

I am not entirely against adding this because it is generally an improvement, but it's really a marginal improvement in load time. It may also make the experience for other users feel "less snappy" since they will have to wait for content to load when they scroll instead of it already being there?

Also, there is no need to add a dependency just as a note. We already have `react-intersection-observer`, which does the same thing :)

Aghassi:
> A few things.
> 
> 
> 
> I have seen other users report a similar issue, but we have never reproduced it. They usually ended up fixing it by adjusting their docker network interface. I think some had success with running Overseerr as host?
> 
> 
> 
> As for the actual improvement here, it will be very marginal. Nothing on this page is blocked by any external API servers like Sonarr, Radarr, or Plex. Technically, this will slightly improve performance, but I don't think you will see any massive improvement in your case. I would look into fixing your docker network first off. The most impactful performance sliders are actually recent requests and recently added. For each item, there is an individual request. They will always be above the fold as well.
> 
> 
> 
> I don't think this is necessarily a bad addition, but it's really a marginal improvement in load time. It may also make the experience for other users feel "less snappy" since they will have to wait for content to load when they scroll instead of it already being there?
> 
> 
> 
> Also, there is no need to add a dependency just as a note. We already have `react-intersection-observer`, which does the same thing :)

I can give you access to my instance if you'd like to try and "feel" my experience. I was gonna see if maybe it's got something to do with next trying to SSR from within the container today. Thanks for the insight, glad I'm not crazy in thinking it feels slow 😂

Aghassi:
I will also try running as host but that would be very surprising if it were the solve.

danshilm:
> 

Try [setting a DNS](https://docs.docker.com/compose/compose-file/compose-file-v3/#dns) for the container itself.

Aghassi:
> Try [setting a DNS](https://docs.docker.com/compose/compose-file/compose-file-v3/#dns) for the container itself.

Curious, why does this change things for the container? Why would it make this faster? Does it have to do with how the API endpoint makes calls?

danshilm:
I can't really see what specific requests are taking time from the screenshot you shared before that, but slow load times on the discover page have been reported by other users and those slow load times were for the poster images that are loaded from TMDb's CDN itself.

Aghassi:
> I can't really see what specific requests are taking time from the screenshot you shared before that, but slow load times on the discover page have been reported by other users and those slow load times were for the poster images that are loaded from TMDb's CDN itself.

I set the DNS to `1.1.1.1` and it didn't change much. All slow blocking requests are to TMDB, so I guess the question is does overseerr make any attempt to cache these images? If not, we are calling the TMDB quite heavily and that would lead to very expensive round trips if we started getting flagged.

sct:
> > I can't really see what specific requests are taking time from the screenshot you shared before that, but slow load times on the discover page have been reported by other users and those slow load times were for the poster images that are loaded from TMDb's CDN itself.
> 
> 
> 
> I set the DNS to `1.1.1.1` and it didn't change much. All slow blocking requests are to TMDB, so I guess the question is does overseerr make any attempt to cache these images? If not, we are calling the TMDB quite heavily and that would lead to very expensive round trips if we started getting flagged.

This is really no concern. TMDb's CDN is heavily cached on their end and in your browser.

Aghassi:
> > > I can't really see what specific requests are taking time from the screenshot you shared before that, but slow load times on the discover page have been reported by other users and those slow load times were for the poster images that are loaded from TMDb's CDN itself.
> > 
> > 
> > I set the DNS to `1.1.1.1` and it didn't change much. All slow blocking requests are to TMDB, so I guess the question is does overseerr make any attempt to cache these images? If not, we are calling the TMDB quite heavily and that would lead to very expensive round trips if we started getting flagged.
> 
> This is really no concern. TMDb's CDN is heavily cached on their end and in your browser.

Yeah I just confirmed this. The thing I'm seeing is that the container is taking long chunks to respond to requests it seems 
image

Aghassi:
I should add some color to my setup:

1. I run a container with compose that is reverse proxied through `traefik`
2. The reverse proxy is registered through Cloudflare and SSL encrypted with Let's Encrypt
3. Requests go through Clouldflare --> `traefik` --> overseerr container. Traefik and overseerr run on the same box, so connectivity should happen over the docker bridge.

PR #2613, opened 2022-03-14
fix(css): rename form-input to form-input-area

TheCatLady:
#### Description

CSS issue reported by @danshilm from latest Tailwind CSS version bump.

#### To-Dos

#### Issues Fixed or Closed

N/A

### DIFF
--- a/src/components/IssueModal/CreateIssueModal/index.tsx
+++ b/src/components/IssueModal/CreateIssueModal/index.tsx
@@ -177,7 +177,7 @@ const CreateIssueModal: React.FC<CreateIssueModalProps> = ({
                     {intl.formatMessage(messages.problemseason)}
                     <span className="label-required">*</span>
                   </label>
-                  <div className="form-input">
+                  <div className="form-input-area">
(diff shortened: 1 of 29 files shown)


danshilm:
To shed some light on this:
Right now, using the tailwindcss forms plugin, form elements are styled globally and we don't use the `class` "strategy" where one would use `form-{name}` classes to style elements. However, we do also happen to have a CSS class named `form-input` which unfortunately coincides with one of the class names that the forms plugin generates by default in v5.0.0 where form elements are styled globally (the `base` strategy) _and_ class names are generated (the `class` strategy) by default.

My understanding of the `form-input` class we use is that it's meant as a style reset, just like the ones that would be generated by the plugin when using the `class` strategy. So I see that we have 2 options:

1. Opt out of the default `class` strategy in v5.0.0.
2. Rename our `form-input` CSS class to `form-input-area` (or something else) so that we can use it alongside the other generated class names on any HTML element.

danshilm (approved):

PR #2680, opened 2022-04-07
feat: plex deep links for iOS devices

### LINKED ISSUE #2548: Plex app deep links
### Description

Open in Plex opens a browser instance.

### Desired Behavior

Open in Plex opens the media in the corresponding Plex account in the app

### Additional Context

iOS/iPadOS

### Code of Conduct

OwsleyJr:
#### Description

Plex deep link functionality that takes you straight to the iOS/iPadOS Plex app instead of the web app. Detects the device client side and serves the mobile link server side. Android deep links that are browser based do not seem to work at the moment so that is still a WIP.
#### To-Dos

#### Issues Fixed or Closed

- Fixes #2548

### DIFF
--- a/server/entity/Media.ts
+++ b/server/entity/Media.ts
@@ -145,6 +145,9 @@ class Media {
   public plexUrl?: string;
   public plexUrl4k?: string;
 
+  public iOSPlexUrl?: string;
+  public iOSPlexUrl4k?: string;
+
   public tautulliUrl?: string;
   public tautulliUrl4k?: string;
 
@@ -164,6 +167,8 @@ class Media {
         this.ratingKey
       }`;
 
+      this.iOSPlexUrl = `plex://preplay/?metadataKey=%2Flibrary%2Fmetadata%2F${this.ratingKey}&server=${machineId}`;
+
       if (tautulliUrl) {
         this.tautulliUrl = `${tautulliUrl}/info?rating_key=${this.ratingKey}`;
       }
@@ -175,6 +180,8 @@ class Media {
       }#!/server/${machineId}/details?key=%2Flibrary%2Fmetadata%2F${
         this.ratingKey4k
       }`;
+
+      this.iOSPlexUrl4k = `plex://preplay/?metadataKey=%2Flibrary%2Fmetadata%2F${this.ratingKey4k}&server=${machineId}`;
 
       if (tautulliUrl) {
         this.tautulliUrl4k = `${tautulliUrl}/info?rating_key=${this.ratingKey4k}`;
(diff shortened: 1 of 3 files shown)


server/entity/Media.ts
```diff
@@ -164,6 +167,8 @@ class Media {
         this.ratingKey
       }`;
 
+      this.iOSPlexUrl = `plex://preplay/?metadataKey=%2Flibrary%2Fmetadata%2F${this.ratingKey}&server=${machineId}`;
```
TheCatLady, line 170: This shouldn't be set unless there is a rating key, similar to `plexUrl`.

src/components/MovieDetails/index.tsx
```diff
@@ -112,6 +112,39 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
...
+  const [deviceUrl, setDeviceUrl] = useState(data?.mediaInfo?.plexUrl);
+  const [deviceUrl4k, setDeviceUrl4k] = useState(data?.mediaInfo?.plexUrl4k);
+
+  useEffect(() => {
+    if (data) {
+      if (
+        /iPad|iPhone|iPod/.test(navigator.userAgent) ||
+        (navigator.userAgent === 'MacIntel' && navigator.maxTouchPoints > 1)
+      ) {
+        if (data.mediaInfo?.ratingKey) {
```
TheCatLady, line 124: Check for rating key shouldn't be happening here, but on server side.

server/entity/Media.ts
OwsleyJr: They are in the same if statement.

src/components/MovieDetails/index.tsx
OwsleyJr: I agree. Let me change it.

src/components/MovieDetails/index.tsx
```diff
@@ -112,6 +112,31 @@ const MovieDetails: React.FC<MovieDetailsProps> = ({ movie }) => {
   useEffect(() => {
     setShowManager(router.query.manage == '1' ? true : false);
   }, [router.query.manage]);
+  const [deviceUrl, setDeviceUrl] = useState(data?.mediaInfo?.plexUrl);
+  const [deviceUrl4k, setDeviceUrl4k] = useState(data?.mediaInfo?.plexUrl4k);
```
TheCatLady, line 116:
```suggestion
  const [plexUrl, setPlexUrl] = useState(data?.mediaInfo?.plexUrl);
  const [plexUrl4k, setPlexUrl4k] = useState(data?.mediaInfo?.plexUrl4k);
```
I think it's best to call this `plexUrl` since we may support additional media servers in the future, in which case a media item could have a Plex URL in addition to a Jellyfin URL, etc.

OwsleyJr: Updated.

asurare:
Nice, would love to have this feature merges 👍

sct (approved):

sct:
Let's give it a shot in dev and see if anything needs to be tweaked for device detection!

PR #2749, opened 2022-05-13
Fix/recommendations scroll

### LINKED ISSUE #2710: Broken scroll on recommendations page
### Description

When you click on "Recommendations" from a movie or a tv show's page and the resulting list of recommendations is greater than 36 (always happens on my screen when there's at least 6 x 6 tiles displayed with my window size), there's a bug where the page is trying to fill with empty grey posters and the scroll bar goes crazy.

I haven't noticed the issue when there's less 36 results, although it's hard to determine the exact amount of results that triggers the issue because I have no control over the amount of results in the Recommendations page from a movie or a tv show.

I tried adjusting my window size, but that doesn't change anything.

### Version

1.29.1

### Steps to Reproduce

1. Go to a movie or tv show page
2. Scroll down and click on the "Recommendations" link to display all the recommended similar movies/tv shows
3. Scroll down the list of recommendations to the bottom, as much as possible and the bug should trigger
a. The bug only triggers when at least (from my testing) there are more than 36 results
b. You should see grey posters trying to load in a loop, with the scroll bar going crazy

### Screenshots

_No response_

### Logs

_No response_

### Platform

desktop

### Device

MacBook

### Operating System

macOS

### Browser

Safari 15.4, Chrome Canary 102

### Additional Context

_No response_

### Code of Conduct

OwsleyJr:
#### Description
TMDB API limits the total results for recommendations to 40. This would cause an infinite amount of network requests on the recommendations page as it was expecting a higher number of results. The API will now only be called if the total results is greater than 40.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #2710

### DIFF
--- a/src/hooks/useDiscover.ts
+++ b/src/hooks/useDiscover.ts
@@ -1,4 +1,3 @@
-import { useCallback } from 'react';
 import useSWRInfinite from 'swr/infinite';
 import { MediaStatus } from '../../server/constants/media';
 import useSettings from './useSettings';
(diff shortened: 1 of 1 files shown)


danshilm (approved):
Nice catch ;)
Thanks a lot for this.

PR #2779, opened 2022-05-24
feat: user delete modal shows username and requires confirmation

### LINKED ISSUE #2706: Add username to local user deletion confirmation dialog.
### Description

The local user deletion confirmation dialog currently prompts: "Are you sure you want to delete this user? All of their request data will be permanently removed."  There is currently no information presented to the person initiating the deletion which user for which they've clicked the delete button.

### Desired Behavior

Modify the deletion confirmation dialog with language similar to: "Are you sure you want to delete **User X**?"

### Additional Context

_No response_

### Code of Conduct

OwsleyJr:
#### Description

The delete user modal will now show the user that is being deleted and the delete button will now ask you to confirm deletion similar to the delete request button.

Screen Shot 2022-05-24 at 12 23 14 AM

#### To-Dos

#### Issues Fixed or Closed

- Fixes #2706

### DIFF
--- a/cypress/e2e/user/user-list.cy.ts
+++ b/cypress/e2e/user/user-list.cy.ts
@@ -54,7 +54,10 @@ describe('User List', () => {
       .contains('Delete')
       .click();
 
-    cy.get('[data-testid=modal-title]').should('contain', 'Delete User');
+    cy.get('[data-testid=modal-title]').should(
+      'contain',
+      `Delete ${testUser.displayName}`
+    );
 
     cy.intercept('/api/v1/user?take=10&skip=0&sort=displayname').as('user');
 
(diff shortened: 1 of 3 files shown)


TheCatLady:
The UX with the confirm button is somewhat odd... the modal is already asking for confirmation and I don't think the confirm button is necessary here:
image

I think the original, non-confirm button was fine, but the updated modal title with the username is definitely a good call.

OwsleyJr:
> The UX with the confirm button is somewhat odd... the modal is already asking for confirmation and I don't think the confirm button is necessary here: image
> 
> I think the original, non-confirm button was fine, but the updated modal title with the username is definitely a good call.

Hmm makes sense. The modal popping up already is probably enough of a confirmation.

TheCatLady (commented):
Just need to update the failing Cypress test to check for `Delete Test User` instead of `Delete User` and this PR is good to go!

sct (approved):

PR #2782, opened 2022-05-24
Fix rectangular avatars getting stretched

sambartik:
#### Description
Currently the aspect ratio of profile pictures in dropdown is not being preserved, resulting in pictures looking "squished". Adding `object-fit: cover;` to the image element fixes this ([caniuse.](https://caniuse.com/object-fit)).

Before:
Before

After:
After

#### To-Dos

#### Issues Fixed or Closed

No prior issue


sambartik:
> Good catch! Use of the [`object-cover` Tailwind class](https://tailwindcss.com/docs/object-fit) would be better than defining styles on each element, however.

Didn't know, thanks for that! I'll change to that.

sambartik:
It should be okay now.

danshilm:
@all-contributors add @sambartik for code

PR #2789, opened 2022-05-28
feat: Add Job to Remove Requests older than 48 Hrs

CyferShepard:
Description

This Job Will run every minute to check for Requests marked as available that are older than 48 hours and then automatically remove them. This is kind of similar to the feature already present on Ombi.

#### To-Dos


sct:
We don't want to remove requests though, even if they are available. What purpose does that serve? Overseerr also uses request history for things like quotas. This would break that.

PR #2809, opened 2022-06-06
fix(api): lookup show using english title whenever possible

### LINKED ISSUE #2801: Wrong list of TV shows for unmatched requests if the overseer language is not English
### Description

When using other than English language in Overseerr it struggles if a TV show request could not have been automatically matched: _"We couldn't automatically match your request. Please select the correct match from the list below."_ 

It would be fine, however the list contains bunch of completely unrelated TV shows in the language selected in Overseerr.

I do not understand how the matching is done, but wouldn't it be possible to do something about this?

Thanks!

### Version

1.29.1

### Steps to Reproduce

1. Set Overseerr language to Czech (Čeština)
2. Search for Man vs Bee TV show
3. Request
4. The list does not display Man vs Bee show.

### Screenshots

_No response_

### Logs

_No response_

### Platform

desktop

### Device

-

### Operating System

-

### Browser

-

### Additional Context

_No response_

### Code of Conduct

danshilm:
#### Description

Looking up a show that has a missing TVDB ID from Sonarr seems to very rarely yield relevant results when searching with a non-english title compared to with an english title, regardless of whether the show uses english as it's main spoken language or not. This makes the Sonarr show lookup use the english title whenever available.

N/A

#### To-Dos

#### Issues Fixed or Closed

- Fixes #2801

### DIFF
--- a/server/routes/service.ts
+++ b/server/routes/service.ts
@@ -194,7 +194,9 @@ serviceRoutes.get<{ tmdbId: string }>(
         language: req.locale ?? (req.query.language as string),
       });
 
-      const response = await sonarr.getSeriesByTitle(tv.name);
+      const response = await sonarr.getSeriesByTitle(
+        tv.original_language === 'en' ? tv.original_name : tv.name
+      );
 
       return res.status(200).json(response);
     } catch (e) {


server/routes/service.ts
```diff
@@ -194,7 +194,9 @@ serviceRoutes.get<{ tmdbId: string }>(
         language: req.locale ?? (req.query.language as string),
       });
 
-      const response = await sonarr.getSeriesByTitle(tv.name);
+      const response = await sonarr.getSeriesByTitle(
+        tv.original_language === 'en' ? tv.original_name : tv.name
```
TheCatLady, line 198: This won't help unless the original language was English. Would be better to fetch the English name (when locale isn't English) and then perform the Sonarr search.

TheCatLady:
Fixed in #2911

PR #2885, opened 2022-08-03
feat: watchlist sync integration

### LINKED ISSUE #2678: Add a "From your Plex Watchlist" Row to Overseerr Discovery Page for Plex-authenticated users
### Description

For users who use authenticate with their Plex credentials, it would be great to show their Plex Watchlist as a row in Discovery with the ability to request items not currently available on the server.

Here is an example query builder for the API (from Sonarr):
```
var requestBuilder = new HttpRequestBuilder("https://metadata.provider.plex.tv/library/sections/watchlist/all")
                    .Accept(HttpAccept.Json)
                    .AddQueryParam("clientID", clientIdentifier)
                    .AddQueryParam("context[device][product]", BuildInfo.AppName)
                    .AddQueryParam("context[device][platform]", "Windows")
                    .AddQueryParam("context[device][platformVersion]", "7")
                    .AddQueryParam("context[device][version]", BuildInfo.Version.ToString())
                    .AddQueryParam("includeFields", "title,type,year,ratingKey")
                    .AddQueryParam("includeElements", "Guid")
                    .AddQueryParam("sort", "watchlistedAt:desc")

if (!string.IsNullOrWhiteSpace(authToken))
{
       requestBuilder.AddQueryParam("X-Plex-Token", authToken);
}
```

### Desired Behavior

On the Discovery page, show a row called "From Your Plex Watchlist" that shows media items for everything you have added to your Plex Watchlist, with the same status and actions as Media Items in other rows (ie, Trending).

### Additional Context

_No response_

### Code of Conduct

sct:
#### Description

This PR will add Plex Watchlist support into Overseerr.

- [x] Add Plex Watchlist Slider to discover page
- [x] Add full Plex Watchlist discover page (performance may be an issue here)
- [x] Sync user requests with plex watchlist
- [x] Add user option to enable/disable watchlist sync

#### To-Dos

#### Issues Fixed or Closed

- Closes #2678


tdeverx:
I've been testing this PR for a few days and I just want to give you the heads up that there is a _minor_ issue..

BUT, theres a possibility you may of fixed it already since I'm a few PR behind (I'm currently on [43dd4ff](https://github.com/sct/overseerr/pull/2885/commits/43dd4ff45d915803f3062fd114d6dc3d56177234))

Basically, pretty much everything is working intended, but Overseerr had a little heart attack and decided to request items that are already present in Sonarr/Radarr/Plex. From what I can tell, it's anything that isn't "Available". Which sounds okay, but "Partially Available" Shows which are already downloaded are getting added to the requests and redownloaded 😅 

Thought it was a fluke at first but when I cancelled all the downloads and deleted all the requests, they re-appeared and started to download again with the next sync 👀

sct:
> I've been testing this PR for a few days and I just want to give you the heads up that there is a _minor_ issue..
> 
> BUT, theres a possibility you may of fixed it already since I'm a few PR behind (I'm currently on [43dd4ff](https://github.com/sct/overseerr/pull/2885/commits/43dd4ff45d915803f3062fd114d6dc3d56177234))
> 
> Basically, pretty much everything is working intended, but Overseerr had a little heart attack and decided to request items that are already present in Sonarr/Radarr/Plex. From what I can tell, it's anything that isn't "Available". Which sounds okay, but "Partially Available" Shows which are already downloaded are getting added to the requests and redownloaded 😅
> 
> _(Maybe there needs to be a new status for Shows that are up to date by using the episode air dates as a baseline?)_
> 
> Thought it was a fluke at first but when I cancelled all the downloads and deleted all the requests, they re-appeared and started to download again with the next sync 👀

Is this happening on the latest version? it should only request things that are not already available. Series it will try to request all the missing seasons (if it can).

tdeverx:
Just to make sure I'm understanding correctly, what you're saying is if an item has not been requested in Overseerr it won't check with Sonarr/Radarr first to see if an item is already present and monitored? 

Also, is that why airing shows that are technically "Available" but marked as "Partially Available" are redownloading all the episodes again? Because the season technically isn't available because it's not finished airing? confused

sct:
> Just to make sure I'm understanding correctly, what you're saying is if an item has not been requested in Overseerr it won't check with Sonarr/Radarr first to see if an item is already present and monitored?
> 
> Also, is that why airing shows that are technically "Available" but marked as "Partially Available" are redownloading all the episodes again? Because the season technically isn't available because it's not finished airing? confused
> 
> I'm still updating to the latest PR, the CI takes forever 😭

Overseerr will let you request any series not already marked as available _in Overseerr_. If you delete your requests and you don't see the seasons marked as available, it will let you re-request that season regardless. Are you seeing different behavior here?

tdeverx:
I'm not sure the best way to describe what's happening, but I'm going to use "Pretty Little Liars: Original Sin" & "American Horror Story" here as this is two of the items it's happening to.

Before the sync (no prior requests):
1. If I visit the show page for PLL and hit request, S01 is greyed out and marked as "Partially Available" (as it should be) as all episodes are monitored (correct logic here?)

2. If I visit the show page for AHS when I hit request, all seasons are greyed out and marked as "Available" excluding S11 which is "Requested"

After the sync:
1. PLL is added to the requests even though Overseerr recognises this show is "Available" on both Sonarr & Plex, this wouldn't normally happen no? Then, all episodes, even those that are available, have started to download again for, seemingly, no reason..

2. AHS is added to the requests when the show is "Available" on both Sonarr & Plex, then all episodes, even those that are available, have started to download again** 😭 

Does this make a little more sense haha, this also happened with a lot more items than the ones in this example 😅 

** a lot of episodes

sct:
> I'm not sure the best way to describe what's happening, but I'm going to use "Pretty Little Liars: Original Sin" & "American Horror Story" here as this is two of the items it's happening to.
> 
> Before the sync (no prior requests):
> 
> 1. If I visit the show page for PLL and hit request, S01 is greyed out and marked as "Partially Available" (as it should be) as all episodes are monitored (correct logic here?). I also cant request anything as the button is greyed out with "Already Requested"
> 2. If I visit the show page for AHS when I hit request, all seasons are greyed out and marked as "Available" excluding S11 which is "Requested", and I'm also unable to request anything.
> 
> After the sync:
> 
> 1. PLL is added to the requests even though Overseerr recognises this show is "Available" on both Sonarr & Plex, this wouldn't normally happen no? Then, all episodes, even those that are available, have started to download again for, seemingly, no reason..
> 2. AHS is added to the requests when the show is "Available" on both Sonarr & Plex, then all episodes, even those that are available, have started to download again** 😭
> 
> Does this make a little more sense haha, this also happened with a lot more items than the ones in this example 😅
> 
> ** a lot of episodes

I got you. Just pushed a fix. Basically, this is our existing requesting logic repurposed. We never checked before if a season was marked as available (without a request) because the frontend UI would prevent the request anyways. Now that it's being done programmatically through the watchlist sync this issue surfaced. The fix I just pushed will prevent requests for seasons that are already marked as available/partially available.

tdeverx:
Perfect, so glad you caught this before this PR was merged 😬

server/entity/MediaRequest.ts
```diff
@@ -242,6 +242,22 @@ export class MediaRequest {
...
+      if (media.seasons) {
+        existingSeasons = [
+          ...existingSeasons,
+          ...media.seasons
+            .filter(
+              (season) =>
+                season[requestBody.is4k ? 'status4k' : 'status'] ===
+                  MediaStatus.AVAILABLE ||
+                season[requestBody.is4k ? 'status4k' : 'status'] ===
+                  MediaStatus.PARTIALLY_AVAILABLE
```
tdeverx, line 255: I think this still needs a filter for seasons that are have been marked as "Requested", I'm was still experiencing the same issue but it seems this is the reason why 😃

tdeverx: Also, I think theres still an issue with whatever logic is used to trigger Sonarr to search for Seasons. Season 11 is the only one missing from AHS as it hasn't aired yet, and we know the only reason AHS is getting added to Overseerr requests (right now) is because it's missing another filter?. BUT, what's curious is that it's still attempting to re-download every single season** 🤔 

**Overseerr only has the badge for S11 under requests

sct: I'm confused. Is it requesting things that are already requested?

tdeverx: When words fail, basically this video will show you what I described above ^

https://vimeo.com/740224465

Side-note, while recording this I came across another bug where it added the entry twice because it appeared on two watchlists 👀

sct: I see. Because you have it present in Sonarr/Radarr. I will fix it.

As for double requests, I can fix this by running each users watchlist requests in succession instead of asynchronously.

tdeverx: Hopefully I won't encounter anymore issues after this, feel like I'm teasing you every time you get close to a PR haha

On that note (sorry again), but I just noticed that if you modify a request that has synced from the watchlist and change it to request X amount of seasons instead of all of them, the next sync will add another request for the missing seasons 😅

sct: Pushed a fix for the other issues. Try it out.

> On that note (sorry again), but I just noticed that if you modify a request that has synced from the watchlist and change it to request X amount of seasons instead of all of them, the next sync will add another request for the missing seasons 😅

That's not really a bug. Watchlist items for shows will always request for all seasons each time. We can't do granular requests here. If two users both add the same show, and you edit one? How do you know the other didn't want all seasons? How do we determine what to request for each user? I think it's either all or nothing. I imagine in most setups, admins will want to disable automated show requests :)

tdeverx: No problem, I'm just waiting on the CI to publish and I'll let you know if I hit anymore issues.

> That's not really a bug.

Oh I assumed as much, people that do have it enabled will most likely have auto approve on too. It was a super small edge case I figured I may as well mention. Out of curiosity, with this fix, if I modified the request to include less seasons but marked the seasons I didn't want as monitored in Sonarr, it should skip the request on the next sync right?

sct: In its current state, it would make a new request for the missing seasons. it's all or nothing.

We are talking now about maybe adding a way to track automated season requests. So if you decline them, you don't get more from the same user.

tdeverx: I actually didn't think of that, what would currently happen if a request is declined? Will it be stuck in an endless loop of requesting until it's accepted? If thats is the case, what would the best solution? Maybe some form of timed item blacklist per user?

sct: Im adding the ability to track auto requests. Each user will only get one auto-request per-media.

tdeverx: Gotcha, assuming this data will be cached, will you be adding an option to flush it?

sct: Deleting the auto-request would allow new ones.

tdeverx:
Alright, seasons have officially been fixed haha! Only issue left really is that Movies are getting requested when they're already available in Radarr. I believe this would've just needed a similar filter, but as I mentioned earlier, the frontend UI doesn't actually block you from requesting a Movie thats already monitored in Radarr (unlike seasons). It's pretty much a non-issue right now and doesn't need to be addressed immediately. I do however think if a movie is monitored in Radarr it should be blocked similar to seasons and tagged as requested. This PR should be gooood to go 🎉

TheCatLady:
> Only issue left really is that Movies are getting requested when they're already monitored in Radarr (unreleased movies etc). I believe this would've just needed a similar filter, but as I mentioned earlier, the frontend UI doesn't actually block you from requesting a Movie thats already monitored in Radarr (unlike seasons).

If a movie is monitored in Radarr and `Enable Scan` is enabled in Overseerr for that Radarr server, then the media item should get marked as requested during the next scan (once every 24 hours). The movie would then no longer be requestable.

tdeverx:
> If a movie is monitored in Radarr and `Enable Scan` is enabled in Overseerr for that Radarr server, then the media item should get marked as requested during the next scan (once every 24 hours). The movie would then no longer be requestable. (This is how seasons monitored in Sonarr get marked as requested as well.)

If that's what the expected behaviour is, then I think something is broken as it's not currently the case for me. From both the frontend UI and the watchlist sync. Overseerr has a link to the Radarr entry, but theres no indicator that the item is "Requested". Guessing thats why the watchlist sync keeps adding the movie back into the requests? I also get the option to request the movie from the frontend too, the option isn't greyed out

tdeverx:
Forcing the refresh added the tag, I'll just have to stay mindful of prematurely deleting approved requests as the watchlist sync is running to frequently to keep up with the Radarr sync.

All is well, might I suggest (if possible) some logic to force update the status of a show without the need of a database refresh? I only ask as Overseerr already had a working direct link to the item in Radarr before the refresh took place to update the status. 

P.S. thanks for all of the your hardwork on this PR, it's going to save me a lot of time with my technically incompetent family members 🙏

sct:
> Forcing the refresh added the tag, I'll just have to stay mindful of prematurely deleting approved requests as the watchlist sync is running to frequently to keep up with the Radarr sync.
> 
> All is well, might I suggest (if possible) some logic to force update the status of a show without the need of a database refresh? I only ask as Overseerr already had a working direct link to the item in Radarr before the refresh took place to update the status.
> 
> P.S. thanks for all of the your hardwork on this PR, it's going to save me a lot of time with my technically incompetent family members 🙏

You should really never delete requests. Overseerr is designed for the request history to stay available. Deleting should be reserved for fixing mistakes.

tdeverx:
While it might not be intended to be removed, I'm probably not the only person with OCD and a burning urge to clear house as soon as they're available 😅 

Thanks again, hope I haven't annoyed you too much hahaha

src/components/UserProfile/UserSettings/UserGeneralSettings/index.tsx
```diff
@@ -49,6 +49,12 @@ const messages = defineMessages({
   discordIdTip:
     'The <FindDiscordIdLink>multi-digit ID number</FindDiscordIdLink> associated with your Discord user account',
   validationDiscordId: 'You must provide a valid Discord user ID',
+  plexwatchlistsyncmovies: 'Auto-Request Movies',
+  plexwatchlistsyncmoviestip:
+    'Automatically request movies on your Plex watchlist',
+  plexwatchlistsyncseries: 'Auto-Request Series',
+  plexwatchlistsyncseriestip:
+    'Automatically request series on your Plex watchlist',
```
TheCatLady, line 57:
```suggestion
  plexwatchlistsyncmovies: 'Auto-Request Movies',
  plexwatchlistsyncmoviestip:
    'Automatically request movies on your Plex Watchlist',
  plexwatchlistsyncseries: 'Auto-Request Series',
  plexwatchlistsyncseriestip:
    'Automatically request series on your Plex Watchlist',
```
Plex capitalizes "Watchlist" as a proper noun in their support articles, so we should do the same?

src/components/UserProfile/UserSettings/UserGeneralSettings/index.tsx
```diff
@@ -409,6 +419,65 @@ const UserGeneralSettings = () => {
...
+                      onChange={() => {
+                        setFieldValue(
+                          'watchlistSyncTv',
+                          !values.watchlistSyncTv
+                        );
+                      }}
+                    />
+                  </div>
+                </div>
+              )}
```
TheCatLady, line 480: Since this feature is only usable by Plex users, should we hide or fade/disable these options for local users?

src/components/UserProfile/UserSettings/UserGeneralSettings/index.tsx
```diff
@@ -409,6 +419,65 @@ const UserGeneralSettings = () => {
                     </div>
                   </>
                 )}
+              {hasPermission(
+                [Permission.AUTO_REQUEST, Permission.AUTO_REQUEST_MOVIE],
+                { type: 'or' }
+              ) && (
```
TheCatLady, line 425: Since this feature is only usable by Plex users, should we hide these options for local users? (Similar to how we only show notification agents that have been enabled.)

TheCatLady (commented):
Looking good! Will try to do some testing in my dev env soon 🙂

src/components/UserProfile/UserSettings/UserGeneralSettings/index.tsx
```diff
@@ -409,6 +419,99 @@ const UserGeneralSettings = () => {
...
+                            ) => (
+                              <a
+                                href="https://support.plex.tv/articles/universal-watchlist/"
+                                className="text-white transition duration-300 hover:underline"
+                                target="_blank"
+                                rel="noreferrer"
+                              >
+                                {msg}
+                              </a>
+                            ),
```
TheCatLady, line 450: Would it be better to define this once instead of duplicating it for both movies & TV? Like we do for the `OpenPgpLink` for email https://github.com/sct/overseerr/blob/5247f14968c1f5a1a2efdc26f850cf81c0c04f0e/src/components/Settings/Notifications/NotificationsEmail.tsx#L49-L55

TheCatLady:
Don't forget to re-run `yarn i18n:extract` :)

src/components/Discover/index.tsx
```diff
@@ -93,6 +104,37 @@ const Discover = () => {
...
         emptyMessage={intl.formatMessage(messages.noRequests)}
       />
+      <div className="slider-header">
+        <Link href="/discover/watchlist">
+          <a className="slider-title">
+            <span>{intl.formatMessage(messages.plexwatchlist)}</span>
+            <ArrowCircleRightIcon />
+          </a>
+        </Link>
+      </div>
```
TheCatLady, line 114: This needs to be conditionally displayed along with the below slider. Otherwise, when there are no watchlist items, you just see the heading.

src/components/Discover/index.tsx
```diff
@@ -44,6 +46,15 @@ const Discover = () => {
...
     );
 
+  const { data: watchlistItems, error: watchlistError } = useSWR<{
+    page: number;
+    totalPages: number;
+    totalResults: number;
+    results: WatchlistItem[];
+  }>('/api/v1/discover/watchlist', {
+    revalidateOnMount: true,
+  });
```
TheCatLady, line 56: We can check if the user is a Plex user and only make this API call if they are.

server/routes/discover.ts
```diff
@@ -704,4 +709,50 @@ discoverRoutes.get<{ language: string }, GenreSliderItem[]>(
...
 
+discoverRoutes.get<
+  { page?: number },
+  {
+    page: number;
+    totalPages: number;
+    totalResults: number;
+    results: WatchlistItem[];
+  }
+>('/watchlist', async (req, res) => {
```
TheCatLady, line 720:
```suggestion
discoverRoutes.get<
  { page?: number },
  {
    page: number;
    totalPages: number;
    totalResults: number;
    results: WatchlistItem[];
  }
>('/watchlist', async (req, res) => {
```
Would be nice to add an optional `user` parameter, which would allow for displaying of watchlists on user profiles. (Would we want to add another permission to view other users' watchlists? Or restrict this only to the user themselves and admins?)

TheCatLady (commented):
I think we should add a new notification type for automatic requests (to notify users that something from their watchlist was requested on their behalf), but maybe good to add it in a separate PR after this is merged?

src/components/Discover/DiscoverWatchlist/index.tsx
```diff
@@ -0,0 +1,51 @@
+import ListView from '../../Common/ListView';
+import { defineMessages, useIntl } from 'react-intl';
+import Header from '../../Common/Header';
+import PageTitle from '../../Common/PageTitle';
+import useDiscover from '../../../hooks/useDiscover';
+import Error from '../../../pages/_error';
+import type { WatchlistItem } from '../../../../server/interfaces/api/discoverInterfaces';
+
+const messages = defineMessages({
+  discoverwatchlist: 'Plex Watchlist',
```
TheCatLady, line 10:
```suggestion
  discoverwatchlist: 'Your Plex Watchlist',
```

src/components/Discover/index.tsx
```diff
@@ -25,6 +26,7 @@ const messages = defineMessages({
   noRequests: 'No requests.',
   upcoming: 'Upcoming Movies',
   trending: 'Trending',
+  plexwatchlist: 'Plex Watchlist',
```
TheCatLady, line 29:
```suggestion
  plexwatchlist: 'Your Plex Watchlist',
```

TheCatLady (commented):
Clarify that it's the user's own Watchlist and not a curated list from elsewhere? Also makes it feel more personal and friendly :)

src/components/Common/ListView/index.tsx
```diff
@@ -34,6 +38,18 @@ const ListView = ({
...
       <ul className="cards-vertical">
+        {plexItems?.map((title, index) => {
+          return (
+            <li key={`${title.ratingKey}-${index}`}>
+              <TmdbTitleCard
+                id={title.tmdbId}
+                tmdbId={title.tmdbId}
+                type={title.mediaType}
+                canExpand
+              />
```
TheCatLady, line 49: If we ever have access to a TheTVDB ID, would be good to include it here as well to aid in identifying items that no longer exist on TMDB.

TheCatLady:
Another idea that just popped into my head -- would be nice to display a `On Your Watchlist` badge on the media detail pages, next to the media status badges. But we can also table this for later 😊

src/components/UserProfile/UserSettings/UserGeneralSettings/index.tsx
sct: I don't think we need to. It's only used in two places and they are fairly close to each other.

src/components/Discover/index.tsx
sct: Huh. Could have sworn this was in the below statement 🤔 Fixing!

server/routes/discover.ts
sct: This is also something we can maybe look at doing later as well.

src/components/Common/ListView/index.tsx
sct: We aren't returning tvdb id's with this data atm. Items that no longer exist in TMDB shouldn't come back from Plex's watchlist API. It would also be filtered out on our end if there was no TMDB ID.

TheCatLady (approved):

server/api/plextv.ts
```diff
@@ -253,6 +287,81 @@ class PlexTvAPI {
...
+          params: {
+            'X-Plex-Container-Start': offset,
+            'X-Plex-Container-Size': size,
+          },
+          baseURL: 'https://metadata.provider.plex.tv',
+        }
+      );
+
+      const watchlistDetails = await Promise.all(
+        response.data.MediaContainer.Metadata.map(async (watchlistItem) => {
```
TheCatLady, line 313: Ah, found one more thing... this can be `undefined` if the user's watchlist is empty.

```
2022-08-21T06:26:31.346Z [error][Plex.TV Metadata API]: Failed to retrieve watchlist items{"errorMessage":"Cannot read properties of undefined (reading 'map')"}
```

sct: Good catch!

PR #2896, opened 2022-08-09
feat(plextv): add custom applicationTitle to Plex OAuth

marcusronnback:
#### Description
When using the custom application title from the settings the Plex OAuth flow still uses the old one (as seen [here](https://github.com/sct/overseerr/issues/404#issuecomment-1199953428). This change makes it so that the applicationTitle is read from the settings.json file on disk. 

I initially tried to read from `server/lib/settings.ts` but I did not succeed. Maybe someone else has better luck, this solution does however work.

#### Issues Fixed or Closed
- Fixes additional issue in comments on #404.

### DIFF
--- a/src/utils/plex.ts
+++ b/src/utils/plex.ts
@@ -1,5 +1,6 @@
 import axios from 'axios';
 import Bowser from 'bowser';
+import config from '../../config/settings.json';
 
 interface PlexHeaders extends Record<string, string> {
   Accept: string;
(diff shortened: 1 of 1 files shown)


TheCatLady:
We actually used to use the application title here, but ended up just hardcoding the value because some users included special characters in their app titles that broke it. 😔

See f7146e41899a59f75b963e1cc9dac9eddf24aebe and the linked issue.

marcusronnback:
> We actually used to use the application title here, but ended up just hardcoding the value because some users included special characters in their app titles that broke it. 😔
> 
> See [f7146e4](https://github.com/sct/overseerr/commit/f7146e41899a59f75b963e1cc9dac9eddf24aebe) and the linked issue.

Ah I see! I missed that issue! Would it be possible to sanitize the applicationTitle instead as to remove forbidden characters? I could give the implementation a try!

TheCatLady:
> Ah I see! I missed that issue! Would it be possible to sanitize the applicationTitle instead as to remove forbidden characters? I could give the implementation a try!

Yeah, it was a long time ago. But yes, I think sanitizing or encoding the value would be fine!

PR #2903, opened 2022-08-14
test: add cypress foundation

sct:
#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -26,3 +26,4 @@ public/os_logo_filled.png
 public/preview.jpg
 snap
 stylelint.config.js
+cypress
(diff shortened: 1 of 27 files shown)


TheCatLady:
LGTM! Do we also want to `.dockerignore` some of this stuff so it doesn't end up in the built Docker images?

sct:
> LGTM! Do we also want to `.dockerignore` some of this stuff so it doesn't end up in the built Docker images?

Oh yes we should! We don't really need the cypress folder in the container.

TheCatLady (approved):

PR #2950, opened 2022-08-19
feat: tooltip foundation

sct:
#### Description

- Adds `<Tooltip />` component, which you can use to wrap anything to add tooltips!
- Also removed React import wherever possible since you no longer need to import it for JSX since React 17

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


sct:
Reviewing this might be annoying so for the tooltip stuff just look at the first commit on its own: https://github.com/sct/overseerr/pull/2950/commits/7f29ac655b9c7585e4581e678aca0ccda0b0ee53

PR #2959, opened 2022-08-21
feat: view other users' watchlists

TheCatLady:
#### Description

Adds new permission to view other users' Plex Watchlists.

Also updates the `prepareTestDb.ts` script to assign the Plex server owner's plex ID to the two created test users, and to modify existing test users if the script has already been run in the past (currently, it fails because emails must be unique). Additionally, sets a Gravatar URL for these users instead of a broken Plex avatar URL.

#### To-Dos


cypress/e2e/discover.cy.ts
```diff
@@ -173,17 +173,17 @@ describe('Discover', () => {
   });
 
   it('loads plex watchlist', () => {
-    cy.intercept('/api/v1/discover/watchlist', { fixture: 'watchlist' }).as(
-      'getWatchlist'
-    );
+    cy.intercept('/api/v1/discover/watchlist', {
+      fixture: 'watchlist.json',
```
sct, line 177: You shouldn't need to append `.json` here?

TheCatLady: For some reason it was broken in the other test if I didn't. Seems to be safer to append it, and the docs examples include the full filename.

sct: Weird. I got the example from the docs without the full filename 🤔 But if this works better, no worries.

sct (approved):

PR #2964, opened 2022-08-22
feat: restore option to cache and optimize images locally

sct:
#### Description

A few versions ago, NextJS fixed the local image generation to pass the cache control header from the origin. This means the locally created images are now correctly cached in the user's browser and performance has improved!

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


TheCatLady (approved):
LGTM!

PR #2969, opened 2022-08-23
feat: improved user dropdown

sct:
#### Description

👀 

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


src/components/Layout/UserDropdown/MiniQuotaDisplay/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+                    className="h-8 w-8"
+                    progress={Math.round(
+                      ((data?.movie.remaining ?? 0) /
+                        (data?.movie.limit ?? 1)) *
+                        100
+                    )}
+                    useHeatLevel
+                  />
+                  <span className="text-lg font-bold text-gray-200">
+                    {data?.movie.remaining}/{data?.movie.limit}
```
TheCatLady, line 45: The heading above is kind of misleading here in that it suggests that this is the number of requests used, not remaining. 🤔

I suggest adding a string `"{used} of {limit} remaining"`, similar to what we have in the `QuotaDisplay` component but with the days omitted.

src/components/Layout/UserDropdown/MiniQuotaDisplay/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+                  </span>
+                </>
+              ) : (
+                <span className="text-sm text-gray-200">
+                  <Infinity className="h-8" />
```
TheCatLady, line 50: Can add `Unlimited` text next to this so that it better matches the mini progress circle with accompanying text?

src/components/Layout/UserDropdown/MiniQuotaDisplay/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+const MiniQuotaDisplay = ({ userId }: MiniQuotaDisplayProps) => {
+  const intl = useIntl();
+  const { data } = useSWR<QuotaResponse>(`/api/v1/user/${userId}/quota`, {
+    suspense: true,
+  });
+
+  return (
+    <Suspense fallback={<SmallLoadingSpinner />}>
+      {(data?.movie.limit ?? 0) !== 0 && (data?.tv.limit ?? 0) !== 0 && (
```
TheCatLady, line 26: On profile pages, we display that both movie & series requests are unlimited. Do we not want to do the same here?

TheCatLady (commented):
Looking so good! 🔥

src/components/Layout/UserDropdown/MiniQuotaDisplay/index.tsx
sct: I'm not sure we have space for that here. I think it's pretty clear this is remaining quota.

src/components/Layout/UserDropdown/MiniQuotaDisplay/index.tsx
sct: Hm. Same thing here. The icon sizes are different. I think it's clear without accompanying text. I'll see how it looks.

src/components/Layout/UserDropdown/MiniQuotaDisplay/index.tsx
sct: No reason to show it if both are unlimited. It's just dead space at that point.

TheCatLady: I was thinking more that if we end up making these shortcuts to view one's own movie/series requests, that it would be helpful to have these elements exist regardless.

Or, what do you think about adding a `Requests` entry along with `Profile` and `Settings` in the dropdown that links to `/profile/requests` (currently missing but added as a fix in #2190)?

sct: Good catch. And because the type is a number so it's better to compare with numbers even though 0 is still falsey

src/components/Layout/UserDropdown/MiniQuotaDisplay/index.tsx
TheCatLady: I think it'd be nicer to "square" the icon so that it occupies the same space as the progress circle? 🤔  Also, we can add the infinity icon to the actual profile pages, too!

src/components/Layout/UserDropdown/MiniQuotaDisplay/index.tsx
sct: > I was thinking more that if we end up making these shortcuts to view one's own movie/series requests, that it would be helpful to have these elements exist regardless.
> 
> 
> 
> Or, what do you think about adding a `Requests` entry along with `Profile` and `Settings` in the dropdown that links to `/profile/requests` (currently missing but added as a fix in #2190)?

I'm okay with adding requests!

src/components/Layout/UserDropdown/MiniQuotaDisplay/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+                  <ProgressCircle
+                    className="h-8 w-8"
+                    progress={Math.round(
+                      ((data?.movie.remaining ?? 0) /
+                        (data?.movie.limit ?? 1)) *
+                        100
+                    )}
+                    useHeatLevel
+                  />
+                  <span className="text-lg font-bold tracking-widest">
```
TheCatLady, line 44: Hm, wouldn't `tracking-widest` make it look odd when the values are multiple digits? My idea behind adding the space was to create some break/spacing between the elements.

sct: Oh you're right! That would look super weird! 🤦 Fixing!

TheCatLady (approved):
LGTM!

PR #2974, opened 2022-08-24
fix: transition animation

OwsleyJr:
#### Description

Fix: Switched to using the headlessui transition component instead of the react-css-transition component due to new version breaking the animation.

 - Modified components to use new transition.
 - Removed transition file as we are no longer using it.

#### To-Dos


TheCatLady:
We no longer need the `react-transition-group` package, then! 😺

OwsleyJr:
> We no longer need the `react-transition-group` package, then! 😺

Ah, good catch!

TheCatLady (approved):
LGTM!

PR #2988, opened 2022-08-26
Run nodejs with the node user

xr09:
#### Description

This is a small change to follow the [best practices](https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#non-root-user) and run nodejs as a non-root user.

By default the Nodejs Docker image [provides](https://github.com/nodejs/docker-node/blob/main/Dockerfile-alpine.template#L5-L6) a `node` user that we can use in this case.

#### To-Dos

#### Issues Fixed or Closed

- Running as root is [not recommended](https://docs.bitnami.com/tutorials/why-non-root-containers-are-important-for-security) if the app doesn't truly need it to function
- Some environments actively prevent containers to gain root privileges (e.g. k8s clusters using Gatekeeper)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,9 @@ RUN apk add --no-cache tzdata tini && rm -rf /tmp/*
 
 # copy from build image
 COPY --from=BUILD_IMAGE /app ./
+RUN chown node:node /app/config -R
 
+USER node
 ENTRYPOINT [ "/sbin/tini", "--" ]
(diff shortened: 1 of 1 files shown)


TheCatLady:
This will break things for existing installations if user ID `1000` on the host does not have read/write access to the volume mounted config directory.

Additionally, it is already possible to run the container as non-root: https://docs.overseerr.dev/getting-started/installation

PR #2998, opened 2022-08-30
fix: do not display 'Request More' button if no requestable seasons

TheCatLady:
#### Description

Partially available seasons are not requestable.

#### To-Dos

### DIFF
--- a/src/components/RequestButton/index.tsx
+++ b/src/components/RequestButton/index.tsx
@@ -292,7 +292,6 @@ const RequestButton = ({
     }) &&
     media &&
     media.status !== MediaStatus.AVAILABLE &&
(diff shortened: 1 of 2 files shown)


src/components/RequestButton/index.tsx
```diff
@@ -292,7 +292,6 @@ const RequestButton = ({
     }) &&
     media &&
     media.status !== MediaStatus.AVAILABLE &&
-    media.status !== MediaStatus.UNKNOWN &&
```
TheCatLady, line 295: This is not necessary due to this now being an `else if` clause.

sct (approved):

PR #3024, opened 2022-09-16
fix: new status indicators added to series list

OwsleyJr:
#### Description

New status indicators/badges added to series list to accommodate mobile.

 - Added a common reusable component.

Screen Shot 2022-09-16 at 4 58 31 PM

Screen Shot 2022-09-16 at 5 00 21 PM

#### To-Dos


src/components/Common/BadgeMobile/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+    | 'warning'
+    | 'warning4K'
+    | 'success'
+    | 'success4K';
+  className?: string;
+  href?: string;
+  children: React.ReactNode;
+}
+
+const BadgeMobile = (
```
sct, line 19: Let's name this StatusBadgeMobile instead and refactor it to handle rendering the icons as well.

src/components/Common/BadgeMobile/index.tsx
```diff
@@ -0,0 +1,83 @@
+import { forwardRef } from 'react';
+
+interface BadgeMobileProps {
+  badgeType?:
```
sct, line 4: We should add an is4k prop instead of having a type for each one.

Also, instead of the styling types like primary or warning, let's instead change this to take a media status as a prop.

src/components/Common/BadgeMobile/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+    | 'primary4K'
+    | 'danger'
+    | 'danger4K'
+    | 'warning'
+    | 'warning4K'
+    | 'success'
+    | 'success4K';
+  className?: string;
+  href?: string;
+  children: React.ReactNode;
```
sct, line 16: We won't need any of these anymore. Let's also drop the forwardRef that is wrapping this component.

src/components/TvDetails/index.tsx
```diff
@@ -35,7 +36,13 @@ import {
   FilmIcon,
   PlayIcon,
 } from '@heroicons/react/outline';
-import { ChevronUpIcon } from '@heroicons/react/solid';
+import {
+  BellIcon,
+  CheckIcon,
+  ChevronUpIcon,
+  ClockIcon,
+  MinusSmIcon,
```
sct, line 44: Move the rendering of these into the badge component itself.

sct (changes requested):

src/components/Common/StatusBadgeMobile/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+      badgeStyle.push('bg-green-500 ring-green-400');
+      indicatorIcon = <MinusSmIcon />;
+      break;
+    default:
+      badgeStyle.push('bg-indigo-500 ring-indigo-400');
+      indicatorIcon = <ClockIcon />;
+  }
+
+  return (
+    <div className="inline-flex whitespace-nowrap rounded-full text-xs font-semibold leading-5 ring-1 ring-gray-700 md:hidden">
```
sct, line 45: We shouldn't handle hiding this on desktop here. If we want to use these in the TitleCards in the future, this would prevent it. Instead handle whether or not the component is displayed from the place that's using it.

We should probably also rename the component for clarity now that I think about it. Maybe `StatusBadgeMini` is a better name,

sct (changes requested):

src/components/Common/StatusBadgeMobile/index.tsx
TheCatLady: I think `StatusBadge` is fine... we could add a parameter later for different sizes? 🤔

sct: > Maybe we could add a parameter later for different sizes and handle this all within the `StatusBadge` component? 🤔

Not sure if it's necessary since every place we would use this wouldn't need the extra functionality the `StatusBadge` offers, such as linking to the media/etc.

TheCatLady: That's true :)

src/components/Common/StatusBadgeMini/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+interface StatusBadgeMiniProps {
+  status?: MediaStatus;
+  is4k?: boolean;
+  indicatorIcon?: React.ReactNode;
+}
+
+const StatusBadgeMini = ({
+  status,
+  is4k = false,
+  indicatorIcon,
```
sct, line 18: This shouldn't be a prop. It should be a "let" inside the component.

sct (changes requested):

src/components/Common/StatusBadgeMini/index.tsx
```diff
@@ -0,0 +1,52 @@
+import {
+  BellIcon,
+  CheckIcon,
+  ClockIcon,
+  MinusSmIcon,
+} from '@heroicons/react/solid';
+import { MediaStatus } from '@server/constants/media';
+
+interface StatusBadgeMiniProps {
+  status?: MediaStatus;
```
sct, line 10: status should probably not be an optional prop. We can then change the switch to get rid of the default and just write explicit cases for each status type.

sct (changes requested):

src/components/Common/StatusBadgeMini/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  CheckIcon,
+  ClockIcon,
+  MinusSmIcon,
+} from '@heroicons/react/solid';
+import { MediaStatus } from '@server/constants/media';
+
+interface StatusBadgeMiniProps {
+  status: MediaStatus;
+  is4k?: boolean;
+  indicatorIcon?: React.ReactNode;
```
sct, line 12: Don't forget to remove this from the type.

OwsleyJr: woops

src/components/Common/StatusBadgeMini/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+const StatusBadgeMini = ({ status, is4k = false }: StatusBadgeMiniProps) => {
+  const badgeStyle = ['w-5 rounded-full p-0.5 text-white ring-1'];
+  let indicatorIcon;
```
sct, line 17:
```
let indicatorIcon: React.ReactNode;
```

sct (approved):

PR #3036, opened 2022-09-19
fix(ui): minor fixes

### LINKED ISSUE #3025: Filter button / Hide already existing
### Description

It would be very nice when searching for movies of an actor to have an option not to show the movies that are already there. As it is now you possibly do not see a movie/series without the green hook icon, which is easily to overlook with the design.

image

### Desired Behavior

Add a button or something similiar to hide the already downloaded movies/series.

### Additional Context

_No response_

### Code of Conduct

danshilm:
#### Description

- Hide available media when the setting "Hide Available Media" is enabled on the `person` page
- Fixed a copy-paste error for the label for the new image cache setting
- Fixed bug that would show tooltips on the `collection` page (that badge doesn't open anything when clicked nor would the tooltip have any description)
- Replaced a weird empty space when there are no episodes in a season and the episodes list is expanded - preventing the episodes list from expanding would be unexpected for the user since there'd be nothing to indicate that it shouldn't be expandable (picture below)
- Set Plex Watchlist Sync job to short interval

Before:
image

After:
image

#### To-Dos

#### Issues Fixed or Closed

- Fixes #3025

### DIFF
--- a/server/models/Tv.ts
+++ b/server/models/Tv.ts
@@ -50,7 +50,7 @@ interface Season {
   seasonNumber: number;
 }
 
-export interface SeasonWithEpisodes extends Season {
+export interface SeasonWithEpisodes extends Omit<Season, 'episodeCount'> {
   episodes: Episode[];
   externalIds: ExternalIds;
 }
@@ -141,7 +141,6 @@ export const mapSeasonWithEpisodes = (
   season: TmdbSeasonWithEpisodes
 ): SeasonWithEpisodes => ({
   airDate: season.air_date,
-  episodeCount: season.episodes.length,
   episodes: season.episodes.map(mapEpisodeResult),
   externalIds: mapExternalIds(season.external_ids),
   id: season.id,
--- a/server/models/Tv.ts
+++ b/server/models/Tv.ts
@@ -141,7 +141,7 @@ export const mapSeasonWithEpisodes = (
   season: TmdbSeasonWithEpisodes
 ): SeasonWithEpisodes => ({
   airDate: season.air_date,
-  episodeCount: season.episode_count,
+  episodeCount: season.episodes.length,
   episodes: season.episodes.map(mapEpisodeResult),
   externalIds: mapExternalIds(season.external_ids),
   id: season.id,
--- a/src/components/StatusBadge/index.tsx
+++ b/src/components/StatusBadge/index.tsx
@@ -23,7 +23,6 @@ interface StatusBadgeProps {
   serviceUrl?: string;
   tmdbId?: number;
   mediaType?: 'movie' | 'tv';
-  enableTooltip?: boolean;
 }
 
 const StatusBadge = ({
@@ -34,7 +33,6 @@ const StatusBadge = ({
   serviceUrl,
   tmdbId,
   mediaType,
-  enableTooltip = true,
 }: StatusBadgeProps) => {
   const intl = useIntl();
   const { hasPermission } = useUser();
@@ -80,7 +78,7 @@ const StatusBadge = ({
           mediaType === 'movie' ? globalMessages.movie : globalMessages.tvshow
         ),
       });
-    } else if (hasPermission(Permission.ADMIN)) {
+    } else if (hasPermission(Permission.ADMIN) && serviceUrl) {
       mediaLink = serviceUrl;
       mediaLinkDescription = intl.formatMessage(messages.openinarr, {
         arr: mediaType === 'movie' ? 'Radarr' : 'Sonarr',
@@ -151,12 +149,12 @@ const StatusBadge = ({
       break;
   }
 
-  // regardless of whether tooltip is enabled or not
+  // regardless of whether badge should have a tooltip
   if (selectedBadge === null) {
     return null;
   }
 
-  return enableTooltip ? (
+  return mediaLink ? (
     <Tooltip content={mediaLinkDescription}>{selectedBadge}</Tooltip>
   ) : (
     selectedBadge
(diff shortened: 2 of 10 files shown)


src/components/TvDetails/Season/index.tsx
```diff
@@ -6,6 +6,7 @@ import useSWR from 'swr';
 
 const messages = defineMessages({
   somethingwentwrong: 'Something went wrong while retrieving season data.',
+  noepisodes: 'Episodes list unavailable',
```
TheCatLady, line 9:
```suggestion
  noepisodes: 'Episode list unavailable.',
```

src/components/StatusBadge/index.tsx
```diff
@@ -23,6 +23,7 @@ interface StatusBadgeProps {
   serviceUrl?: string;
   tmdbId?: number;
   mediaType?: 'movie' | 'tv';
+  enableTooltip?: boolean;
```
TheCatLady, line 26: As discussed via DMs, we shouldn't have to add this parameter. The issue is that the case where `serviceUrl` is `null` is not handled properly, and we can resolve this by fixing the `else if` clause.

src/components/TvDetails/Season/index.tsx
```diff
@@ -6,6 +6,7 @@ import useSWR from 'swr';
 
 const messages = defineMessages({
   somethingwentwrong: 'Something went wrong while retrieving season data.',
+  noepisodes: 'Episode list unavailable',
```
TheCatLady, line 9:
```suggestion
  noepisodes: 'Episode list unavailable.',
```

src/components/StatusBadge/index.tsx
```diff
@@ -87,80 +88,77 @@ const StatusBadge = ({
...
+  // regardless of whether badge should have a tooltip
+  if (selectedBadge === null) {
+    return null;
+  }
+
+  return mediaLink ? (
+    <Tooltip content={mediaLinkDescription}>{selectedBadge}</Tooltip>
+  ) : (
+    selectedBadge
+  );
```
TheCatLady, line 161: These changes should be reverted.

src/components/TvDetails/Season/index.tsx
```diff
@@ -29,34 +30,38 @@ const Season = ({ seasonNumber, tvId }: SeasonProps) => {
...
-              className="flex flex-col space-y-4 py-4 xl:flex-row xl:space-y-4 xl:space-x-4"
-              key={`season-${seasonNumber}-episode-${episode.episodeNumber}`}
-            >
-              <div className="flex-1">
-                <div className="flex flex-col space-y-2 xl:flex-row xl:items-center xl:space-y-0 xl:space-x-2">
-                  <h3 className="text-lg">{episode.name}</h3>
-                  {episode.airDate && (
-                    <AirDateBadge airDate={episode.airDate} />
-                  )}
+      {data.episodeCount === 0 ? (
```
TheCatLady, line 33: Is it possible for the episode count to be nonzero but to have no episode data? I haven't looked into this, just wondering if that might be the case for announced upcoming seasons.

server/models/Tv.ts
```diff
@@ -141,7 +141,7 @@ export const mapSeasonWithEpisodes = (
   season: TmdbSeasonWithEpisodes
 ): SeasonWithEpisodes => ({
   airDate: season.air_date,
-  episodeCount: season.episode_count,
+  episodeCount: season.episodes.length,
```
TheCatLady, line 144: Ah, I don't think we should be using the length instead of the value returned by TMDB here. The number of planned episodes is useful information even if we don't have the episode data yet.

src/components/TvDetails/Season/index.tsx
TheCatLady: Edit: See following comment. Should check length of `episodes` here instead.

src/components/StatusBadge/index.tsx
TheCatLady: (Line 43 above as well.)

danshilm: If I don't remove the `Tooltip` element, there's an empty tooltip that does appear on hover:
image

So, either I refactored the Tooltip component, or I conditionally wrapped the badge in the StatusBadge component, and I chose the later.

TheCatLady: I think that it's better for the tooltip to not display when empty.

server/models/Tv.ts
danshilm: The `episode_count` isn't actually returned by the API call for the season details. And the planned episodes are present in the `episodes` field.

src/components/StatusBadge/index.tsx
danshilm: In that case, is there a specific way to do this that you have in mind or does what I have here work for you?

server/models/Tv.ts
TheCatLady: Regardless, I don't think it is good to manipulate this data here. This should be the data from TMDB...

src/components/StatusBadge/index.tsx
TheCatLady: Ah sorry I meant the Tooltip component should not display anything when content is empty. The logic doesn't belong here in the StatusBadge component IMO.

danshilm: Gotcha, will refactor 👍🏼

server/models/Tv.ts
danshilm: Gotcha, will remove this in this case 👍🏼

sct (approved):
LGTM

PR #3045, opened 2022-09-22
fix: slideover close button placement when using PWA

OwsleyJr:
#### Description

Safe area inset on top placement was being overridden by top placed on the component. This would only affect the slideover when using the PWA version.

 - Added !important to prevent this issue again.

IMG_5056 IMG_5055

#### To-Dos


src/styles/globals.css
```diff
@@ -42,7 +42,7 @@
   }
 
   .sidebar-close-button {
-    top: env(safe-area-inset-top);
+    top: env(safe-area-inset-top) !important;
```
sct, line 45: If you are removing the `top-0` class from the classes is `!important` still required here?

OwsleyJr: Yes because it's !important

OwsleyJr: I'll remove it

sct (approved):

PR #3086, opened 2022-10-17
feat(ui): add episode shorthand to tv details UI (SxxExx)

aedelbro:
#### Description
To provide the user more information concisely when viewing TV Details, this includes a "shorthand" for the episode `SxxExx`.  In the future, I'd like to add an on hover that shows the _absolute_ episode number. 

As it currently stands:
Screen Shot 2022-10-17 at 5 23 31 PM

With the shorthand badge:
Screen Shot 2022-10-17 at 5 21 49 PM

#### To-Dos

- [X] Translation keys `yarn i18n:extract`

#### References
[Here's](https://discord.com/channels/783137440809746482/1027318650409062450) the discord discussion thread

#### Potential Issues
While testing a few other TV Shows, sometimes it shows the absolute number. I've so far only found 1 show where this happens, so it might be limited to just it. 
When viewing One Piece, it shows the _absolute_ episode number, instead of the episode number for just that season.
Example:
Screen Shot 2022-10-17 at 5 28 33 PM
While it's not the end of the world, it's not desired (I think an on hover for absolute episode number would be best).

Additionally, it's not currently padding the episode or season number. That would require additional information to be returned to the UI, or significantly more logic. I _could_ setup padding on the episode number within the season, but not between seasons.

### DIFF
--- a/src/components/TvDetails/Season/index.tsx
+++ b/src/components/TvDetails/Season/index.tsx
@@ -44,7 +44,9 @@ const Season = ({ seasonNumber, tvId }: SeasonProps) => {
               >
                 <div className="flex-1">
                   <div className="flex flex-col space-y-2 xl:flex-row xl:items-center xl:space-y-0 xl:space-x-2">
-                    <h3 className="text-lg">{episode.name}</h3>
+                    <h3 className="text-lg">
+                      {episode.episodeNumber} - {episode.name}
+                    </h3>
                     {episode.airDate && (
                       <AirDateBadge airDate={episode.airDate} />
                     )}


danshilm:
We were actually just talking about this a few days ago. The season number in the badge is redundant, so that's not needed. Just the episode number would be enough. But I'm not sure whether it would be better as just `1. <Episode Title>` or in a badge.

As for the absolute episode number being shown, that's an issue for TMDB to deal with as far as I know. There is an absolute order numbering that's available on TMDB, so I'm really not sure why they're using the absolute numbering together with the seasons. Not ideal but well...

sct:
Like @danshilm said. I think showing the season number again is redundant. We can keep it simple like `1. Episode Name`.

aedelbro:
While seasons with only 8-12 episodes, the `Sxx` is not really needed. However, if a user has multiple seasons open, with 60-80 episodes per season, it could be useful to know where you are in the scroll.

Additionally, if an absolute episode number is implemented later (my thought is on hover of the badge), do you have thoughts on how to implement that with just the number up front?

aedelbro:
@sct @danshilm any thoughts on the above?

Or just put the episode number in text up front and call it good?

sct:
> @sct @danshilm any thoughts on the above?
> 
> Or just put the episode number in text up front and call it good?

Sorry for the late reply. I think your argument about the season number being helpful in very long lists is one heck of an edge case but I don't mind it so much.

Looking at this again, I think my least favorite part is that we have a badge both before _and_ after the title. Maybe moving the season/episode number badge next to the date would look better?

aedelbro:
@sct I agree on the sandwich feeling weird. I was really trying to make the badge work, as I'd really like the absolute episode number available in the future, and on hover of the badge feels like a natural way to do it. However, I feel like the episode number should be up front, and the air date info / badges feel like they should at the end. 

In the interest of getting this out, I'll revert these commits, and just make the episode number show up front. 

I feel like that's the easiest, quickest, least controversial change to get the episode number out, and the absolute number / badge / whatever can be addressed later if it ever comes back up.

sct:
> @sct I agree on the sandwich feeling weird. I was really trying to make the badge work, as I'd really like the absolute episode number available in the future, and on hover of the badge feels like a natural way to do it. However, I feel like the episode number should be up front, and the air date info / badges feel like they should at the end. 
> 
> 
> 
> In the interest of getting this out, I'll revert these commits, and just make the episode number show up front. 
> 
> 
> 
> I feel like that's the easiest, quickest, least controversial change to get the episode number out, and the absolute number / badge / whatever can be addressed later if it ever comes back up.

Sounds good. Let's start with that and see if we can improve afterwards.

aedelbro:
@sct got the simple change made. I just kept it in the same title line for formatting.

ex:
Screen Shot 2023-01-13 at 11 55 58 AM

sct (approved):
LGTM! Thank you!

sct:
@all-contributors please add @aedelbro for code

sct:
@holopin-bot @aedelbro add please :)

Issue #1638, opened 2021-05-13
Login with other services


ghost:
#### Description

Is your feature request related to a problem? If so, please provide a clear and concise description of the problem. E.g., "I'm always frustrated when [...]."

I have some family members who I would like to give access to but will probably result in me being constantly asked for their password. Adding support for providers like auth0 would allow me to enable sending a code via sms or email for them to enter on the page to login as an authentication method which would work better than a link if the web app is added to their home screen.

#### Desired Behavior

Allow a third party authentication service like auth0 to either login directly to the web app similar to plex or using SSO from a reverse proxy.

danshilm:
There was already an issue similar to this: #332 but it was closed by the stale bot.

lenaxia:
There's also this issue: https://github.com/sct/overseerr/issues/1555

I'd make an argument here again for the implementing Forward Authentication. Because it'd also give us LDAP support for free since, for example, Authelia would handle it.

To summarize my response from the othe rissue, Forward Authentication is not a security risk as it is a well accepted way of authenticating applications, as can bee seen in Grafana (https://grafana.com/docs/grafana/latest/auth/auth-proxy/). 

Also, Forward Auth supports the core mission of Overseerr by enabling people to safely present Overseerr to other users, over the open internet in a safe manner. This would be done by placing Overseerr behind a reverse proxy like Traefik, and allowing the reverse proxy to handle authentication through a service like Authelia. This means that a security issue in Overseerr isn't a critical failure that would risk exposing my entire network to an intruder. Instead my attack surface is still limited to just Authelia and Traefik. 

The Traefik forward auth docs: https://doc.traefik.io/traefik/middlewares/forwardauth/
The Authalia documentation on Forward Authentication: https://www.authelia.com/docs/community/using-remote-user-header-for-sso-with-jira.html

Authelia supports LDAP, Yubikeys, Duo, and other 2 factor authentication, so you're really getting all of this stuff for free when you support forward auth.

sct:
> There's also this issue: #1555
> 
> I'd make an argument here again for the implementing Forward Authentication. Because it'd also give us LDAP support for free since, for example, Authelia would handle it.
> 
> To summarize my response from the othe rissue, Forward Authentication is not a security risk as it is a well accepted way of authenticating applications, as can bee seen in Grafana (https://grafana.com/docs/grafana/latest/auth/auth-proxy/).
> 
[...]

Issue #1555, opened 2021-05-03
X-Remote-User


yeyeoke:
#### Description

Not really a problem, more of an enhancement since right now it requires my users to login once to my domain, and once again to Overseerr. A simple fix would be to accept X-Remote-User or something like that. 

#### Desired Behavior

What I would like to be implemented is header based authentication, i.e. I login with my username and password to my domain, my webserver then forwards X-Remote-User to Overseerr and Overseerr accepts this and logs me in to a user with the same username I previously used to login to my domain. 

I.e. User1 logs in to domain/authelia (which im using right now) it forwards X-Remote-User to Overseerr which proceeds to log me in as User1.

sct:
This would be a massive security hole. Won't be adding this.

yeyeoke:
Hi, I'm sorry I was probably unclear in my suggestion. What I'm really asking for is Basic Authentication using a HTTP Header. 

For a better explanation see Grafanas implementation: 

https://grafana.com/docs/grafana/latest/auth/auth-proxy/

I'm essentially asking for a way to outsource authentication to my webserver which instead of letting Overseerr handle it. Is this still a massive security hole? 

If so I very much respect your stance, I'm simply trying to get a better understanding of if and why it's unsafe.

lenaxia:
Hi, I'd like to second this. This isn't a security issue. This is a standard implementation of remote-user. As yeyeoke pointed out, grafana is a huge example.

I would not feel comfortable putting Overseerr on the open internet, however I would be comfortable putting it behind Authelia and Traefik as a reverse proxy setup that provide a layer of security that I can be confident in.

Here is the Authalia documentation on Forward Authentication: https://www.authelia.com/docs/community/using-remote-user-header-for-sso-with-jira.html

And the Traefik forward auth docs: https://doc.traefik.io/traefik/middlewares/forwardauth/

sami-shakir:
I agree with them also. It is not the best user experience to have them authenticate with authelia, then authenticate to overseerr.

PR #3015, opened 2022-09-12
feat: decouple Plex as a requirement for setting up/using Overseerr

sct:
#### Description

This is the first step to opening up Overseerr to allow other media server integrations. Previously, Overseerr required signing in with a Plex account to get it configured. You can now create a local account instead of using Plex. If a local account is used, you will need to connect your Plex account still to use the Plex Media Server features.

#### Screenshot (if UI-related)

image
image

#### To-Dos

- [x] Decouple plex from the setup process
- [x] Remove UserType from users (opens possibility for multiple user types per user)
- [x] Add option to link/unlink Plex account from user settings (instead of only while logging in)
- [ ] Allow changing emails
- [ ] Disable Plex login when owner account does not have plex configured
- [ ] Do not allow new plex connections to accounts if owner account unlinks plex
- [ ] Successful build `yarn build`
- [ ] Translation keys `yarn i18n:extract`
- [ ] Database migration (if required)

PR #3105, opened 2022-11-01
feat(auth): add ForwardAuth support via `X-Plex-Token` header

tobz:
#### Description

This is an attempt to partially add ForwardAuth support to Overseerr, specifically for Plex-based logins.

#1555 and #1638 have already covered much of the ground on the wants/needs of Overseerr users when it comes to authentication, but this solution focuses purely on the ForwardAuth usecase: an application running _behind_ a reverse proxy (or similar) system that derives authentication information itself (or via some other mechanism) and "forwards" that authentication to the proxied application. This PR does _not_ concern itself with augmenting Overseerr's existing authentication methods with support for external identity providers, and so on.

In the ForwardAuth model, "trusted headers" are sent to the underlying application which carry the authentication information: username, e-mail, etc. In this PR, we're exclusively focusing on support for passing an existing Plex token as a trusted header, which drives newly-added logic to automatically create and/or log in the user based on that Plex token.

This new feature obeys the following rules/logic:
- we do not create a new user if an imported one does not already exist and "Enable New Plex Sign-In" is disabled
- we have added a new setting, "Enable Forward Auth Via Plex Token", which must be enabled for any of this new logic to take effect
- we always attempt to create and/or login a user via this mechanism so long as the above constraints are meant, regardless of which page they access

Essentially, the login flow is skipped entirely if the above constraints are met: on a normal load to `/`, the first call to `/api/v1/auth/me` would log the user in via their Plex token, create and initialize their session, set their session cookie, and allow the UI to send them straight to the normal dashboard landing page.

To say it out loud: this is predicated on at least sending the ForwardAuth header along to the `/api/v1/auth/me` endpoint, but generally ForwardAuth implementations will send their authentication headers uniformly to all requests being proxied.

Beyond that, I've tried to refactor some of the Plex login code to be reusable between the actual Plex login endpoint and the code I've added for ForwardAuth support. I'm not a TypeScript developer by trade, so apologies if things are in the wrong place, not idiomatic, etc.

#### To-Dos


lenaxia:
Request to make the header configurable. To simplify it should be configurable via environment variable (no need for something as fancy as a UX). I use Authelia which uses the Remote-User and would absolutely be onboard for using this. I just added a comment in #1638 revisiting why forward auth is superior to OIDC, however it really limits the use of this feature if its limited to just Plex's token.

tobz:
> Request to make the header configurable. To simplify it should be configurable via environment variable (no need for something as fancy as a UX). I use Authelia which uses the Remote-User and would absolutely be onboard for using this. I just added a comment in https://github.com/sct/overseerr/issues/1638 revisiting why forward auth is superior to OIDC, however it really limits the use of this feature if its limited to just Plex's token. Not all my users have plex accounts (in my case they share a plex pass account) since Plex doesn't allow loading of local accounts, but they do have their own accounts for other services.

I can understand your use case, and it makes sense. That said, adding support here to additionally support local accounts via ForwardAuth would be a significant increase in logic on top of what I've already built... which, given the seemingly low chance that this PR ever gets a real review and approved/merged, is just that much more work for (potentially) nothing.

tobz:
@sct @TheCatLady @danshilm 

Apologies for the direct ping, but, is it possible to review this? Or, if nothing else, to tell me whether or not you're even willing to accept this feature?

sct:
Hi @tobz. Sorry for the wait. We are still very busy, but we want to get to this! I am also in the process of refactoring some login and user type logic with the update to decouple plex in #3015. This will likely affect some of the logic you have here. So please give us a little longer.

tobz:
> Hi @tobz. Sorry for the wait. We are still very busy, but we want to get to this! I am also in the process of refactoring some login and user type logic with the update to decouple plex in #3015. This will likely affect some of the logic you have here. So please give us a little longer.

Understandable. 👍🏻 

I'll rebase my branch on top of yours at some point and see what it looks like, and get it as close as I reasonably can for whenever #3015 lands.

tobz:
@sct Checking in again.

It looks like your work on the Plex decoupling stuff stalled out around 2-3 months ago. No judgment at all whatsoever, but given that we talked about that work landing before coming back to revisit this PR getting merged... could we potentially come to a decision around whether to merge this or not?

The PR has been open for almost a year at this point. 😅

geman220:
Would love to see this implemented

PR #3108, opened 2022-11-04
[Documentation] Updated docker volumes from /app/config to /config

GabrielJean:
#### Description
Hello,

I just discovered overseerr recently and when I was trying to install it using docker and the snippets provided by the documentation the app was complaining that the /config mount wasn't working. I realized that the app is expecting the mount to be at /config but the documentation has it at /app/config. So, I simply want to update the doc to make it easier for everyone. 

Thanks. 

#### To-Dos

- [x] docker compose up works as expected

#### Issues Fixed or Closed

No open issues


TheCatLady:
The correct path for the official container is `/app/config`.

PR #3123, opened 2022-11-17
feat(ui): request card progress bar

### LINKED ISSUE #994: Add User Permission to see Movie/TvShow Download Status
#### Description

When a user requests a movie/tvshow for the admin to approve and then movie/tvshow is approved and subsequently downloading, the state transitions to "processing". Whilst in processing, only the admin can see the state of the movie/tvshow downloading with an ETA and percentage completion bar.

#### Desired Behavior

Please add a permission e.g. Download Status that will allow the admin to grant users the ability to see the status of movies/tvshows that are downloading.

OwsleyJr:
#### Description

**Status Badge Progress Bar**

As I know that this is a very requested feature, I thought I could implement it! The main feature of this PR is to display a movie/show progress on the status badge until it is finished processing. I also added to the tooltip a quick peek that will display more download information similar to the manageSlideover. Below I will go into detail on the exact changes.

- Status badge
   - The core change is that it will now show the progress of each movie or show that is currently downloading. If the case is a show, then it will display the progress of each episode with the season and episode number.
   - If a show is partially available and in the process of downloading, the status badge will now display processing instead of partially available.
   - The progress is calculated from the download item that is being passed in from each parent component that houses the status badge.

- Tooltip
   - Download block is now being passed into the tooltip on the status badge to display more information about the downloads progress.
   - Used ReactDOM.createPortal to prevent overflow-hidden on the parent component.
   - Added classname prop to make it dynamic.
   - Title on download block will change depending on user permissions.

- Servarr backend
   - Added params to retrieve the season and episode number from the Sonarr API.
   - New optional type called episode added to DownloadBlock interface.

Screenshot 2022-11-17 at 5 18 06 PM
Screenshot 2022-11-17 at 5 18 14 PM
Screenshot 2022-11-17 at 5 18 33 PM

#### Issues Fixed or Closed

- Fixes #994


server/api/servarr/base.ts
```diff
@@ -61,6 +77,7 @@ interface QueueItem {
   protocol: string;
   downloadClient: string;
   indexer: string;
+  episode: EpisodeResult;
```
sct, line 80: `episode` is not returned for Radarr so we should make this not required or extend the type just for the Sonarr response.

src/components/CollectionDetails/index.tsx
```diff
@@ -51,6 +54,22 @@ const CollectionDetails = ({ collection }: CollectionDetailsProps) => {
...
+        ? item.mediaInfo?.downloadStatus4k?.forEach((status) =>
+            tempArray.push(status)
+          )
+        : item.mediaInfo?.downloadStatus?.forEach((status) =>
+            tempArray.push(status)
+          )
+    );
+
+    setDownloadStatus(tempArray);
+  }, [data?.parts]);
```
sct, line 71: We don't need a useEffect or a separate state to do this. you can instead just set a const and use `useMemo` with the data.parts as a dependency.

src/components/CollectionDetails/index.tsx
```diff
@@ -218,6 +260,29 @@ const CollectionDetails = ({ collection }: CollectionDetailsProps) => {
...
+                                (item) =>
+                                  item.mediaInfo?.externalServiceId4k ===
+                                  status.externalId
+                              )
+                              .map((title) => title.title)
+                              .join('')}
+                          />
+                        </li>
+                      ))}
+                    </ul>
```
sct, line 284: Feels strange we are duplicating this everywhere we use the `StatusBadge`. Can we just move this into `StatusBadge` itself?

src/components/DownloadBlock/index.tsx
```diff
@@ -1,23 +1,42 @@
...
+}: DownloadBlockProps) => {
   const intl = useIntl();
+  const { hasPermission } = useUser();
+
+  const formattedTitle: string =
+    title +
+    intl.formatMessage(messages.seasonepisodenumber, {
+      seasonNumber: downloadItem?.episode?.seasonNumber,
+      episodeNumber: downloadItem?.episode?.episodeNumber,
+    });
```
sct, line 30:
```suggestion
  const formattedTitle =
    intl.formatMessage(messages.seasonepisodenumber, {
    title,
      seasonNumber: downloadItem?.episode?.seasonNumber,
      episodeNumber: downloadItem?.episode?.episodeNumber,
    });
```

And then update the i18n string to have {title} in it

Fallenbagel:
Hi. Love the work on this PR, I just have a few comments on the progress bar design. I was thinking, wouldn't this design be more inline with the overall overseerr ui? (Just a thought)
(The purple background with translucent white progress bar)

Screenshot_20221212_202758_Firefox Beta

OwsleyJr:
> Hi. Love the work on this PR, I just have a few comments on the progress bar design. I was thinking, wouldn't this design be more inline with the overall overseerr ui? (Just a thought)
> 
> (The purple background with light purple hue progress bar instead of transparent background with dark purple progress bar)
> 
> Screenshot_20221212_202758_Firefox Beta

Readability is easier if the background is black to contrast the progress bar. Also, the design already used in the PR is more inline with the direction Overseerr is headed. The new design of the request modal is a good example.

sct (approved):

PR #3147, opened 2022-12-01
fix: series displayed an empty season with series list/request modal

OwsleyJr:
#### Description

**Empty Episode Count**

Season list and request modal on the tv details page would show empty seasons. Specifically for the request modal, it would allow you to request a season with 0 episodes. Another addition in this PR is a change to the ChevronIcon for consistency across the app.

- Episode count is now added as a conditional to check for seasons.
- ChevronUpIcon changed to ChevronDownIcon

#### To-Dos

### DIFF
--- a/src/components/RequestModal/TvRequestModal.tsx
+++ b/src/components/RequestModal/TvRequestModal.tsx
@@ -232,7 +232,7 @@ const TvRequestModal = ({
 
   const getAllSeasons = (): number[] => {
     return (data?.seasons ?? [])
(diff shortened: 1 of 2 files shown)


danshilm (dismissed):
On the TV details page, the `seasonCount` also needs to not count seasons where there are zero episodes since that's also displayed (under the show's title). The way the show's availability (both non-4K and 4K) is calculated for use in the request button then needs to be updated to correctly use that updated `seasonCount`.

Issue #1744, opened 2021-06-07
Ability to block certain media from being requested.


Iguana9999phil:
#### Description

Not really a problem but it's frustrating me a lot. ;-) 
I would like to see a feature added that will give me the ability to put media on a "blacklist" when it get's requested. While I got rules in place regarding what to request and what gets added, some users don't read nor do they think before requesting. A button to add a request to the blacklist which after that would auto reject a next request for that same media would be great.

#### Desired Behavior

Someone requests all seasons of "East enders" I deny and put it on the "blacklist" next time someone requests "East enders" it will get auto rejected.

Magikarplvl4:
Just to add a small request to this case:
In my case, dutch tv shows and movies are really hard to find for Sonarr/Radarr.
Almost 99% of the time, a dutch tv show or movie request will fail because a file can't be found. 

It will be really nice if you had an option to block, manually approve this kind of request or set the status to manual process so I can deal with it manually.

costaht:
There's another issue, that I can't find right now, that suggests something different which is to black-list Declined requests.

Melantrix:
I would like to add a filter option for this; If it would be possible to blacklist media that are available on certain networks. e.g. i don't want to download media that's on Netflix. If i'm correct, the metadata is already available, so only a filter for this (or in this case a blacklist) would be ideal.

TheCatLady:
@Melantrix Sounds like you're looking for #835 instead. 😉

PR #3167, opened 2022-12-17
feat: blacklist media when declined a media request

Maxentr:
#### Description

First of all, this is my first PR in an open-source project like that. I am open to constructive criticism.

Add the ability to blacklist a media from being requested (idea submit in https://github.com/sct/overseerr/issues/1744). Currently, the user can only blacklist a media when this one has been requested. Can a page managing all blacklisted media be a good idea?

image

##### Back
- [X] Database migration
- [X] Add query in delete request route to get users' blacklist choice
- [X] before inserting a new request, if the media requested is blacklisted then deny the request
- [ ] Manage all blacklisted media?

##### Front
- [X] Display modal with a checkbox to blacklist or not the media on media request deletion
- [ ] Add a page to manage blacklisted media?

#### Issues

- Feature: https://github.com/sct/overseerr/issues/1744


sct:
Hi @Maxentr, thanks for making the PR, and sorry for taking so long to get to it.

One thing we don't want to encourage users to do is to delete requests. Even in the case of blacklisting, we want to know that a request has still happened. Request history is important to Overseerr for many reasons.

I am thinking of two things:
1. An option to automatically blacklist anything declined. This could be a global toggle; simply clicking decline will blacklist the item. This means we only need to check when a request is created if _any_ any request already exists. (Currently, we check if requests already exist but ignore declined ones)
2. We can still add the blacklist boolean you have in this PR, but it should be an option to toggle when we Decline the request instead of deleting it. Consider using something like our ButtonWithDropdown component for the decline buttons on request cards and in the request list. You would have the typical "Decline" but also a dropdown to "Decline and Blacklist" or something?

What do you think?

Maxentr:
> Hi @Maxentr, thanks for making the PR, and sorry for taking so long to get to it.
> 
> One thing we don't want to encourage users to do is to delete requests. Even in the case of blacklisting, we want to know that a request has still happened. Request history is important to Overseerr for many reasons.
> 
> I am thinking of two things:
> 
>     1. An option to automatically blacklist anything declined. This could be a global toggle; simply clicking decline will blacklist the item. This means we only need to check when a request is created if _any_ any request already exists. (Currently, we check if requests already exist but ignore declined ones)
> 
>     2. We can still add the blacklist boolean you have in this PR, but it should be an option to toggle when we Decline the request instead of deleting it. Consider using something like our ButtonWithDropdown component for the decline buttons on request cards and in the request list. You would have the typical "Decline" but also a dropdown to "Decline and Blacklist" or something?
> 
> 
> What do you think?

Both ideas sound good to me. Do you have any ideas on how to handle blacklisted media?

PR #3161, opened 2022-12-10
feat: Media availability synchronization job 

Nimelrian:
#### Description
A friend notified me of requiring the changes made in #2890, but the PR being dormant. From what I gathered, the only thing missing - aside from pulling in upstream changes - was adding pagination for the media entity loading. I have added pagination (with a page size of 50 entities) via a simple async generator function.

In addition, I pulled in the upstream changes via a rebase.

Disclaimer: I do not have an actual use for overseerr myself and don't run an instance, so I hope someone else can actually test the changes I made.

#### To-Dos

- [x] Successful build `yarn build`
- [x] Translation keys `yarn i18n:extract`

#### Issues Fixed or Closed

- Fixes #377

PR #3219, opened 2023-01-06
feat: availability sync rework

### LINKED ISSUE #377: Deleted titles are stuck on "Available"

OwsleyJr:
#### Description

Well, well, if it isn't Availability Sync number three. To spare everyone with what's already been said in the other two PRs, here are links to them #2890 #3161. Before I give an update on what I've have added with this PR, I would first like to give a shout out to Nimelrian and Jariz for the work they already put in. Now to the details.

 _New Additions/Changes_
   - 4k Compatibility
      - Due to the fact that we can not delete an entity if a non-4k or 4k version still exists, we must only update the entity and set its status to UNKNOWN if unavailable. This also includes updating all of its keys such as the ratingKey, etc to NULL.
      - We will cross compare radarr/sonarr versions with plex and update accordingly.
      - Media entity will be updated and set to unavailable if it does not exist in any radarr/sonarr instance and plex is unable to retrieve a movie/show with its rating keys. Deletion is prevented due to the fact that we do not want associated issues deleted.
   - Request handling
      - Before requests were only removed if the media entity was deleted as well. Now if either a non-4k or 4k version and one still remains, we will delete the correlated request.
      - Season request deletion was a little trickier as there is the child season request and then the parent media request. The availability sync itself will now handle the season requests that need to be removed and then the parent will automatically delete if empty.
      - Media requests will now also be deleted if a season entity is deleted.
   - Sonarr/Radarr media retrieval
      - Made sure loop that checks every media id will continue if one of the APIs fails to find a show/movie with it's tvdb/tmdb id.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #377 
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

### DIFF
--- a/server/entity/MediaRequest.ts
+++ b/server/entity/MediaRequest.ts
@@ -26,6 +26,7 @@ import {
   Column,
   CreateDateColumn,
   Entity,
+  JoinColumn,
   ManyToOne,
   OneToMany,
   PrimaryGeneratedColumn,
@@ -377,6 +378,7 @@ export class MediaRequest {
     eager: true,
     onDelete: 'CASCADE',
   })
+  @JoinColumn()
   public media: Media;
 
   @ManyToOne(() => User, (user) => user.requests, {
@@ -1187,3 +1189,5 @@ export class MediaRequest {
     }
   }
 }
+
+export default MediaRequest;
--- a/server/entity/MediaRequest.ts
+++ b/server/entity/MediaRequest.ts
@@ -26,7 +26,6 @@ import {
   Column,
   CreateDateColumn,
   Entity,
-  JoinColumn,
   ManyToOne,
   OneToMany,
   PrimaryGeneratedColumn,
@@ -378,7 +377,6 @@ export class MediaRequest {
     eager: true,
     onDelete: 'CASCADE',
   })
-  @JoinColumn()
   public media: Media;
 
   @ManyToOne(() => User, (user) => user.requests, {
--- a/server/api/servarr/sonarr.ts
+++ b/server/api/servarr/sonarr.ts
@@ -1,7 +1,7 @@
 import logger from '@server/logger';
 import ServarrBase from './base';
 
-interface SonarrSeason {
+export interface SonarrSeason {
   seasonNumber: number;
   monitored: boolean;
   statistics?: {
--- a/server/entity/Media.ts
+++ b/server/entity/Media.ts
@@ -114,29 +114,29 @@ class Media {
   @Column({ type: 'datetime', nullable: true })
   public mediaAddedAt: Date;
 
-  @Column({ nullable: true })
-  public serviceId?: number;
+  @Column({ nullable: true, type: 'text' })
+  public serviceId?: number | null;
 
-  @Column({ nullable: true })
-  public serviceId4k?: number;
+  @Column({ nullable: true, type: 'text' })
+  public serviceId4k?: number | null;
 
-  @Column({ nullable: true })
-  public externalServiceId?: number;
+  @Column({ nullable: true, type: 'text' })
+  public externalServiceId?: number | null;
 
-  @Column({ nullable: true })
-  public externalServiceId4k?: number;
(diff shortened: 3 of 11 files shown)


server/entity/MediaRequest.ts
```diff
@@ -377,6 +378,7 @@ export class MediaRequest {
     eager: true,
     onDelete: 'CASCADE',
   })
+  @JoinColumn()
```
sct, line 381: Does this not make a change to the database? Not sure why this wasn't needed until now?

OwsleyJr: It shouldn't be necessary if specify the relation when we try to find the correct media. Will update in the code.

server/entity/MediaRequest.ts
```diff
@@ -1187,3 +1187,5 @@ export class MediaRequest {
     }
   }
 }
+
+export default MediaRequest;
```
sct, line 1191: Why did we add this default export?

OwsleyJr: To keep consistency with the other entities that had default exports.

MattKobayashi:
What's holding this one up? Anything I can do to help get it across the line?

OwsleyJr:
> What's holding this one up? Anything I can do to help get it across the line?

Nothing to be done at this point. Just needs to be reviewed.

sct (approved):

sct:
@holopin-bot @jariz pin for you!

sct:
@holopin-bot @Nimelrian and you too!

sct:
@all-contributors please add @Nimelrian and @jariz for code

sct:
@all-contributors please add @jariz for code

epitsi:
This does not work for me. I tried setting the log level to DEBUG, did a full library scan, and searched the logs for the movie that I'd expect to be marked as unavailable after this fix, but I could find no such events. Anything I can do to help debug this? I'm running 1.33.0 via Docker.

MattKobayashi:
@epitsi Did you run the media availability sync task?

image

epitsi:
Ooops sorry. I thought this would run as part of the full library scan.

So I just ran the "Media Availability Sync" task (even though the Docker image has been on 1.33 for a few days now so it has been running automatically anyway). Checking the logs, I don't see any errors, other than a bunch of messages that look like this "1692 exists in at least one Radarr or Plex instance. Media will be updated if set to available". Not sure how to get more info for the specific movie I know it's not being handled properly.

MattKobayashi:
@epitsi If the movie still exists in either Plex or Radarr (regardless of whether the actual file is present), the Media Availability Sync task won't remove the movie from Overseerr. If you want the movie to be removed by the task, you'll have to make sure it's definitely gone from both Plex and Radarr.

What movie is it that's giving you issues? I've seen some rare instances where a media item is deleted from one of the upstream databases, consequently causing issues with various tools not being able to reconcile the local presence of a media item with the lack of an upstream database entry.

epitsi:
The movie is Fast & Furious 6 (tmdb id 82992). I've quadruple-checked in both Radarr and Plex, and neither have a listing for it.

MattKobayashi:
@epitsi I just tried to replicate the issue on my own instance, but could not. In my case, deleting the movie then running the `Media Availability Sync` task removed the request/availability data from Overseerr.

Might I suggest the following troubleshooting:

- Add the movie directly in Radarr (no need to start a search for it if you don't want to grab it again)
- Delete the movie in Radarr, ensuring you check the `Delete Movie Folder` checkbox like so:
image
- Run the `Media Availability Sync` task in Overseerr again.

This should clear out any 'stuck' records in Radarr/Plex for the movie.

PR #3244, opened 2023-01-14
fix(request): mark request as approved if media is already available when retrying failed request

danshilm:
#### Description

Failed requests whose media have become available while the request wasn't retried would fail with the error message `Media already available` when retried. This changes that behaviour to mark the request as approved instead.

N/A

#### To-Dos

#### Issues Fixed or Closed

- Fixes none

### DIFF
--- a/server/entity/MediaRequest.ts
+++ b/server/entity/MediaRequest.ts
@@ -767,7 +767,7 @@ export class MediaRequest {
         if (
           media[this.is4k ? 'status4k' : 'status'] === MediaStatus.AVAILABLE
         ) {
-          logger.warn('Media already available, marking request as APPROVED', {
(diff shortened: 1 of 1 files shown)


server/entity/MediaRequest.ts
```diff
@@ -767,7 +767,16 @@ export class MediaRequest {
         if (
           media[this.is4k ? 'status4k' : 'status'] === MediaStatus.AVAILABLE
         ) {
-          throw new Error('Media already available');
+          logger.warn('Media already available, marking request as APPROVED', {
```
sct, line 770: Maybe we should change the wording to "Media already exists in Radarr" since it may not actually be fully available?

danshilm: Sounds good!

sct (approved):

PR #3275, opened 2023-01-25
fix: added a refresh interval if download status is in progress

### LINKED ISSUE #3270: Status badge shows inaccurate progress when updating without page refresh
### Description

When viewing the download progress of a request via the *Status* badge, and the progress live-updates (without refreshing the page) the progress percentage has increased by the wrong amount. If you then immediately refresh the page, the progress percentage is correct again.

### Version

1.32.0

### Steps to Reproduce

1. Request a season of a TV show
2. Wait for Overseerr to get the download progress from Sonarr and show it on the *Status* badge
3. Wait for Overseerr to update the progress percentage
4. Check progress against what's shown in the Sonarr UI at that time and note that they're not the same
5. Reload Overseerr UI (normal F5 refresh)
6. Check progress against what's shown in the Sonarr UI at that time and note that they're now exactly the same

### Screenshots

_No response_

### Logs

_No response_

### Platform

desktop

### Device

PC, OnePlus 9 Pro

### Operating System

Windows 11, Android 13

### Browser

Edge

### Additional Context

I'm running Overseerr and all other apps as Docker Compose services, they are communicating over a local Docker network. (http)

### Code of Conduct

OwsleyJr:
#### Description

Progress bar would not fetch new download status progress if page was not refocused or reloaded. To solve this, I added a refreshInterval if there is a download to watch. I also added in a change to make the download sync editable, in case you want to grab the download status at a faster rate.

- Each component we will call refreshIntervalHelper to see if downloadStatus contains items and then update the timer accordingly.
- Timer set to 5000ms if items are present but this can be altered if needed.
- Download sync schedule is now editable. Editable times are 30, 45, and 60 seconds. 

#### To-Dos

#### Issues Fixed or Closed

- Fixes #3270

### DIFF
--- a/server/job/schedule.ts
+++ b/server/job/schedule.ts
@@ -14,7 +14,7 @@ interface ScheduledJob {
   job: schedule.Job;
   name: string;
   type: 'process' | 'command';
-  interval: 'short' | 'long' | 'fixed';
+  interval: 'seconds' | 'minutes' | 'hours' | 'fixed';
   cronSchedule: string;
   running?: () => boolean;
   cancelFn?: () => void;
@@ -30,7 +30,7 @@ export const startJobs = (): void => {
     id: 'plex-recently-added-scan',
     name: 'Plex Recently Added Scan',
     type: 'process',
-    interval: 'short',
+    interval: 'minutes',
     cronSchedule: jobs['plex-recently-added-scan'].schedule,
     job: schedule.scheduleJob(jobs['plex-recently-added-scan'].schedule, () => {
       logger.info('Starting scheduled job: Plex Recently Added Scan', {
@@ -47,7 +47,7 @@ export const startJobs = (): void => {
     id: 'plex-full-scan',
     name: 'Plex Full Library Scan',
     type: 'process',
-    interval: 'long',
+    interval: 'hours',
     cronSchedule: jobs['plex-full-scan'].schedule,
     job: schedule.scheduleJob(jobs['plex-full-scan'].schedule, () => {
       logger.info('Starting scheduled job: Plex Full Library Scan', {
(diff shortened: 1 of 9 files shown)


server/lib/settings.ts
```diff
@@ -410,7 +410,7 @@ class Settings {
           schedule: '0 30 4 * * *',
         },
         'download-sync': {
-          schedule: '0 * * * * *',
+          schedule: '*/60 * * * * *',
```
danshilm, line 413: This can be reverted back to 0 to match the the schedule of the other jobs.

src/components/MovieDetails/index.tsx
```diff
@@ -110,6 +111,13 @@ const MovieDetails = ({ movie }: MovieDetailsProps) => {
     mutate: revalidate,
   } = useSWR<MovieDetailsType>(`/api/v1/movie/${router.query.movieId}`, {
     fallbackData: movie,
+    refreshInterval: refreshIntervalHelper(
+      {
+        downloadStatus: movie?.mediaInfo?.downloadStatus,
+        downloadStatus4k: movie?.mediaInfo?.downloadStatus4k,
+      },
+      5000
```
danshilm, line 119: 5 seconds interval is way too fast considering that a user can only modify the job schedule to go as fast as 30 seconds in the UI. 
30 or even 15 seconds would be good, though I'm leaning towards the former since IMO a user would need to be actively looking at the progress bar and comparing it to their *arr instance to notice the difference between 30 seconds and 15 seconds interval.

OwsleyJr: I'm okay with changing it to 15 seconds. I feel like any higher might be too long especially since the default is 60.

server/lib/settings.ts
OwsleyJr: Ah, yep let me change that.

src/components/MovieDetails/index.tsx
danshilm: Sounds good 👍

server/job/schedule.ts
```diff
@@ -139,7 +139,7 @@ export const startJobs = (): void => {
     id: 'image-cache-cleanup',
     name: 'Image Cache Cleanup',
     type: 'process',
-    interval: 'long',
+    interval: 'minutes',
```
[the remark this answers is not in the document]
OwsleyJr: Not an intended change, I think the comment above it threw me off while I was changing them. Will fix and push.

PR #3332, opened 2023-02-11
perf(server/index.ts): do not set cookies to image proxy so CDNs can cache images

lunks:
This is only a problem if CSRF protection is enabled.

#### Description

CDNs like Cloudflare bypass their cache if cookies are set in the response.
clearCookies middleware removes the header before imageproxy serves the image.

I initially considered creating a `middlewares` collection and removing CSRF protection from the list if it is present for the image proxy, but removing the header is a much simpler solution.

#### To-Dos

#### Issues Fixed or Closed

_none_

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -17,6 +17,7 @@ import WebhookAgent from '@server/lib/notifications/agents/webhook';
 import WebPushAgent from '@server/lib/notifications/agents/webpush';
 import { getSettings } from '@server/lib/settings';
 import logger from '@server/logger';
+import clearCookies from '@server/middleware/clearcookies';
 import routes from '@server/routes';
 import imageproxy from '@server/routes/imageproxy';
 import { getAppVersion } from '@server/utils/appVersion';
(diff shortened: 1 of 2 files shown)


sct:
Curious why you are having cookies sent with your image proxy. I seem unable to reproduce. My images served through Cloudflare are returning hits for the cache.

lunks:
> Curious why you are having cookies sent with your image proxy. I seem unable to reproduce. My images served through Cloudflare are returning hits for the cache.

If you enable CSRF, it gets set on every request. Cloudflare works fine if I disable CSRF, but I want the extra security the CSRF token provides.

sct:
Ah right CSRF. I see. Got it.

sct (approved):

sct:
@all-contributors please add @lunks for code

sct:
@holopin-bot @lunks lets get a badge here!

lunks:
Thank you for merging my PR and for the awesome project :)

PR #3361, opened 2023-02-25
feat: adding streaming services as a slider is now an option

OwsleyJr:
#### Description

Thought streaming services should be available to add as a slider so I did just that. Works similar to the filterSlideover. You will be able to select the region and provider(s) to your liking. Included a screenshot.

As a footnote, this also includes a change to the useEffect included in the watchProviderSelector. The onChange was removed as a dependency due to Formik causing an infinite loop when setting the field value.

Screenshot 2023-02-24 at 5 11 33 PM

#### To-Dos


src/components/Selector/index.tsx
```diff
@@ -305,7 +305,7 @@ export const WatchProviderSelector = ({
 
   useEffect(() => {
     onChange(watchRegion, activeProvider);
-  }, [activeProvider, watchRegion, onChange]);
+  }, [activeProvider, watchRegion]);
```
sct, line 308: This probably has a linter warning above it. We should add a comment disabling eslint on this line and explain why we disabled it.

OwsleyJr: Added.

src/components/Discover/DiscoverSliderEdit/index.tsx
```diff
@@ -195,7 +199,9 @@ const DiscoverSliderEdit = ({
           className={`${slider.data ? 'mb-4' : 'mb-0'} flex space-x-2 md:mb-0`}
         >
           <Bars3Icon className="h-6 w-6" />
-          <div>{getSliderTitle(slider)}</div>
+          <div className="w-7/12 truncate md:w-full">
```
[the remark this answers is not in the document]
OwsleyJr: I set a specific width because we need to truncate before the chevron arrows.

sct (approved):

Issue #3378, opened 2023-03-05
500 - Internal Server Error


nrgbistro:
### Description

When I try to search for certain words, for example "cocaine", I get 500 - Internal Server Error instead of search results. Is this some kind of language filtering? If so, is it possible to disable in Overseerr?

### Version

1.32.5

### Steps to Reproduce

1. Login to Overseerr
2. Type in "cocaine" in the search bar.
3. See that you get "500 - Internal Server Error" instead of search results

### Screenshots

![image](https://user-images.githubusercontent.com/35301702/222987189-411e443e-1ca9-4693-8e54-524e33f7af54.png)
![image](https://user-images.githubusercontent.com/35301702/222987200-c20e4e5b-94db-404a-ba7b-dc4571f1fab1.png)
reloading the page with "cocaine" in the search bar as the search query:
![image](https://user-images.githubusercontent.com/35301702/222987313-7813cd67-5a04-4954-9fcf-431bfef9667d.png)

### Platform

desktop

### Device

Desktop

### Operating System

Windows 11

### Browser

Brave

### Code of Conduct

- [X] I agree to follow Overseerr's Code of Conduct

kevin-wynn:
This started happening to me as well. Was happening on v1.32.4 so I upgraded to v1.32.5 and still getting the error. This is the error I get in logs ```2023-04-05T18:56:38.864Z [debug][API]: Something went wrong retrieving search results{"errorMessage":"Cannot read properties of undefined (reading 'map')","query":"rocky"}```

nrgbistro:
Can confirm that "rocky" causes the issue on my end as well.

kevin-wynn:
This seems to have been resolved for me with no action on my end /shrug

RantMaster:
I am seeing this when searching for Metalocalypse. It seems to be when you get 5 or more characters in some words it gives the error

Alexays:
My PR (https://github.com/sct/overseerr/pull/3393) should fix this issue.

teuchezh:
same issue with this log
`[debug][API]: Something went wrong retrieving trending items {"errorMessage":"[TMDB] Failed to fetch all trending: connect ECONNREFUSED 127.0.0.1:443"}`
![image](https://user-images.githubusercontent.com/40827089/233155480-af7d3b04-20c3-4802-b25b-02e2078972ef.png)
[...]

PR #3393, opened 2023-03-19
Fix 500 & handle search of collection

Alexays:
#### Description

When searching for "ce soir c'", there's a 500 error, because tmdb return a 'collection' type that is not handled.

Before:
image
After:
image

#### To-Dos

#### Issues Fixed or Closed

- #3378

### DIFF
--- a/server/utils/typeHelpers.ts
+++ b/server/utils/typeHelpers.ts
@@ -35,7 +35,7 @@ export const isCollection = (
     | TmdbPersonResult
     | TmdbCollectionResult
 ): collection is TmdbCollectionResult => {
-  return (collection as TmdbCollectionResult).title !== undefined;
+  return (collection as TmdbCollectionResult).media_type === 'collection';
 };
 
 export const isMovieDetails = (
--- a/server/utils/typeHelpers.ts
+++ b/server/utils/typeHelpers.ts
@@ -1,4 +1,5 @@
 import type {
+  TmdbCollectionResult,
   TmdbMovieDetails,
   TmdbMovieResult,
   TmdbPersonDetails,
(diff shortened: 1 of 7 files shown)


sct:
Thanks for the PR! Sorry, I have been crazy busy the last few weeks (and also traveling) but I will get to this PR soon! Didn't want to leave you wondering if this would ever be looked at. 🙏

src/components/TitleCard/index.tsx
```diff
@@ -130,14 +138,16 @@ const TitleCard = ({
...
+                mediaType === 'movie' || mediaType === 'collection'
                   ? 'border-blue-500 bg-blue-600'
                   : 'border-purple-600 bg-purple-600'
               }`}
             >
               <div className="flex h-4 items-center px-2 py-2 text-center text-xs font-medium uppercase tracking-wider text-white sm:h-5">
                 {mediaType === 'movie'
                   ? intl.formatMessage(globalMessages.movie)
+                  : mediaType === 'collection'
+                  ? intl.formatMessage(globalMessages.movies)
```
TheCatLady, line 150: I think the badge should say "Collection" rather than "Movies" here -- there are languages where nouns don't have plural forms. Otherwise, this PR looks good to me!

server/utils/typeHelpers.ts
```diff
@@ -8,17 +9,35 @@ import type {
...
 };
 
+export const isCollection = (
+  collection:
+    | TmdbMovieResult
+    | TmdbTvResult
+    | TmdbPersonResult
+    | TmdbCollectionResult
+): collection is TmdbCollectionResult => {
+  return (collection as TmdbCollectionResult).title !== undefined;
```
danshilm, line 38: This won't work correctly since that condition is also valid for a `TmdbMovieResult` object. 

Unfortunately for us, all the properties of a `TmdbCollectionResult` are also present in a `TmdbMovieResult` and are of the same type - the only difference being the `media_type`. Hence, this should be updated to check the `media_type` instead.
```suggestion
  return (collection as TmdbCollectionResult).media_type === 'collection';
```

danshilm (changes requested):
Looks good!
Just one small nitpick

src/components/TitleCard/index.tsx
danshilm: Also, to me, the difference is not really noticeable since the colour of the badge is the same.

TheCatLady: I think keeping the same color makes sense, since there are really just two types of requests in Overseerr, movies and series.

sct (approved):

sct:
@all-contributors please add @Alexays for code

sct:
@holopin-bot @Alexays heres a badge! thanks again!

WashingtonMatt:
I'm experiencing this issue with "Riddick".

PR #3407, opened 2023-04-03
feat(discover): support filtering by tmdb user vote count on discover page

SMores:
#### Description

Adds a new MultiRangeSlider to FilterSlideover to support filtering by TMDB user vote count.

Sorting by TMDB Rating on the Discover pages currently isn't very useful, because there are _hundreds_ of TV series and movies that have exactly one perfect rating!

This allows users to make better use of that sort feature by filtering by vote count. This dramatically improves the discover results in my subjective opinion.

The maximum value was set somewhat arbitrarily, but it does actually seem like very few series or movies have more than 1000 user votes, soooo maybe it's a good value? I'd also be open to the idea that there's not much value in setting a max, and perhaps this shouldn't be a MultiRangeSlider at all, but just a simple slider (it just seemed nice to leave the possibility there in case someone wanted to look for content that was less mainstream).

image

#### To-Dos

- [ ] ~Database migration (if required)~

#### Issues Fixed or Closed

I haven't opened an Issue for this, but I'm happy to if it helps with tracking!


sct:
Thanks so much for the contribution! I will try to review this as soon as I have a little free time!

sct (approved):
This is great. Thank you so much.

sct:
@all-contributors please add @SMores for code

sct:
@holopin-bot @SMores heres a badge!

PR #3411, opened 2023-04-06
feat: pwa badge notifications

OwsleyJr:
#### Description

This pull request implements the app badging API, enabling the addition of a badge to the app while using the PWA. The badge will display the number of pending requests and will only be visible to admin users.

- Utilizes the navigator object to set the app badge.
- Created a specific notification payload only for admins that will tell the service worker to update the badge.

Ignore the icon logo as it was for testing purposes.

IMG_5220

#### To-Dos

### DIFF
--- a/public/sw.js
+++ b/public/sw.js
@@ -3,7 +3,7 @@
 // previously cached resources to be updated from the network.
 // This variable is intentionally declared and unused.
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-const OFFLINE_VERSION = 3;
+const OFFLINE_VERSION = 4;
 const CACHE_NAME = 'offline';
 // Customize this with a different URL if needed.
 const OFFLINE_URL = '/offline.html';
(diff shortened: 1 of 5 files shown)


server/entity/MediaRequest.ts
```diff
@@ -1104,11 +1110,9 @@ export class MediaRequest {
       switch (type) {
         case Notification.MEDIA_APPROVED:
           event = `${this.is4k ? '4K ' : ''}${mediaType} Request Approved`;
-          notifyAdmin = false;
```
TheCatLady, line 1107: The `notifyAdmin` value is set to `false` here as this notification type doesn't pertain to admin users, but is of interest to the requesting user.

server/entity/MediaRequest.ts
```diff
@@ -1137,7 +1141,9 @@ export class MediaRequest {
           request: this,
           notifyAdmin,
           notifySystem,
-          notifyUser: notifyAdmin ? undefined : this.requestedBy,
+          notifyUser: Notification.MEDIA_AUTO_REQUESTED
+            ? this.requestedBy
+            : undefined,
```
TheCatLady, line 1146: This will impact the notification logic as well. When this is set, the users are sent a notification.

TheCatLady (changes requested):
If we need to always send certain web push events, I think that logic should be added to the web push agent. Modifying the `notifyAdmin`/`notifyUser` values will result in undesired notifications being sent for all agents.

Additionally, we don't want to display web push notifications for events where we currently aren't sending anything... we just want to update the badge?

server/entity/MediaRequest.ts
OwsleyJr: I pushed a new change that should correct this. Let me know what you think.

OwsleyJr:
> If we need to always send certain web push events, I think that logic should be added to the web push agent. Modifying the `notifyAdmin`/`notifyUser` values will result in undesired notifications being sent for all agents.
> 
> Additionally, we don't want to display web push notifications for events where we currently aren't sending anything... we just want to update the badge?

We need to send a web push notification to update the admin(s) badge. So now even on media approved and declined, the admin should receive a (silent) notification but it will be a custom payload for only setting the badge.

sct (approved):

sct (approved):

sct (approved):

PR #3412, opened 2023-04-07
build: create a devcontainer configuration 

MrDynamo:
#### Description

Creates a devcontainer configuration for remote development contained within a docker container. Useful as you don't need to install ANY dependencies on the local machine or WSL environment. Keeps your environment for Overseerr development contained to a container.

#### To-Dos


MrDynamo:
Current implementation requires running 'yarn dev' in terminal after the container is created. This can be switched to auto run if wanted.

PR #3500, opened 2023-06-12
fix(web/discover): filter duplicates in pagination

### LINKED ISSUE #3499: TMDB recommendations have duplicates
### Description

The last element of the first movie recommendations' page is the same of the next pagination. Root cause is the TMDB API's response.

There are two ways to solve this:
1. return a page of 19 from the backend (there's no way to check with previous page, without caching or network calls),
2. (if found to be the same) skip the first element of the next page whilst stitching together the pages in the frontend.

Both have pros and cons, I'd implement a PR for the second one.

### Version

1.4.1

### Steps to Reproduce

Fetch the recommendations API for a TMDB movie, look at the last recommendation of the first page. It's identical to the first recommendation of the second page.

### Screenshots

immagine

### Logs

_No response_

### Platform

desktop

### Device

any

### Operating System

any

### Browser

any

### Additional Context

Coming from [downstream jellyseerr issue](https://github.com/Fallenbagel/jellyseerr/issues/405)

### Code of Conduct

marcofaggian:
#### Description

Duplicates coming directly from the TMDB API, filtered at rendering time (within `useDiscover`) so that a cached state at the backend is not needed.

Before
immagine

After
immagine

#### To-Dos

#### Issues Fixed or Closed

- Fixes #3499

### DIFF
--- a/src/hooks/useDiscover.ts
+++ b/src/hooks/useDiscover.ts
@@ -95,6 +95,11 @@ const useDiscover = <
 
   let titles = (data ?? []).reduce((a, v) => [...a, ...v.results], [] as T[]);
 
+  // Check for duplicates, see https://github.com/sct/overseerr/issues/3499
+  titles = (titles as any[]).filter(
+    (value, index, self) => index === self.findIndex((t) => t.id === value.id)
+  );
+
   if (settings.currentSettings.hideAvailable && hideAvailable) {
     titles = titles.filter(
       (i) =>


sct:
The performance impact of code like this is too costly for what is, at most, a very minor (and inconsistent) inconvenience of seeing double.

sct:
I think it's the same deal. It doesn't really make sense to me to introduce this for such a minor inconvenience. And even though it's only 40 items, id rather not introduce code that's looping through each item X number of times for a problem that most of the time doesn't even appear.

PR #3516, opened 2023-06-22
fix: handle issue causing incorrect media to change to unknown

OwsleyJr:
#### Description

This pull request includes improved handling of certain edge cases where media statuses were being changed to unknown and incorrect requests being were removed. Also included is a minor fix for users with a single radarr or sonarr server.

This is a temporary fix at least until we add the new media and request statuses.

- Prevent the incorrect media being set to unknown on rare occasions.
- Enhanced request handling.
- Better checks for users with single *arr servers.

N/A

#### To-Dos

### DIFF
--- a/server/lib/availabilitySync.ts
+++ b/server/lib/availabilitySync.ts
@@ -1,14 +1,13 @@
-import type { PlexLibraryItem, PlexMetadata } from '@server/api/plexapi';
+import type { PlexMetadata } from '@server/api/plexapi';
 import PlexAPI from '@server/api/plexapi';
-import type { RadarrMovie } from '@server/api/servarr/radarr';
-import RadarrAPI from '@server/api/servarr/radarr';
+import RadarrAPI, { type RadarrMovie } from '@server/api/servarr/radarr';
 import type { SonarrSeason, SonarrSeries } from '@server/api/servarr/sonarr';
 import SonarrAPI from '@server/api/servarr/sonarr';
 import { MediaRequestStatus, MediaStatus } from '@server/constants/media';
 import { getRepository } from '@server/datasource';
 import Media from '@server/entity/Media';
 import MediaRequest from '@server/entity/MediaRequest';
-import Season from '@server/entity/Season';
+import type Season from '@server/entity/Season';
 import SeasonRequest from '@server/entity/SeasonRequest';
 import { User } from '@server/entity/User';
 import type { RadarrSettings, SonarrSettings } from '@server/lib/settings';
@@ -18,10 +17,6 @@ import logger from '@server/logger';
 class AvailabilitySync {
   public running = false;
   private plexClient: PlexAPI;
-  private radarrClient: RadarrAPI;
-  private radarrClient4k: RadarrAPI;
-  private sonarrClient: SonarrAPI;
-  private sonarrClient4k: SonarrAPI;
   private plexSeasonsCache: Record<string, PlexMetadata[]> = {};
   private sonarrSeasonsCache: Record<string, SonarrSeason[]> = {};
   private radarrServers: RadarrSettings[];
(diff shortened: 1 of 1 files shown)


server/lib/availabilitySync.ts
```diff
@@ -57,141 +54,93 @@ class AvailabilitySync {
...
+
+          // Check if a season is processing or pending to
+          // make sure we set media to the correct status
+          let mediaStatus = MediaStatus.UNKNOWN;
+          let mediaStatus4k = MediaStatus.UNKNOWN;
+
+          if (media.mediaType === 'tv') {
+            mediaStatus = await this.findMediaStatus(requests, false);
+            mediaStatus4k = await this.findMediaStatus(requests, true);
+          }
```
OwsleyJr, line 89: Switched to using query builder to make sure requests are related to only available media. We then check those requests to set the media to the correct status. For example, if the media is deleted and a request still remains, the media should be set to processing.

server/lib/availabilitySync.ts
```diff
@@ -57,141 +54,93 @@ class AvailabilitySync {
...
+                  ? media.externalServiceId
+                  : null),
+              (media.externalServiceSlug =
+                mediaStatus === MediaStatus.PROCESSING
+                  ? media.externalServiceSlug
+                  : null),
+              (media.ratingKey =
+                mediaStatus === MediaStatus.PROCESSING
+                  ? media.ratingKey
+                  : null);
```
OwsleyJr, line 111: Switched to using query builder to make sure requests are related to only available media. We then check those requests to set the media to the correct status. For example, if the media is deleted and a request still remains, the media should be set to processing.

server/lib/availabilitySync.ts
```diff
@@ -57,141 +54,93 @@ class AvailabilitySync {
...
+                  : null),
+              (media.externalServiceSlug4k =
+                mediaStatus === MediaStatus.PROCESSING
+                  ? media.externalServiceSlug4k
+                  : null),
+              (media.ratingKey4k =
+                mediaStatus === MediaStatus.PROCESSING
+                  ? media.ratingKey4k
+                  : null);
+          }
```
OwsleyJr, line 135: Switched to using query builder to make sure requests are related to only available media. We then check those requests to set the media to the correct status. For example, if the media is deleted and a request still remains, the media should be set to processing.

server/lib/availabilitySync.ts
```diff
@@ -57,141 +54,93 @@ class AvailabilitySync {
...
-                  }
-                );
-
-                if (seasonToBeDeleted) {
-                  await seasonRequestRepository.remove(seasonToBeDeleted);
-                }
-
-                didDeleteSeasons = true;
-              }
-            }
```
OwsleyJr, line 170: All season handling has been moved to the existsInSonarr function. This simplifies the flow of checking through each show.

server/lib/availabilitySync.ts
```diff
@@ -57,141 +54,93 @@ class AvailabilitySync {
...
-                  });
-                }
-
-                if (media.status4k === MediaStatus.AVAILABLE) {
-                  await mediaRepository.update(media.id, {
-                    status4k: MediaStatus.PARTIALLY_AVAILABLE,
-                  });
-                }
-              }
-            }
```
OwsleyJr, line 194: This has also been moved along with the season checker as we still need to check for season removal to change the media status to partially available.

server/lib/availabilitySync.ts
```diff
@@ -234,62 +183,115 @@ class AvailabilitySync {
...
+      )
+    ) {
+      mediaStatus = MediaStatus.PENDING;
+    } else {
+      mediaStatus = MediaStatus.UNKNOWN;
+    }
+
+    return mediaStatus;
+  }
+
```
OwsleyJr, line 216: New reusable function created which will sort through the requests to see if any are still approved or pending. This is used on the top level of the mediaExists and mediaUpdater function.

server/lib/availabilitySync.ts
```diff
@@ -234,62 +183,115 @@ class AvailabilitySync {
...
-      const request = await requestRepository.findOne({
-        relations: {
-          media: true,
-        },
-        where: { media: { id: media.id }, is4k: is4k ? true : false },
-      });
-
       logger.info(
-        `Media ID ${media.id} does not exist in your ${
+        `Media with TMDB ID ${media.tmdbId} does not exist in your ${
```
OwsleyJr, line 223: All logs that showed media id will now show the tmdbId instead. This will allow better debugging.

server/lib/availabilitySync.ts
```diff
@@ -234,62 +183,115 @@ class AvailabilitySync {
...
+              externalServiceId:
+                mediaStatus === MediaStatus.PROCESSING
+                  ? media.externalServiceId
+                  : null,
+              externalServiceSlug:
+                mediaStatus === MediaStatus.PROCESSING
+                  ? media.externalServiceSlug
+                  : null,
+              ratingKey:
+                mediaStatus === MediaStatus.PROCESSING ? media.ratingKey : null,
```
OwsleyJr, line 281: Same thing we are doing before which is to make sure we are setting the media to the correct status if requests remain.

server/lib/availabilitySync.ts
```diff
@@ -312,67 +316,49 @@ class AvailabilitySync {
...
-          meta = await api.getMovie({ id: media.externalServiceId4k });
-        }
-
-        if (!server.is4k && (!meta || !meta.hasFile)) {
-          existsInRadarr = false;
-        }
-
-        if (server.is4k && (!meta || !meta.hasFile)) {
-          existsInRadarr4k = false;
+          radarr4k = await api.getMovie({ id: media.externalServiceId4k });
```
OwsleyJr, line 324: For the radarr and sonarr functions, the returned data has been moved to outside of the for loop for better checking if the movie/show exists. This also prevents a bug that was causing single server users to show that they had a 4k server or vise versa.

server/lib/availabilitySync.ts
```diff
@@ -395,72 +383,70 @@ class AvailabilitySync {
...
     ) {
-      if (media.status4k !== MediaStatus.UNKNOWN) {
-        this.mediaUpdater(media, true);
+      showExists4k = false;
+    }
+
+    // Here we check each season for availability
+    for (const season of media.seasons) {
+      await this.seasonExists(media, season, showExists, showExists4k);
+    }
```
OwsleyJr, line 427: The season checker function has been moved here. Explanation is at the top.

server/lib/availabilitySync.ts
```diff
@@ -567,124 +548,158 @@ class AvailabilitySync {
...
+            (season.status === MediaStatus.AVAILABLE ||
+              season.status === MediaStatus.PARTIALLY_AVAILABLE)) ||
+          (seasonRequest.request.is4k &&
+            !seasonExists4k &&
+            (season.status4k === MediaStatus.AVAILABLE ||
+              season.status4k === MediaStatus.PARTIALLY_AVAILABLE))
+      );
+
+      let deletedSeason = false;
+
```
OwsleyJr, line 564: Similar concept to our media requests checker. This will go through each season and make sure that season is available to prevent the wrong request deletion.

server/lib/availabilitySync.ts
```diff
@@ -694,7 +709,7 @@ class AvailabilitySync {
       // If true, media exists in at least one radarr or plex instance.
       if (existsInRadarr) {
         logger.warn(
-          `${media.id} exists in at least one Radarr or Plex instance. Media will be updated if set to available.`,
```
OwsleyJr, line 697: Removed second sentence as it can be confusing to the user.

server/lib/availabilitySync.ts
```diff
@@ -727,74 +744,77 @@ class AvailabilitySync {
...
-        `Season ${season.seasonNumber}, media ID ${media.id} exists in at least one Sonarr or Plex instance. Media will be updated if set to available.`,
-        {
-          label: 'AvailabilitySync',
-        }
-      );
-
-      return true;
-    }
-
-    return false;
```
OwsleyJr, line 797: We do not have to return a boolean here anymore as we have now nested the season checker within the sonarr checker.

server/lib/availabilitySync.ts
```diff
@@ -31,21 +31,26 @@ class AvailabilitySync {
...
+        label: 'AvailabilitySync',
+      });
 
-      if (!this.plexClient) {
-        return;
-      }
+      return;
+    }
 
+    try {
```
OwsleyJr, line 48: Added a separate try catch to make sure we can access the Plex server. If the call fails, we will return and stop the loop.

server/lib/availabilitySync.ts
```diff
@@ -31,21 +35,65 @@ class AvailabilitySync {
...
+            label: 'AvailabilitySync',
+          }
+        );
+        return;
+      }
+    } else {
+      logger.info(
+        'Could not find a Sonarr server with sync enabled. Skipping Sonarr detection.'
+      );
+    }
```
OwsleyJr, line 89: This will initialize your plex, radarr, and sonarr clients to prevent the scan from running if the status check fails.

server/lib/availabilitySync.ts
```diff
@@ -301,78 +354,78 @@ class AvailabilitySync {
...
+      }
 
-        let meta: RadarrMovie | undefined;
+      if ((!radarr || !radarr.hasFile) && !existsInPlex) {
+        movieExists = false;
+      }
+    } catch (ex) {
+      if (ex.message.includes('404')) {
+        movieExists = false;
+      }
```
OwsleyJr, line 377: Set the movie to not existing only if the catch returns a 404.

OwsleyJr: Added this to all of the Radarr/Sonarr API calls.

server/lib/availabilitySync.ts
```diff
@@ -727,89 +867,167 @@ class AvailabilitySync {
...
+
+      if (!server.is4k) {
+        await radarrAPI.getSystemStatus();
+        this.radarrClient = radarrAPI;
+      }
+
+      if (server.is4k) {
+        await radarrAPI.getSystemStatus();
+        this.radarrClient4k = radarrAPI;
+      }
```
sct, line 1008: Won't this override existing `radarrClient` properties if there are multiple 4k or non-4k servers?

sct (approved):

PR #3526, opened 2023-06-29
fix(statusbadge): handle missing season/episode number

TheCatLady:
#### Description

Minor bug fix to handle cases where the download item does not have a season/episode number.

Current behavior:
image

#### To-Dos


OwsleyJr (approved):
LGTM!

PR #3551, opened 2023-07-15
fix: importing friends

### LINKED ISSUE #3550: Error checking user access: This user does not exist on the main Plex account's shared list
### Description

I had a Plex server set up with user access. I changed the server and set up a new server.
I gave access to this new server to users who already had access to the old server and to new users who only had access to this new server.

Overseerr I set it up with only the new server and when I import users from Plex (or when they try to log in) I only get the users that had access to the old server, the new users do not appear in the list and I get the following error:

`Error checking user access: This user does not exist on the main Plex account's shared list`

I tried the answers in issue #3068 but it doesn't work. 

### Version

1.33.0

### Steps to Reproduce

1. I click on Users.
2. I click on "import Plex Users".
3. Only the users that had access to the old server that no longer exists are listed.
4. The logs show the above error, one for each user.

### Screenshots

_No response_

### Logs

```shell
[error]: Error checking user access: This user does not exist on the main Plex account's shared list 
[error]: Error checking user access: This user does not exist on the main Plex account's shared list
```

### Platform

desktop

### Device

Desktop

### Operating System

Docker

### Browser

Vivaldi

### Additional Context

_No response_

### Code of Conduct

scorp200:
update checkUserAccess to use getUsers

#### Description

getFriends() calls `pms/friends/all` which is deprecated.
getUsers() should be used instead.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #3550

### DIFF
--- a/server/api/plextv.ts
+++ b/server/api/plextv.ts
@@ -234,6 +234,9 @@ class PlexTvAPI extends ExternalAPI {
     }
   }
 
+  /**
+   * @deprecated Plex has deprecated '/pms/friends/all', use getUsers(); instead.
+   */
   public async getFriends(): Promise<FriendResponse> {
     const response = await this.axios.get('/pms/friends/all', {
       transformResponse: [],
@@ -255,7 +258,7 @@ class PlexTvAPI extends ExternalAPI {
         throw new Error('Plex is not configured!');
       }
 
-      const friends = await this.getFriends();
+      const friends = await this.getUsers();
 
       const users = friends.MediaContainer.User;
 
--- a/server/api/plextv.ts
+++ b/server/api/plextv.ts
@@ -82,21 +82,6 @@ interface ServerResponse {
   };
 }
 
-interface FriendResponse {
-  MediaContainer: {
-    User: {
-      $: {
-        id: string;
-        title: string;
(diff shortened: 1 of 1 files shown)


server/api/plextv.ts
```diff
@@ -234,6 +234,9 @@ class PlexTvAPI extends ExternalAPI {
     }
   }
 
+  /**
+   * @deprecated Plex has deprecated '/pms/friends/all', use getUsers(); instead.
+   */
```
OwsleyJr, line 239: We can just remove getFriends as it will no longer be used instead of adding this comment.

server/api/plextv.ts
```diff
@@ -255,7 +258,7 @@ class PlexTvAPI extends ExternalAPI {
         throw new Error('Plex is not configured!');
       }
 
-      const friends = await this.getFriends();
+      const friends = await this.getUsers();
```
OwsleyJr, line 261: This variable should be renamed.

OwsleyJr (approved):
Looks good! Thanks for the PR

OwsleyJr:
@all-contributors please add @scorp200 for code

PR #3620, opened 2023-09-19
feat: add hideTags option in user settings

### LINKED ISSUE #3350: Hide tags from detail page
### Description

The ability to hide tags from the detail page

Apart from it not being potentially useful to some users and causing visual clutter (especially on mobile screens), it also gives away spoilers for the content.

image

### Desired Behavior

The ability to hide tags

### Additional Context

_No response_

### Code of Conduct

andromidasj:
#### Description
This new option will allow for a user to hide the tags that show in a Movie or TV Show details page. If the setting `Hide Tags` is checked, the tags will not show. If it is unchecked, the tags will show.

Setting:
image

Before:
image

After:
image

#### To-Dos

#### Issues Fixed or Closed

- Fixes #3350

### DIFF
--- a/src/hooks/useUser.ts
+++ b/src/hooks/useUser.ts
@@ -1,5 +1,4 @@
 import { UserType } from '@server/constants/user';
-import type { UserSettingsGeneralResponse } from '@server/interfaces/api/userSettingsInterfaces';
 import type { PermissionCheckOptions } from '@server/lib/permissions';
 import { hasPermission, Permission } from '@server/lib/permissions';
 import type { NotificationAgentKey } from '@server/lib/settings';
@@ -26,14 +25,15 @@ export interface User {
 
 type NotificationAgentTypes = Record<NotificationAgentKey, number>;
 
-type UserSettings = Pick<
-  UserSettingsGeneralResponse,
-  'discordId' | 'region' | 'originalLanguage' | 'locale' | 'hideTags'
-> & {
+export interface UserSettings {
+  discordId?: string;
+  region?: string;
+  originalLanguage?: string;
+  locale?: string;
   notificationTypes: Partial<NotificationAgentTypes>;
   watchlistSyncMovies?: boolean;
   watchlistSyncTv?: boolean;
-};
+}
 
 interface UserHookResponse {
   user?: User;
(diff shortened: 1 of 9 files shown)


src/hooks/useUser.ts
```diff
@@ -25,15 +26,14 @@ export interface User {
 
 type NotificationAgentTypes = Record<NotificationAgentKey, number>;
 
-export interface UserSettings {
-  discordId?: string;
-  region?: string;
-  originalLanguage?: string;
-  locale?: string;
+type UserSettings = Pick<
```
andromidasj, line 29: I noticed here that we're duplicating code and having multiple sources of truth. This Pick type allows us to honor the types already set for these settings in the UserSettingsGeneralResponse interface. If it's desired to have it typed out explicitly in this file though, we can simply remove what I added and instead add `hideTags?: boolean;` underneath line 32 of the original code.

src/hooks/useUser.ts
```diff
@@ -25,15 +26,14 @@ export interface User {
...
+  UserSettingsGeneralResponse,
+  'discordId' | 'region' | 'originalLanguage' | 'locale' | 'hideTags'
+> & {
   notificationTypes: Partial<NotificationAgentTypes>;
   watchlistSyncMovies?: boolean;
   watchlistSyncTv?: boolean;
-}
+};
```
andromidasj, line 36: I noticed here that we're duplicating code and having multiple sources of truth. This Pick type allows us to honor the types already set for these settings in the UserSettingsGeneralResponse interface. If it's desired to have it typed out explicitly in this file though, we can simply remove what I added and instead add `hideTags?: boolean;` underneath line 32 of the original code.

OwsleyJr: I don't think we should modify any older code unless necessary. The hideTags boolean should suffice.

andromidasj: Sounds good, I'll remove this chunk then!

andromidasj:
@OwsleyJr do you know how I can get help with data migration stuff? My understanding is pretty limited in that side of things, and there seems to be some SQL build errors in Cypress like:

```
driverError: Error: SQLITE_ERROR: no such column: User_settings.hideTags
```

OwsleyJr:
> @OwsleyJr do you know how I can get help with data migration stuff? My understanding is pretty limited in that side of things, and there seems to be some SQL build errors in Cypress like:
> 
> ```
> driverError: Error: SQLITE_ERROR: no such column: User_settings.hideTags
> ```

Would just need to generate a migration file. You can look in the package.json for the script we use.

PR #3628, opened 2023-09-27
feat: standard series type selector

OwsleyJr:
#### Description

The sonarr modal will now also include a standard series type selector which you can change from standard to daily. Similar to the new anime series type selector.

Screenshot 2023-09-27 at 8 19 42 AM

#### To-Dos


Gauvino:
Is it really necessary cause I don't think anyone wants to have their series or anything added as daily?

sct (approved):

PR #3648, opened 2023-10-16
fix(docs): correct openapi docs minor issues

Fuochi:
#### Description
This is to correct what I think are a couple of small issue with the api doc:
- correct indentation of get user response
- add missing parameter in slider update (as per https://swagger.io/docs/specification/describing-parameters/)

### DIFF
--- a/overseerr-api.yml
+++ b/overseerr-api.yml
@@ -3473,7 +3473,7 @@ paths:
                   results:
                     type: array
                     items:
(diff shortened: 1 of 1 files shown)


danshilm (approved):
Looks good! 
Thank you for the fix 👍🏽

danshilm:
@allcontributors please add @Fuochi for doc

PR #3650, opened 2023-10-18
Update SameSite policy of session cookie to Lax

RemiRigal:
#### Description

The session cookie `connect.sid` currently has a `Strict` `SameSite` policy, which prevents users to access pages from an external link. Requests with an external `Referrer` are systematically redirected to the login page because the session cookie is not added to the request. Changing this policy will not add vulnerabilities to the application.

Also, an alternative would be to set this policy to `Lax` only when the the `CSRF Protection` is disabled, let me know if this is preferable.

#### To-Dos

- [X] Translation keys `yarn i18n:extract`

#### Issues Fixed or Closed

- Fixes [Overseerr-Assistant #11](https://github.com/RemiRigal/Overseerr-Assistant/issues/11)

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -152,7 +152,7 @@ app
         cookie: {
           maxAge: 1000 * 60 * 60 * 24 * 30,
           httpOnly: true,
-          sameSite: true,
+          sameSite: 'lax',
           secure: 'auto',
         },
         store: new TypeormStore({


sct:
I haven't taken the time to look into fixing the issue, so thanks for doing this.

Since this is a security-related issue, I do think it makes sense to put this change behind having CSRF protection off. And then keep it Strict if CSRF protection is on.

RemiRigal:
Thank you for the fast feedback @sct !

I've updated the PR to set the `SameSite` policy based on the CSRF Protection setting. Let me know if anything else needs to be updated.

sct (approved):
LGTM. Let's give it a try!

sct:
@all-contributors please add @RemiRigal for code

PR #3690, opened 2023-11-16
feat(manageslideover): button to delete a media from Radarr/Sonarr

dd060606:
#### Description
New button in ManageSlideOver to delete a media and its file from Radarr/Sonarr
I have already created a similar [pull request](https://github.com/Fallenbagel/jellyseerr/pull/182) on the Jellyseerr repository and have been asked to redo this feature on Overseerr.

(Sorry for my English, I'm French)
image

#### To-Dos


OwsleyJr:
Great work on this so far but I don't see us adding any sort of media deletion functionality until we add individual episode support.

danielcherubini:
Would it at all be possible if this PR was split into two? So that whilst TV is requiring individual episode support, Movie deletion could make the light of day earlier?

noleska:
> Would it at all be possible if this PR was split into two? So that whilst TV is requiring individual episode support, Movie deletion could make the light of day earlier?

This and/or allow tv ep deletion at the same or coarser granularity as is supported in adding. It looks like this would already do that, in facilitating only entire series deletion.

PR #3695, opened 2023-11-19
fix(middleware): enhanced user privacy on profile pages

Fallenbagel:
#### Description
Addresses a security vulnerability where the `/users/[:id]` route was accessible to users without the necessary permissions. Adds middleware that protects that route so that only authenticated users with the `MANAGE_USERS` and `VIEW_WATCHLIST` permissions can access other user's profile pages as intended.

(unless if the current behavior is the intended behavior, then feel free to close the PR :smile: ).

#### To-Dos

### DIFF
--- a/server/routes/user/index.ts
+++ b/server/routes/user/index.ts
@@ -180,21 +180,25 @@ router.post<
   }
 });
 
-router.get<{ id: string }>('/:id', async (req, res, next) => {
-  try {
-    const userRepository = getRepository(User);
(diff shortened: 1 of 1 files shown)


Fallenbagel:
> The current behavior is the intended behavior. The content visible on profile pages is dependent on the active user's permissions.

Ah I see. This was raised because there were users who wanted the usernames/pfps and such to not be exposed to other users unless given the permission to do so as it concerns privacy

PR #3700, opened 2023-11-27
Correct config path in docker docs

Bre77:
#### Description
The path in all the documentation doesn't match what's required by the actual docker container.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX


sct:
The documentation is correct for the official image.

PR #3764, opened 2024-01-26
feat(server): fix queue processing

bertho-zero:
#### Description

This PR fixes an issue that does not move all downloads up the queue. The server is currently requesting a page of 10 items.

The fix requests the size of the queue before request for the content of the queue to allow more than 10 elements to be seen.

### DIFF
--- a/server/api/servarr/base.ts
+++ b/server/api/servarr/base.ts
@@ -48,6 +48,17 @@ export interface QualityProfile {
   name: string;
 }
 
+interface QueueStatus {
+  id: number;
+  totalCount: number;
+  count: number;
+  unknownCount: number;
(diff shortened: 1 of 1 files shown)


bertho-zero:
Don't stale !

OwsleyJr:
If we were to implement this, we would need some way to handle the pagination if there were was a large queue.

danshilm:
Agree with @OwsleyJr .
I think that a page size of 25 with a limit at 500 (or maybe 250?) should be good here?
I can definitely imagine a scenario where someone's download client bugged out/got corrupted and a user finds themselves with 100's or even 1000's of items in their Sonarr/Radarr queue.

PR #3808, opened 2024-03-12
Update workflow version

Gauvino:


Gauvino:
Can also be merged without problem, removing all the warning in action

PR #3810, opened 2024-03-12
feat(usersettings): add ability to hide tags on details page

### LINKED ISSUE #3350: Hide tags from detail page

jordanjones98:
fix #3350

#### Description

Add ability for a user to collapse tags on the movie/tv detail page from a checkbox. I made the text the same size as the season text, maybe it should be smaller? Up to do whatever there.

[question] This is at the user level right now, because that is where the other PR had it, should it be a server level setting?

Collapse Tags checkbox in user settings page
image

Collapse Tags checkbox unchecked (default)
image

Collapse Tags checkbox checked
image

#### To-Dos

#### Issues Fixed or Closed

- Fixes #3350


sct:
What about instead of just hiding it entirely, we add an option to auto collapse the tags? That way users could still see them if they want to?

jordanjones98:
> What about instead of just hiding it entirely, we add an option to auto collapse the tags? That way users could still see them if they want to?

Yeah I like that idea, I can implement that.

jordanjones98:
> What about instead of just hiding it entirely, we add an option to auto collapse the tags? That way users could still see them if they want to?

I updated it, I am probably going to pull out UI into it's own component so it is in one spot, but that is a tomorrow problem. Functionality and UI is there right now.

jordanjones98:
@sct Updated to use a shared component. The only configurable thing besides the tags is the type of `movie` or `tv` in the URL. I think just a prop with those two options is safe, but if you would rather do it a different way let me know. That migration should be fixed now too.

src/components/UserProfile/UserSettings/UserGeneralSettings/index.tsx
```diff
@@ -55,6 +55,8 @@ const messages = defineMessages({
   plexwatchlistsyncseries: 'Auto-Request Series',
   plexwatchlistsyncseriestip:
     'Automatically request series on your <PlexWatchlistSupportLink>Plex Watchlist</PlexWatchlistSupportLink>',
+  collapsetags: 'Collapse Tags',
+  collapsetagstip: 'Collapse tags by default in Movie/Series detail page',
```
OwsleyJr, line 59: Think this sounds slightly better: Collapse tags by default on the Movie/Series details page

PR #3847, opened 2024-04-18
feat: remove NSFW category image

jordanjones98:
#### Description

The category image for romance was VERY NSFW so this changes the index from 4 to 5 for that image which is no longer NSFW. I looked to see if the TMDB API included anything to make it non NSFW but it doesn't seem to. This is the best solution for now I think.

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/src/components/Discover/MovieGenreList/index.tsx
+++ b/src/components/Discover/MovieGenreList/index.tsx
@@ -39,7 +39,7 @@ const MovieGenreList = () => {
               name={genre.name}
               image={`https://image.tmdb.org/t/p/w1280_filter(duotone,${
                 genreColorMap[genre.id] ?? genreColorMap[0]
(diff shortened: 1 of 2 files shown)


sct:
This is the 4th index from the popular list. Changing the index may fix it for now but the position could easily change. The proper fix it to go and properly mark the content as adult in TMDB.

jordanjones98:
> This is the 4th index from the popular list. Changing the index may fix it for now but the position could easily change. The proper fix it to go and properly mark the content as adult in TMDB.

Totally, I am not sure how much I like this change at all, but it does solve the issue. There seems to be a big issue generally with adult content that isn't marked as adult. I am not a pro with TMDB so will look into how to mark those as adult.

jordanjones98:
I will close this

PR #3901, opened 2024-07-21
feat(watchlist): Cache watchlist requests with matching E-Tags

### LINKED ISSUE #3754: Watchlist visible instantly but long wait time for the request
### Description

I activated the automatic request when adding a movie to the Plex watchlist, I see it arrive immediately on Overseerr app but the request takes between 14 and 24 minutes.

I activated the automatic request when adding a film to Plex, I see it arrive immediately on Overseerr but the request takes between 14 and 24 minutes.

### Desired Behavior

The ideal would be not to wait long before launching the request

### Additional Context

_No response_

### Code of Conduct

maxnatamo:
#### Description

Prevents fetching the entire watchlist for a Plex user, when the watchlist hasn't been updated. It will now utilize the `If-None-Match` HTTP header to check whether the watchlist has been updated. Since Plex sends back a `X-Plex-Cache` header in the response, I'd presume they have some sort of caching built in to their backend.

This will also limit most sync requests going to Plex, so it is possible to increase the frequency of the sync task. I've set it to an arbitrary value of every 3 minutes, but feel free to comment about it.

Given the increased frequency, I've also limited the watchlist request, when updated, to 20 items instead of 200. I can't imagine most people add more than 20 items in 3 minutes, anyway.

#### To-Dos

- [N/A] Database migration (if required)

#### Issues Fixed or Closed

- Fixes #3754

### DIFF
--- a/server/api/plextv.ts
+++ b/server/api/plextv.ts
@@ -127,6 +127,11 @@ export interface PlexWatchlistItem {
   title: string;
 }
 
+export interface PlexWatchlistCache {
+  etag: string;
+  response: WatchlistResponse;
+}
+
 class PlexTvAPI extends ExternalAPI {
   private authToken: string;
 
@@ -270,6 +275,11 @@ class PlexTvAPI extends ExternalAPI {
     items: PlexWatchlistItem[];
   }> {
     try {
+      const watchlistCache = cacheManager.getCache('plexwatchlist');
+      let cachedWatchlist = watchlistCache.data.get<PlexWatchlistCache>(
+        this.authToken
+      );
+
       const response = await this.axios.get<WatchlistResponse>(
         '/library/sections/watchlist/all',
         {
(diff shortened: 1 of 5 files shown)


sct:
This is great. Just some questions.

> this header is likely handled by Cloudflare

Have we done any testing? I'd hate to revert this again and then crash the watchlist servers like we did last time.

maxnatamo:
Without actually asking Plex, I'm unsure of how to confirm that responses are actually being cached. When sending a request to Plex' watchlist endpoint, they send back a `X-Plex-Cache` header, which likely means they have some sort of caching on their end.

I've updated the description of the PR, since I don't actually think Cloudflare *could* effectively cache this. If any caching is done, it'd have to be on Plex' end.

sct:
Okay Cloudflare or not, I just want to make sure we are not hitting their server every time. So if the etag looks like it's not changing for identical watchlist requests that should be good enough.

maxnatamo:
I can confirm the e-tag does not change or identical watchlists. If there was some test I could write to prove it, I would. What I observed is:
- When sending a request to Plex, they send back an `E-Tag` header.
- Subsequent requests with the same watchlist sends back the same e-tag
- Likewise, updating the watchlist also updates the e-tag.
- Sending a `If-None-Match` header with the cached e-tag causes Plex to send back `HTTP 304` without any body, unless the watchlist has updated.

sct (approved):
LGTM. Lets give it a try. We can revert if there is any issues.

sct:
@all-contributors please add @maxnatamo for code

jminardi:
@maxnatamo Does this fix require a plex pass to work?

PR #3909, opened 2024-08-01
feat: persistent filters

### LINKED ISSUE #3889: Remember Filters on Movies & Series
### Description

I only use the Japanese language filter for content. But it always defaults back to english content primarily. I set my filter, but then it gets cleared. Is there anyway for it to rememeber it (like local storage ?)

### Desired Behavior

Remember what I set as my default filters

### Additional Context

_No response_

### Code of Conduct

leejayhsu:
#### Description
Use localstorage to save filters and orderBy settings in discover tv/movie pages.

output

#### To-Dos

#### Issues Fixed or Closed

- Closes #3889

### DIFF
--- a/src/components/Discover/DiscoverMovies/index.tsx
+++ b/src/components/Discover/DiscoverMovies/index.tsx
@@ -9,6 +9,7 @@ import {
 } from '@app/components/Discover/constants';
 import FilterSlideover from '@app/components/Discover/FilterSlideover';
 import useDiscover from '@app/hooks/useDiscover';
+import usePersistentFilters from '@app/hooks/usePersistentFilteres';
 import { useUpdateQueryParams } from '@app/hooks/useUpdateQueryParams';
 import Error from '@app/pages/_error';
 import { BarsArrowDownIcon, FunnelIcon } from '@heroicons/react/24/solid';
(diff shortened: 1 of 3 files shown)


sct:
One issue I can imagine with this is that the sliders on the home page all apply filters when clicked on. These would then persist and override any of the expected persistent filters. This might be super confusing.

On top of that I would probably like to improve how we show filters are active on the movies/tv list. It's a bit hard to actually realize you have filters active (especially if you didn't actively set them).

BillTheGoat:
I agree that this is absolutely critical. I was about to abandon Jellyseerr until I saw this PR. However @sct is right that it is not fully integrated. I added this pr and rebuilt. What happens is the slider links do nothing, you just return to your previously defined filters.

Perhaps the best desired functionality would be that the slider link overwrites only the persistent filter for the filter category it is in. For example, a genre link form the main discover page would overwrite the persistent genres, but not the persistent languages or other active filters.

This would not be hard to implement. Just read the query params from the url, and merge with the `savedFilters`.

Does that sound the right approach?

sct:
I will close this PR now since the concerns haven't been addressed. Please feel free to re-open this PR if we can make the relevant changes and I'll be happy to review it.

Issue #3580, opened 2023-08-09
Customize issue reporting or allow blank report


mudsloth:
### Description

Reporting an issue has Video, Audio, Subtitle, and Other as categories, then requires a description of the problem. I'd like to customize those categories to what would be simpler for me to triage or my users to handle. And, I'd like to be able to allow an issue report with a specific category get reported without a full description, simplifying the process for my users.

### Desired Behavior

For users that have control over managing issues, have a setting to add/remove/rename the issue categories that are used for reporting. For each category, have a setting to require a full description which can be set to on or off.

### Code of Conduct

- [X] I agree to follow Overseerr's Code of Conduct

kirsch33:
i second this request!

PR #3915, opened 2024-08-11
fix: allow blank issue descriptions

kirsch33:
#### Description
This PR allows blank/empty issue descriptions to be submitted and removes the "*" required label on the modal form.

#### To-Dos

#### Issues Fixed or Closed

- partially addresses https://github.com/sct/overseerr/issues/3580

### DIFF
--- a/src/components/IssueModal/CreateIssueModal/index.tsx
+++ b/src/components/IssueModal/CreateIssueModal/index.tsx
@@ -19,7 +19,6 @@ import useSWR from 'swr';
 import * as Yup from 'yup';
 
 const messages = defineMessages({
-  validationMessageRequired: 'You must provide a description',
   whatswrong: "What's wrong?",
   providedetail:
     'Please provide a detailed explanation of the issue you encountered.',
@@ -84,9 +83,7 @@ const CreateIssueModal = ({
     .map((season) => season.seasonNumber);
 
   const CreateIssueModalSchema = Yup.object().shape({
-    message: Yup.string().required(
-      intl.formatMessage(messages.validationMessageRequired)
-    ),
+    message: Yup.string().notRequired(),
   });
 
   return (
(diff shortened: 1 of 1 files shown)


danshilm:
More changes are required for this to be properly implemented; an issue comment is still created for the issue, just with an empty text. The endpoint to create issues needs to be modified to accept Issues without comments.

kirsch33:
I don't follow. In my testing this performs exactly as intended. As issue is created without any comments and with a blank description. 

After peeking around the code base some I still don't see what else would need to be modified. Can you provide a bit more info as to what endpoint needs to be modified ?

sct:
I would also not like to encourage people to not leave comments. I think they can spend a small amount of time writing in what the issue is.

kirsch33:
@sct in my case I have an external program that operates via webhooks and looks for subtitles issues. If found it initiates a sync.

Its much easier to explain to family/friends that they simply need to click the issue button and change the type to "subtitle" versus asking them to also describe that its out of sync (since that's the only possible subtitle related issue at least with my specific setup it is redundant). Kind of a niche situation, but that's what lead to this PR.

Anyways, if this doesn't jive with the vision for Overseerr, I understand.

sct:
> @sct in my case I have an external program that operates via webhooks and looks for subtitles issues. If found it initiates a sync.
> 
> Its much easier to explain to family/friends that they simply need to click the issue button and change the type to "subtitle" versus asking them to also describe that its out of sync (since that's the only possible subtitle related issue at least with my specific setup it is redundant). Kind of a niche situation, but that's what lead to this PR.
> 
> Anyways, if this doesn't jive with the vision for Overseerr, I understand.

That's a really cool use-case! I love the idea but unfortunately I would still like to avoid making this optional (For my own sanity, and probably many others). For now I think best to just let your users know they should just write in literally any text in the description field. Hopefully that's not too much of a bother.

kirsch33:
@sct not at all, I understand. Thank you for your time and for Overseerr itself.

Lockszmith-GH:
Sorry to ressurrect this, but how about a default description when subtitle is selected?

PR #3939, opened 2024-09-22
fix: prevent icon from displaying on certain phones when not pulled

OwsleyJr:
#### Description

This PR fixes an issue where the pull-to-refresh icon was incorrectly displaying in its default state on certain phones.

To resolve this, I moved the logic for detecting the initial pull-to-refresh gesture to the layout component. This change will make sure that the icon will only render when the user is at the top of the page and actively pulling down.

IMG_5997

#### To-Dos

### DIFF
--- a/src/components/Layout/PullToRefresh/index.tsx
+++ b/src/components/Layout/PullToRefresh/index.tsx
@@ -4,7 +4,6 @@ import { useEffect, useRef, useState } from 'react';
 
 const PullToRefresh = () => {
   const router = useRouter();
-
   const [pullStartPoint, setPullStartPoint] = useState(0);
   const [pullChange, setPullChange] = useState(0);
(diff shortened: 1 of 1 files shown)


src/components/Layout/index.tsx
```diff
@@ -49,6 +51,36 @@ const Layout = ({ children }: LayoutProps) => {
...
+        refreshDiv.current?.classList.add('hidden');
+      }
+    };
+
+    window.addEventListener('touchstart', pullStart, { passive: false });
+
+    return () => {
+      window.removeEventListener('touchstart', pullStart);
+    };
+  }, []);
```
sct, line 82: Why can't this just be in the pull to refresh component and you have the component return null when the conditions are not met? That we we don't have to juggle this state between two components?

sct (approved):

PR #4006, opened 2024-12-28
Add displayname to plex tv api and use in watchlist log

Samywamy10:
#### Description

Adds a displayname to the `getWatchlist` endpoint so in the error logs we can see which users need to reauthenticate

#### To-Dos


Samywamy10:
why close

PR #4019, opened 2025-01-15
Merge series tags instead of overwriting them

andrew-kennedy:
#### Description
This PR ensures we merge tags with existing tags in sonarr and radarr instead of overwriting them with media request tags.


danshilm:
@all-contributors please add @andrew-kennedy for code

PR #4021, opened 2025-01-20
feat(watchlistsync): add 4K support for auto request from plex watchlist

### LINKED ISSUE #3849: [Feature Request] Auto-Request 4K from watchlist
### Description

I'm currently using the auto-request feature with my users' Plex watchlists, but it only requests HD. I have a separate 4K default instance configured in Overseerr, but I still have to manually request things in 4K, which is slow and frustrating. 

### Desired Behavior

There should be a new Auto-Request setting for 4K content alongside the Auto-Request setting for non-4K content. 
When enabled, watchlist syncs should automatically send requests to the default 4K radarr/sonarr instance. 

### Additional Context

I'm currently achieving this with a list sync from the HD instance to the 4K instance, but it's slow and doesn't allow me any of the control or notifications that I can set up in Overseerr. It would be better if Overseerr could push requests to both instances automatically.

### Code of Conduct

alperenkaplan:
modified the auto request logic to allow users with auto_request, request_4k and auto_approve_4k permissions to request content with 4k profile

#### Description
I wanted the users to be able to auto request content with the existing plex watchlist sync feature with the 4k profile. I was actually going to add extra settings to user permissions but then it made more sense to auto request in 4k if the user already had auto_request, request_4k and auto_approve_4k permissions. If a user can auto request, request in 4k and auto approve in 4k, I don't see why we would need any extra settings or permissions and why the users couldn't auto request in 4k. If they can already do this by logging into overseer and request and auto approve in 4k, I think we should also request their plex watchlist in 4k.
#### To-Dos

- [+] Successful build `yarn build`

#### Issues Fixed or Closed

- Fixes #3849

### DIFF
--- a/server/lib/watchlistsync.ts
+++ b/server/lib/watchlistsync.ts
@@ -117,7 +117,39 @@ class WatchlistSync {
               mediaItem.type === 'show' ? MediaType.TV : MediaType.MOVIE,
             seasons: mediaItem.type === 'show' ? 'all' : undefined,
             tvdbId: mediaItem.tvdbId,
-            is4k: false,
+            is4k: 
+              (mediaItem.type === 'movie' &&
+                user.hasPermission(
+                  [
+                    Permission.REQUEST_4K,
+                    Permission.REQUEST_4K_MOVIE,
+                  ],
+                  { type: 'or' }
+                ) &&
+                user.hasPermission(
+                  [
+                    Permission.AUTO_APPROVE_4K,
+                    Permission.AUTO_APPROVE_4K_MOVIE,
+                  ],
+                  { type: 'or' }
+                )
+              ) ||
+              (mediaItem.type === 'show' &&
+                user.hasPermission(
+                  [
+                    Permission.REQUEST_4K,
+                    Permission.REQUEST_4K_TV,
+                  ],
+                  { type: 'or' }
+                ) &&
+                user.hasPermission(
+                  [
+                    Permission.AUTO_APPROVE_4K,
+                    Permission.AUTO_APPROVE_4K_TV,
+                  ],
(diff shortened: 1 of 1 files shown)


dodofarm:
Maybe post this in the Jellyseerr repo? Overseerrs development has been on ice for quite a bit now and Jellyseerr is a fork that added more new features.

OwsleyJr:
This would just make all watchlist requests 4K if they have the correct permissions. Until Plex provides a way to distinguish between 4K and non-4K media in the watchlist, I don't think we can add this.

alperenkaplan:
> This would just make all watchlist requests 4K if they have the correct permissions. Until Plex provides a way to distinguish between 4K and non-4K media in the watchlist, I don't think we can add this.

I guess that makes sense. I have the storage capacity and a gigabit internet so I download everything in 4K remux but I didn't think people might not want EVERYTHING in 4K. I just removed the 4K server from the settings and made it the default so everything downloads in 4K now.

On the other hand, I doubt Plex will ever add a way to distinguish 4K media so we need to come up with a different solution. And still, I can't think of anything else other than my current implementation (giving certain users the necessary permissions so their watchlist items are downloaded in 4K, unfortunately all the time)

oxfordllama:
I just want to chime in and say that I have this exact use case. I would like to be able to choose which users request 4K and which request standard content. I don't need a way to differentiate between standard/4K on a per-request basis, just per-user.

seeARMS:
I also wanted to chime in and say that this would also be valuable for me. On my Plex/Overseer user I want to request everything in 4K if possible, but for other users I want to force them always to request non-4K. Similar to @oxfordllama -- this is on a per-user, not per-request basis.

PR #4066, opened 2025-03-04
fix(ipv6 regex matching): replaced existing regex for IPv4/6 address …

### LINKED ISSUE #4012: IPv6 URLs not supported in Services
### Description

IPv6 IPs are not supported in the Services hostname and external URL fields.
With and without brackets around the address, overseerr complains "You must provide a valid hostname or IP address"

### Version

1.33.2

### Steps to Reproduce

1. Go to Settings -> Services
2. Add a Radarr/Sonarr server
3. Put an IPv6 address in the `Hostname or IP Address` field (with and without brackets)
4. Click out of the field
5. Observe error `You must provide a valid hostname or IP address`

### Screenshots

Screenshot_20250102_212328
Screenshot_20250102_212448

### Logs

_No response_

### Platform

desktop

### Device

Desktop

### Operating System

Debian 12

### Browser

Firefox 133

### Additional Context

_No response_

### Code of Conduct

lmiklosko:
…matching with new one

Current regex does not correctly capture valid ipv6 addresses, such as fd01::3:172:20:24:5 and thinks 127.0.0.0.0.0.1 is a valid address. This PR addresses these issues.

fix #4012

#### To-Dos

#### Issues Fixed or Closed

- Fixes #4012

### DIFF
--- a/src/components/Settings/Notifications/NotificationsEmail.tsx
+++ b/src/components/Settings/Notifications/NotificationsEmail.tsx
@@ -3,6 +3,7 @@ import LoadingSpinner from '@app/components/Common/LoadingSpinner';
 import SensitiveInput from '@app/components/Common/SensitiveInput';
 import SettingsBadge from '@app/components/Settings/SettingsBadge';
 import globalMessages from '@app/i18n/globalMessages';
+import { hostnameValidationRegex } from '@app/utils/validationHelpers';
 import { ArrowDownOnSquareIcon, BeakerIcon } from '@heroicons/react/24/outline';
 import axios from 'axios';
 import { Field, Form, Formik } from 'formik';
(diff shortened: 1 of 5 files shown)


MilanJa:
Since they're being reused, you might as well extract those regex and put them into a `src/utils/validationHelpers.ts` file that could be the single point for all complex regex and all the uses of it have descriptive names instead. But otherwise lgtm

sct:
> Since they're being reused, you might as well extract those regex and put them into a `src/utils/validationHelpers.ts` file that could be the single point for all complex regex and all the uses of it have descriptive names instead. But otherwise lgtm

I think this is great feedback. Would you mind adopting this @lmiklosko?

PR #4068, opened 2025-03-09
feat(watchlist): display media from external API as a DiscoverSliderType

dohnutt:
#### Description

Defines a new `DiscoverSliderType` named `ExternalWatchlistSlider`, which pulls TMDB IDs from a given external API and displays them in a slider view, very similar to how Overseerr displays your Plex Watchlist.

I'm pretty inexperienced with React so I leaned heavily on existing conventions within the project—namely the `PlexWatchlistSlider`. This implementation displays the first 20 results from your external API (in my example, a scraped Google Watchlist from [a personal project](https://github.com/dohnutt/searchlistarr), but ideally it would be able to display the full results of the API in a paginated/infinite scroll view similar to the Plex Watchlist.

The challenges here, as far as I can tell, are:
- Creating a `ListView` route per slider—though a way around this is to have one route with parameters passed to it (`/discover/external-watchlist?apiUrl=http://example.com/api&pageParam=p`)
- As subtly referenced in the above bullet, you'd need to somehow define how to paginate the API. Is the URL param `p=X` or `page=X`, or perhaps something else?
- How would you map fields from the external API to a `WatchlistItem[]`?

Perhaps these challenges could be out of scope of this feature request and it could simply be a `DiscoverSliderType`. This could especially work if it were paired with a `linkUrl` property so as to link to your external list rather than try to build a frontend into Overseerr—but of course then you lose the request functionality on the rest of your external list.

I am very green to contributing to open source, very green to React, and more generally comfortable with HTML and CSS than with complex JS so please feel free to tear it apart! If you've read this far, thank you for your time ❤️

localhost_5055_ (2)*<small>External Watchlist slider being added — note that I am not sure the best way to use`TmdbTitleCard` here to display these items properly, so they are displaying without proper TMDB metadata.</small>*

localhost_5055_ (1)
*<small>External Watchlist slider in the discover edit view</small>*

localhost_5055_
*<small>External Watchlist slider on the discover page</small>*

#### To-Dos

#### Issues Fixed or Closed

N/A


dohnutt:
I figured out the standalone page fed by the external API. You can see here dohnutt/overseerr@e121845

The page, at `/discover/externalwatchlist`, accepts a couple URL parameters so that the page can be built on the fly. For example,
/discover/externalwatchlist?apiUrl=http:example.com/api&title=Google%20Watchlist

could return this page
localhost_5055_discover_externalwatchlist_apiUrl=http%3A%2F%2F192 168 86 226%3A5155%2Fapi%2Fv1%2Fwatchlist title=Google%20Watchlist

PR #4107, opened 2025-04-13
User activity tracking in User list

Richardvaio:
#### What It Does

This pull request adds a  `/active` endpoint which is used in the user list to display which users are currently active and when they were last active. It also supports filtering users by their `lastActive` status for better user management.

Screenshot

#### To-Dos


server/routes/user/index.ts
```diff
@@ -23,7 +23,23 @@ import gravatarUrl from 'gravatar-url';
...
+  // Fetch users active in the last 5 minutes
+  const fiveMinutesAgo = new Date(Date.now() - 5 * 60 * 1000);
+
+  const activeUsers = await userRepository.find({
+    where: { lastActive: MoreThan(fiveMinutesAgo.toISOString()) },
+    select: ['id', 'displayName'], // Select only necessary fields
+  });
+
+  return res.json(activeUsers);
+});
```
sct, line 42: Why does this route even exist?

server/routes/index.ts
```diff
@@ -148,6 +148,7 @@ router.use('/media', isAuthenticated(), mediaRoutes);
 router.use('/person', isAuthenticated(), personRoutes);
 router.use('/collection', isAuthenticated(), collectionRoutes);
 router.use('/service', isAuthenticated(), serviceRoutes);
+router.use(trackUserActivity);
```
sct, line 151: This should probably be higher up alongside our other middleware. Also, I am not sure how I feel about us running a user update on every single API request. I think just doing this in our usual auth check would be enough.

src/components/UserList/index.tsx
```diff
@@ -76,9 +77,74 @@ const messages = defineMessages({
...
     'The <strong>Enable Local Sign-In</strong> setting is currently disabled.',
+  neverSignedIn: 'Never Signed In',
+  activeNow: 'Active Now',
+  lastActive: 'Last Active',
 });
 
-type Sort = 'created' | 'updated' | 'requests' | 'displayname';
+
+// Helper function to format the last active time
+const formatLastActive = (lastActive: string | null, intl: IntlShape): string => {
```
sct, line 87: You can use formatRelativeTime instead of doing all of this manually. Look at other examples in Overseerr on how to use it.

PR #4124, opened 2025-05-01
refactor(MediaRequest): consolidate lifecycle hooks into subscriber & restrict cascade to prevent recursion

fallenbagel:
#### Description
This PR consolidates all database‐accessing lifecycle hooks into the `MediaRequestSubscriber` as this is the recommended way of doing database calls. 
See: https://orkhan.gitbook.io/typeorm/docs/listeners-and-subscribers#what-is-an-entity-listener.

It also tightens the cascade options on the `MediaRequest→Media` relation to only `insert` and `remove` to avoid`hook‐within‐hook` recursions that could cause unexpected behavior. 

These changes should guard against recursive hook invocations in any database backend and simplify the entity lifecycle.

Reference: https://github.com/fallenbagel/jellyseerr/pull/1619

#### To-Dos

#### Issues Fixed or Closed

- Fixes #XXXX

### DIFF
--- a/server/subscriber/MediaRequestSubscriber.ts
+++ b/server/subscriber/MediaRequestSubscriber.ts
@@ -21,6 +21,7 @@ import { getSettings } from '@server/lib/settings';
 import logger from '@server/logger';
 import { isEqual, truncate } from 'lodash';
 import type {
+  EntityManager,
   EntitySubscriberInterface,
   InsertEvent,
   RemoveEvent,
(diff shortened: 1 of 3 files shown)


server/subscriber/MediaRequestSubscriber.ts
```diff
@@ -125,6 +773,28 @@ export class MediaRequestSubscriber
...
+  public afterInsert(event: InsertEvent<MediaRequest>): void {
+    if (!event.entity) {
+      return;
+    }
+
+    this.sendToRadarr(event.entity as MediaRequest);
+    this.sendToSonarr(event.entity as MediaRequest);
+
+    this.updateParentStatus(event.entity as MediaRequest);
+  }
```
OwsleyJr, line 785: Do we want to make afterUpdate and afterInsert here async?

fallenbagel: Oh yeah! Both `afterInsert` and `afterUpdate` should be async and await the async methods inside. Otherwise, we might get unhandled promises or race conditions 🤔

I'll update it today

fallenbagel: @OwsleyJr Should be resolved in the latest commit

OwsleyJr (approved):

PR #4136, opened 2025-05-09
fix(mediasubscriber): remove async from afterInsert and afterUpdate

gauthier-th:
#### Description

Asynchronous operations introduced in commit 3b9e195496c3db1f794549786b123a3ff55b08a0 caused issues during media requests: the mediaId was not properly assigned and remained null.

#### To-Dos

#### Issues Fixed or Closed

- Related to #4124


fallenbagel (commented):
Yup. My bad. It seems to affect some setups. Like on my native postgresql, it works. But when setup with docker i can reproduce. It should not have been async

sct (approved):

PR #4216, opened 2025-08-06
fix(translation): incorrect Ukrainian "pending" translation

gshpychka:
#### Description

Changed one incorrect string


0xSysR3ll:
Translations are exclusively handled by [Weblate](https://hosted.weblate.org/engage/overseerr/). You can signup, search for the right string and correct it :)

